PR #1, opened 2017-06-22
Error to generate Orderer Genesis block

batidiane:
Add FABRIC_CFG_PATH variable to allow configtxgen to find configuration file

### DIFF
--- a/first-network/base/peer-base.yaml
+++ b/first-network/base/peer-base.yaml
@@ -10,9 +10,11 @@ services:
     environment:
       - CORE_VM_ENDPOINT=unix:///host/var/run/docker.sock
       # the following setting starts chaincode containers on the same
-      # bridge network as the peers
+      # bridge network as the peers.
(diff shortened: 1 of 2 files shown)


rjones-lf:
Thank you for your contribution! This is a read only mirror, however. Please submit your change using gerrit.

you would need to check out the repo here: https://gerrit.hyperledger.org/r/#/admin/projects/fabric-samples and make your edits, pushing them to gerrit

here is a walkthrough for zephyrproject which is much the same WRT setting up an LFID, adding ssh keys, etc:
https://www.zephyrproject.org/doc/1.1.0/collaboration/code/gerrit_accounts.html

feel free to ask on rocket.chat - https://chat.hyperledger.org/ - discuss in #fabric

one note: when you set up your LFID do not use social logins the first time, create the account and you can add social logins later

batidiane:
Thank you for the information, I will create a gerrit account then.

TB.

PR #3, opened 2017-08-15
update to byfn.sh exec, fix bug in networkUp function

rddill-IBM:
This is posted in StackOverflow [here](https://stackoverflow.com/questions/45679395/hyperledger-fabric-first-network-fails-with-bad-request/45694630#45694630). The networkUp function in byfn.sh calls docker compose twice, once prior to the ```IF_COUCHDB``` test and a second time within the scope of that if statement. The networkUp function also fails to set the DELAY argument on either of the docker-compose requests within the if statement. In this pull request, I've commented out line 113 (the duplicate docker-compose request) and added ```DELAY=$CLI_DELAY``` to each of the docker-compose statements inside the if structure. The default setting of 3 seconds is now passed correctly to docker-compose and the network up function processes correctly.


rjones-lf:
Thank you for your contribution! This is a read only mirror, however. Please submit your change using gerrit.

you would need to check out the repo here: https://gerrit.hyperledger.org/r/#/admin/projects/fabric-samples and make your edits, pushing them to gerrit

here is a walkthrough for zephyrproject which is much the same WRT setting up an LFID, adding ssh keys, etc:
https://www.zephyrproject.org/doc/1.1.0/collaboration/code/gerrit_accounts.html

feel free to ask on rocket.chat - https://chat.hyperledger.org/ - discuss in #fabric

one note: when you set up your LFID do not use social logins the first time, create the account and you can add social logins later

PR #6, opened 2017-09-03
Added quotes to $PWD

lezwon:
Without quotes the shell executables fail to run, when the directory path
has spaces in Windows


rjones-lf:
Thank you for your contribution. This is a read only mirror, however. Please [read the guide here](http://hyperledger-fabric.readthedocs.io/en/latest/CONTRIBUTING.html) for directions on contributing.

PR #9, opened 2018-01-20
Fix "No identity type provided" Error

armujahid:
Fix the following errors:
[[{"code":0,"message":"No identity type provided. Please provide identity type"}]]


armujahid:
Closing PR because it is read only mirror

PR #12, opened 2018-03-28
balance transfer: fix register response error

zhayujie:
It responses empty secret value after successfully register and enroll, and the value of "enrollmentSecret" in persistence is also empty string. The added code set the "_enrollmentSecret" value of the User object and update the user context in persistence again.


ryjones:
Thanks for the contribution! this is a read-only mirror. please read the [contribution guide](http://hyperledger-fabric.readthedocs.io/en/release-1.1/CONTRIBUTING.html).

PR #18, opened 2018-04-25
Added env var support in docker-compose, updated documentation

tboloo:
Added .env specyfing default tag for the fabric docker images, which means that one does not have to explicitly  pull and retag images. Documentation updated accordingly.


ryjones:
Thanks for the contribution! this is a read-only mirror. please read the [contribution guide]().

Thanks for the contribution! However, this is a read-only mirror. please read the [contribution guide](http://hyperledger-fabric.readthedocs.io/en/release-1.1/CONTRIBUTING.html). Also, feel free to join us on [chat!](https://chat.hyperledger.org/channel/fabric)

PR #46, opened 2018-12-11
Add filtering function to the dkcl function

hag3:
The original dkcl function deletes all containers regardless of whether they were started by this sample or not (e.g. rancher/agent).
Doing so would corrupt the host's environment and cause inconvenience. So I added conditional filtering to the dkfc function.

### DIFF
--- a/balance-transfer/runApp.sh
+++ b/balance-transfer/runApp.sh
@@ -6,7 +6,7 @@
 #
 
 function dkcl(){
-        CONTAINER_IDS=$(docker ps -aq)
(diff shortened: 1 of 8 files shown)


ryjones:
Thanks for the contribution! However, this is a read-only mirror. please read the [contribution guide](http://hyperledger-fabric.readthedocs.io/en/release-1.1/CONTRIBUTING.html). Also, feel free to join us on [chat!](https://chat.hyperledger.org/channel/fabric)

PR #67, opened 2019-10-22
[FAB-16844] Correct BYFN CC name/Add cli for org2

mbwhite:
- Correct chaincode name regexp in byfn.sh
- Add in the useful cli for org2 in first-network

Change-Id: I6904db1786ae431f8a7069296ff610695d5a72c0
Signed-off-by: Matthew B. White <whitemat@uk.ibm.com>

### DIFF
--- a/first-network/byfn.sh
+++ b/first-network/byfn.sh
@@ -90,7 +90,7 @@ function askProceed() {
 # Obtain CONTAINER_IDS and remove them
 # TODO Might want to make this optional - could clear other containers
 function clearContainers() {
-  CONTAINER_IDS=$(docker ps -a | awk '($2 ~ /dev-peer.*.mycc.*/) {print $1}')
+  CONTAINER_IDS=$(docker ps -a | awk '($2 ~ /dev-peer.*/) {print $1}')
   if [ -z "$CONTAINER_IDS" -o "$CONTAINER_IDS" == " " ]; then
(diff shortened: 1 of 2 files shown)


lehors:
Hi,
I like having the org2cli too but without BFYN making use of it it's not that useful.
I had a chat with @nikhil550 a while ago and he agreed that we should move towards separating the operations done by the different orgs from one another and not use the same cli for both because that's not how a production system would get setup.

nikhil550:
@lehors The new test network: https://github.com/hyperledger/fabric-samples/pull/80 actually removes the fabric tools cli all together. Removing the cli container makes it easier to install chaincode, create channels, and edit channels without having to be worried on what you mount in the container. The certificates are still different for each org, which requires you to set different environment variables depending on which org you are operating from.

lehors:
Right, I actually meant to say that at the end of my comment but had to get on a plane and ended up posting it without it. My bad.
So, to finish my thought, given that direction there is no real value in adding the org2cli. I suggest dropping it from this PR.

nikhil550:
The current chaincode container cleanup leaves the fabcar chaincode images if you use the fabcar tutorial. That part of the PR would be welcome. Those changes are in the new test network.

mbwhite:
Will delete this PR and recreate one with just the byfn.sh cc name changes

PR #68, opened 2019-10-23
created missing scripts/bootstrap.sh file

dappsar:
Missing file in release-1.4 that is references in https://github.com/IBM/loyalty-points-evm-fabric


lehors:
Actually, this file was removed on purpose because fabric/scripts/bootstrap.sh should be used instead. It is https://github.com/IBM/loyalty-points-evm-fabric that should be updated.

PR #78, opened 2019-11-25
[FAB-17139] Update fabcar go to new programming model

awjh-ibm:
Signed-off-by: Andrew Hurt <andrew.hurt1@ibm.com>


awjh-ibm:
The contract API is stable. The only update to this will be to fix it to the version release of the contract API which will be done once chaincode-go is released to a fixed version

PR #94, opened 2019-12-27
Update fabric version

btl5037:

### DIFF
--- a/ci/install-fabric.yml
+++ b/ci/install-fabric.yml
@@ -5,26 +5,26 @@
 steps:
   - script: |
       set -ex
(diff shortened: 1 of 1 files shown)


btl5037:
@mbwhite @sstone1 While doing artifactory work, I discovered 1.4 CI was still using 1.4.4-stable images instead of 1.4.5-stable. There are errors in both java and typescript chaincode, I don't have the cycles to dig into this right now, this PR contains both of the failures

btl5037:
This isn't going to pass, someone needs to take a look at why 1.4.5 doesn't work with Samples, I unfortunately don't have the cycles at the time to look into it

lehors:
Ok, I'm going to override the CI failure because it has nothing to do with this PR.

ryjones:
shouldn't these have been artifactory urls?

lehors:
There is another PR to change to artifactory (#93) but it's true that this may render this PR moot...

PR #98, opened 2020-01-08
[FAB-16147] Update Commercial Paper to work with v2

mbwhite:
Update the applications and scripts to use the new v2
SDKs and the new lifecycle

Add in a basic script based on the readme.md that does
a basic run of the scenario


mbwhite:
Note: this is using the basic-network; have started moving this to test-network as it better represents the scenario,

commercial-paper/organization/magnetocorp/contract-java/src/main/java/org/example/ledgerapi/State.java
```diff
@@ -54,7 +53,7 @@ public static String makeKey(String[] keyParts) {
     }
 
     public static String[] splitKey(String key) {
-        System.out.println("Splitting key " + key + "   " + java.util.Arrays.asList(key.split(":")));
+        System.out.println("Splittin gkey " + key + "   " + java.util.Arrays.asList(key.split(":")));
```
sstone1: I think this was right before 😂

commercial-paper/organization/magnetocorp/contract-java/src/main/java/org/example/CommercialPaper.java
```diff
@@ -1,6 +1,7 @@
 /*
- *  SPDX-License-Identifier: Apache-2.0
+ * SPDX-License-Identifier:
```
sstone1: No license identifier?

commercial-paper/organization/magnetocorp/contract-java/src/main/java/org/example/ledgerapi/State.java
mbwhite: 😞  so need a compiler for English. ..

PR #104, opened 2020-01-17
Add 1.x versions of fabric to blacklisted versions

nikhil550:

Add 1.x versions of Fabric to the blacklist. The test network can only run on 2.0. If run on 1.4.x, the orderer will exit due to 2.0 capabilities.

Removing unused variable from script.

### DIFF
--- a/test-network/network.sh
+++ b/test-network/network.sh
@@ -79,8 +79,8 @@ function removeUnwantedImages() {
   fi
 }
 
-# Versions of fabric known not to work with this release of first-network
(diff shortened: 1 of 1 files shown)


test-network/network.sh
```diff
@@ -80,7 +80,7 @@ function removeUnwantedImages() {
 }
 
 # Versions of fabric known not to work with this release of first-network
```
mastersingh24:
```suggestion
# Versions of fabric known not to work with this release of test-network
```

nikhil550: done.

PR #113, opened 2020-02-05
FAB-17456 fabric-samples - pre-read ccp

harrisob:
Applications must pass an 'Common Connection Profile' object
to the Gateway as of v2. NodeSDK should not have access to the
file system.


harrisob:
using https://github.com/hyperledger/fabric-samples/pull/117 instead

PR #114, opened 2020-02-05
Modified command

adityainterwork:


lehors:
@adityainterwork Hi, it looks like you only meant to change the README but somehow you pulled in a whole bunch of other changes. Can you please update your PR to only include your change?
Also, to be able to merge your change we will need proper sign-off. I'll admit that it is moot in this case given the small change but our CI scripts won't let us merge it without it. See https://hyperledger-fabric.readthedocs.io/en/master/CONTRIBUTING.html#legal-stuff
Thanks.

PR #117, opened 2020-02-07
FAB-17456 fabric-samples read ccp

harrisob:
Have the application read the common connection profile to
build a JSON object used for the gateway connect and for
a fabric-ca client instance.


zillerium:
I have an error getClient is not a function. I would appreciate any ideas.

PR #123, opened 2020-02-13
test-network deployCC.sh invoke chaincode inconsistency

woodyjon:
Keep consistency of the function called in the "no TLS" and "with TLS" cases


woodyjon:
But in fabcar, which is used by this script, there is no init function. The call to init (as it is in the test-network script in the "no TLS" case) does not work.

nikhil550:
HI @woodyjon, sorry for the delay. I asked around and it turns out I had this wrong. If you use the new fabric programing model, you can initialize the chaincode using any transaction in the chaincode. (usually this would be done with an empty instantiate or Init function added by the developer). This means that the second invoke can actually be removed. It also means that this PR can go ahead and be merged.

woodyjon:
OK, thanks for the clarification!

lehors:
@woodyjon Thanks for your contribution but we won't be able to merge it without a sign-off. Can you please amend your commit with a sign-off in the log? See https://hyperledger-fabric.readthedocs.io/en/master/CONTRIBUTING.html#legal-stuff

woodyjon:
I just added the sign-off message to the commits. Is it ok like so?

lehors:
The DCO check is still failing. Please, follow the instructions:
https://github.com/hyperledger/fabric-samples/pull/123/checks?check_run_id=526791523
Thanks.

woodyjon:
There you go 😀

woodyjon:
If needed, I can redo the PR more cleanly, with a single commit.

lehors:
No worries, we'll squash the commits. Thanks.

lehors:
Wait, now there are 87 files changed... something went wrong.

woodyjon:
Do I make a new PR?

lehors:
If that's easier for you then just do that. What's clear is that we can't merge this PR as it stands so something needs to be done.

woodyjon:
I close this PR because I messed it up. Cleaner PR is there: https://github.com/hyperledger/fabric-samples/pull/141

PR #131, opened 2020-02-26
Revert #118

btl5037:
This reverts commit 9ef61e2d7204964ad02b53b171c05f87a673e42d.

Introduced regressions when running without explicitly specifying
a chaincode label

PR #132, opened 2020-02-26
first-network/scripts/*: Make Chaincode name configurable

maniankara:
References #131 #118 

Signed-off-by: Anoop Vijayan Maniankara <anoop@tuxera.com>


btl5037:
You'll need to do the same for `eyfn.sh` passing around the parameter

btl5037:
You can test this by doing `./byfn.sh down` then `./byfn.sh up` then `./eyfn.sh up`

maniankara:
@btl5037 Ah yes. Had missed that. 
Have applied changes now and also have updated the commit message to be more descriptive.

Thanks,

btl5037:
Thanks for the quick response @maniankara I'll take a look at it later tonight!

btl5037:
LGTM, tested locally and ran as expected, backwards functionality remains intact

first-network/scripts/script.sh
```diff
@@ -15,12 +15,14 @@ CC_SRC_LANGUAGE="$3"
 TIMEOUT="$4"
 VERBOSE="$5"
 NO_CHAINCODE="$6"
+NAME="$7"
```
lehors: I think CC_NAME would be more descriptive and consistent with other variables like CC_SRC_LANGUAGE and while this adds an argument to this script specifying that argument requires editing byfn.sh, doesn't it?
On testing I noticed that a simple addition of that argument to byfn.sh doesn't even work because NO_CHAINCODE isn't set and the cc name argument is gobbled up as the NO_CHAINCODE argument.

denyeart:
Since byfn (first-network) is being retired soon, let's go ahead and retire this PR.

PR #141, opened 2020-03-27
test-network chaincode deployment script: init function does not exist on fabcar

woodyjon:
_initLedger_ must be called instead of _init_, that does not exist.
That also keeps consistency between the "with TLS" and "withtout" TLS cases.

Signed-off-by: Jonathan Gross <woodyjon@gmail.com>


woodyjon:
some discussion already happened here: https://github.com/hyperledger/fabric-samples/pull/123
This is a new PR because I messed up the other one.

PR #147, opened 2020-03-31
Fix Fabric Binary Pull Logic

btl5037:

### DIFF
--- a/ci/install-fabric.yml
+++ b/ci/install-fabric.yml
@@ -5,17 +5,15 @@
 steps:
   - script: |
       set -eo pipefail
-      wget -q -P /tmp https://hyperledger.jfrog.io/hyperledger/fabric-binaries/hyperledger-fabric-linux-amd64-1.4.5-stable.tar.gz
-      sudo tar xzvf /tmp/hyperledger-fabric-linux-amd64-1.4.5-stable.tar.gz -C /usr/local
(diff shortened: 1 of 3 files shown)


ci/install-fabric.yml
```diff
@@ -5,17 +5,15 @@
 steps:
   - script: |
       set -eo pipefail
-      wget -q -P /tmp https://hyperledger.jfrog.io/hyperledger/fabric-binaries/hyperledger-fabric-linux-amd64-1.4.5-stable.tar.gz
-      sudo tar xzvf /tmp/hyperledger-fabric-linux-amd64-1.4.5-stable.tar.gz -C /usr/local
+      curl -L --retry 5 --retry-delay 3 https://hyperledger.jfrog.io/hyperledger/fabric-binaries/hyperledger-fabric-linux-amd64-1.4-stable.tar.gz | tar xz
```
C0rWin: I always wondered why are we keep pulling these binaries rather than just using fabric-tools container, i.e. it already contains all required binaries and we can utlize its tag to specify release.

btl5037: I don't disagree, but someone needs to do the work of modifying all of our samples. Many of them use relative paths to the binaries, not the system `PATH` so its a pain. I definitely think there is work to be done in this area, but not something I'm going to take up myself.

lehors: I agree with @btl5037, although a valid point, this is more than what's on the table right now so I will merge this and a ticket can be opened to deal with the bigger issue.

PR #153, opened 2020-04-05
Documentation doesn't match sample files

TheBookThief:
I don't know what is more accurate. I just want to point at this mismatch.
https://hyperledger-fabric.readthedocs.io/en/release-2.0/developapps/application.html#network-channel


TheBookThief:
Sorry for this. I'm going to let know the guys of the documentation that they should check the missmatches here:
https://github.com/hyperledger/fabric/blame/master/docs/source/developapps/application.md#L219
https://github.com/hyperledger/fabric/blame/master/docs/source/developapps/application.md#L246

PR #154, opened 2020-04-08
upgraded abstore golang chaincode to use contract-api package

Zzocker:
upgraded abstore golang chaincode to use [contract-api](https://github.com/hyperledger/fabric-contract-api-go) package.

Signed-off-by: Pritam Singh <pkspritam10@gmail.com>


chaincode/abstore/go/abstore.go
```diff
@@ -17,177 +17,119 @@ limitations under the License.
...
 	if Avalbytes == nil {
 		jsonResp := "{\"Error\":\"Nil amount for " + A + "\"}"
-		return shim.Error(jsonResp)
+		return "", errors.New(jsonResp)
 	}
 
 	jsonResp := "{\"Name\":\"" + A + "\",\"Amount\":\"" + string(Avalbytes) + "\"}"
 	fmt.Printf("Query Response:%s\n", jsonResp)
-	return shim.Success(Avalbytes)
+	return jsonResp, nil
```
lehors, line 124: Instead of jsonResp this needs to return string(Avalbytes)

chaincode/abstore/go/abstore.go
```diff
@@ -17,177 +17,119 @@ limitations under the License.
...
 }
 
 func main() {
-	err := shim.Start(new(ABstore))
-	if err != nil {
-		fmt.Printf("Error starting ABstore chaincode: %s", err)
+	cc,err:=contractapi.NewChaincode(new(ABstore))
+	if err!=nil{
```
lehors, line 129: Please, add the missing spaces (we use gofmt):
```
       cc, err := contractapi.NewChaincode(new(ABstore))
       if err != nil {
```

chaincode/abstore/go/go.mod
```diff
@@ -1,12 +1,5 @@
-module github.com/hyperledger/fabric-samples/chaincode/abstore/go
+module github.com/Zzocker/fabric-samples/chaincode/abstore/go
```
lehors, line 1: This line ought not to change.

chaincode/abstore/go/go.mod
```diff
@@ -1,12 +1,5 @@
...
 
-go 1.12
+go 1.14
```
lehors, line 3: We need to keep this at 1.13 max for now for the chaincode to install successfully. Otherwise we run into the following error:

```
Error: chaincode install failed with status: 500 - failed to invoke backing implementation of 'InstallChaincode': could not build chaincode: docker build failed: docker image build failed: docker build failed: Error returned from build: 1 "go: inconsistent vendoring in /chaincode/input/src/github.com/hyperledger/fabric-samples/chaincode/abstore/go:
	github.com/hyperledger/fabric-contract-api-go@v1.0.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt

run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory
```
This is because we currently don't package vendor/modules.txt with the chaincode. This is the same issue as described here: https://github.com/golang/go/issues/37948

So until we fix our chaincode packaging we need to avoid the consistency check by sticking to the go 1.13 spec in go.mod.

chaincode/abstore/go/abstore.go
Zzocker: resolved the issue

chaincode/abstore/go/abstore.go
Zzocker: did the go fmt.

chaincode/abstore/go/go.mod
Zzocker: resolved the issue

chaincode/abstore/go/go.mod
Zzocker: thanks for linking the go package error.
resolved the issue

PR #155, opened 2020-04-09
Remove TLS = false if statement in test network

nikhil550:

The create Channel and deploy chaincode scripts contained an if statement on whether TLS was enabled for many commands
```
  if [ -z "$CORE_PEER_TLS_ENABLED" -o "$CORE_PEER_TLS_ENABLED" = "false" ] ; then
```
I assume this is left over from older versions of 1.x, and from Kafka or solo consensus. Since TLS is required for RAFT and other features, I removed the TLS not enabled option.

### DIFF
--- a/test-network/scripts/createChannel.sh
+++ b/test-network/scripts/createChannel.sh
@@ -50,23 +50,15 @@ createAncorPeerTx() {
 
 createChannel() {
 	setGlobals 1
-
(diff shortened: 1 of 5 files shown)


nikhil550:
> TLS_ENABLED

Good catch. Removed these a well. Removed the `parsePeerConnectionParameters` function from the Org3 script since it is not used.

PR #160, opened 2020-04-16
making commercial-paper application-java to fit hyperledger-fabric-2.0.0+

charlielin:
modify digibank/application and magnetocorp/application to fit 2.0.0


charlielin:
Thanks for your working.
About the reslove error, I suggrest  adding `orderer.example.org` into hosts.

PR #164, opened 2020-04-20
Fix Fabcar java dependencies

lehors:

### DIFF
--- a/fabcar/java/pom.xml
+++ b/fabcar/java/pom.xml
@@ -28,7 +28,7 @@
       <dependency>
          <groupId>org.hyperledger.fabric</groupId>
          <artifactId>fabric-gateway-java</artifactId>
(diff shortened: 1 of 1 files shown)


fabcar/java/pom.xml
```diff
@@ -28,7 +28,7 @@
       <dependency>
          <groupId>org.hyperledger.fabric</groupId>
          <artifactId>fabric-gateway-java</artifactId>
-         <version>1.4.1-SNAPSHOT</version>
+         <version>1.4.3</version>
```
denyeart, line 31: why would master use 1.4.3 rather than fabric-gateway-java 2.0.0?

lehors: Well, we'll need to update the code for that upgrade. I just want to fix the CI for now. :-)

PR #167, opened 2020-04-23
FGJ-84 Update Fabcar Java app to 2.1.1

lehors:
This updates the fabcar java application to work with the fabric
gateway java version 2.1.1.
It also changes pom.xml to produce a jar file that is self contained
for ease of use.

### DIFF
--- a/fabcar/java/src/main/java/org/example/RegisterUser.java
+++ b/fabcar/java/src/main/java/org/example/RegisterUser.java
@@ -10,7 +10,10 @@ import java.util.Properties;
 import java.util.Set;
 
 import org.hyperledger.fabric.gateway.Wallet;
-import org.hyperledger.fabric.gateway.Wallet.Identity;
+import org.hyperledger.fabric.gateway.Wallets;
+import org.hyperledger.fabric.gateway.Identities;
+import org.hyperledger.fabric.gateway.Identity;
+import org.hyperledger.fabric.gateway.X509Identity;
 import org.hyperledger.fabric.sdk.Enrollment;
 import org.hyperledger.fabric.sdk.User;
 import org.hyperledger.fabric.sdk.security.CryptoSuite;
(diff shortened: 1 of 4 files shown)


heatherlp:
Thanks @lehors I'm going to defer to @bestbeforetoday on this one, he implemented the new wallet work into gateway-java

fabcar/java/src/main/java/org/example/RegisterUser.java
```diff
@@ -36,22 +39,20 @@ public static void main(String[] args) throws Exception {
 		caClient.setCryptoSuite(cryptoSuite);
 
 		// Create a wallet for managing identities
-		Wallet wallet = Wallet.createFileSystemWallet(Paths.get("wallet"));
+		Wallet wallet = Wallets.newFileSystemWallet(Paths.get("wallet"));
 
 		// Check to see if we've already enrolled the user.
-		boolean userExists = wallet.exists("appUser");
-		if (userExists) {
+		Identity user = wallet.get("appUser");
```
bestbeforetoday, line 45: Since we don't use the value returned here anywhere else, I wonder if it would be clearer to combine this line and the one below rather than the (potentially confusing) use of the `user` variable here and again further down in the code for different values? For example:
```
if (wallet.get("appUser") != null) {
```

lehors: Ok, and I've changed EnrollAdmin the same way.

lehors:
> One code style / clarity comment, and I notice that there is inconsistent indentation with a mix of spaces and tabs, but the code changes to align with the new wallet API look fundamentally good to me

Thank you checking on my changes! I fixed the indentation and addressed the code style issue.

mbwhite:
@lehors  we release 2.1.1 this week... fixed a perf issue with query

lehors:
> @lehors we release 2.1.1 this week... fixed a perf issue with query

Ok, I updated to 2.1.1.
Thanks.

dvrajitha:
is it is correct to have the following line in RegisterUser.java??
Identity user = Identities.newX509Identity("CRCDMSP", adminIdentity.getCertificate(), adminIdentity.getPrivateKey());
The enrollment is not used in above line
Enrollment enrollment = caClient.enroll("appUser", enrollmentSecret);

PR #174, opened 2020-05-05
Add secure asset transfer (marbles transfer) example chaincode

nikhil550:

Add secure asset transfer example to highlight the use of access control, private data, and state based endorsement to establish ownership of an asset, and how to securely transfer ownership of that asset.


chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	// Set the endorsement policy such that an owner org peer is required to endorse future updates
+	err = setMarbleStateBasedEndorsement(ctx, marble.ID, clientOrgID)
+	if err != nil {
+		return fmt.Errorf("failed setting state based endorsement for owner: %s", err.Error())
+	}
+
+	// Persist private immutable marble properties to owner's private data collection
+	collection := "_implicit_org_" + clientOrgID
```
sstone1, line 97: Not your fault, but this `_implicit_org_` thing looks like a pretty ugly hack. I wonder if we should make a tactical fix to the stub APIs to add a `GetImplicitCollectionName` method so we can hide it behind the scenes where it belongs?

nikhil550:
> Is the `test-network/marblesp.tar.gz` file meant to be part of this PR?

Gone.

PR #178, opened 2020-05-07
Add CA version argument, CA version does not match fabric version, so it must use another argument

duanmeng1992:
Signed-off-by: duanmeng <duanmeng@cmss.chinamobile.com>


test-network/network.sh
```diff
@@ -37,6 +37,7 @@ function printHelp() {
   echo "    -l <language> - the programming language of the chaincode to deploy: go (default), java, javascript, typescript"
   echo "    -v <version>  - chaincode version. Must be a round number, 1, 2, 3, etc"
   echo "    -i <imagetag> - the tag to be used to launch the network (defaults to \"latest\")"
+  echo "    -cai <CA_imagetag> - the image tag to be used for CA (defaults to \"latest\")"
```
lehors, line 40: Instead of saying defaults to "latest" I think this should read defaults to "$CA_IMAGETAG" to match what it is actually set to by defaults.

duanmeng1992: Yes, I agree with you. I'll revise it

PR #191, opened 2020-05-26
FAB-17930: Update licenses

ryjones:


lehors:
I can't believe we still have some of those lying around...

hacera-jonathan:
FabCars? :-)

PR #192, opened 2020-05-27
FABG-935 GoSDK Fabcar sample

andrew-coleman:
Fabcar sample demonstrating the use of the new Gateway package in the Go SDK

Signed-off-by: andrew-coleman <andrew_coleman@uk.ibm.com>


khow11:
in the "ccp-template.yaml" file:
--------------------------------------------------------------
peers:
  peer0.org${ORG}.example.com:
    url: grpcs://localhost:${P0PORT}
---------------------------------------------------------------

Can I set the url to "url: grpcs://peer0.org${ORG}.example.com:${P0PORT}" or "url: grpcs://ip-address:${P0PORT}"?

My problem is that I used the default setting of ccp-template.yaml, I deployed test-network across two network domains using 6 servers. When I commit a chaincode definition, it always tells **"failed to retrieve endorser client for commit: endorser client failed to connect to peer0.org2.example.com:7051"** .

Because the commitChaincodeDefinition relies on the Majority endorsement policy, so the commit command need two hosts (in my case, **peer0.org1.example.com** and **peer0.org2.example.com**) when I execute **commitChaincodeDefinition 1 2** on host **peer0.org1.example.com**.

I guess this is because I use the default setting of ccp-template.yaml.

Looking forward to your reply!!!

Thanks!

PR #193, opened 2020-05-29
Remove file transfer files from commercial paper

nikhil550:

Remove artifacts that were necessary for the transfer file, but not needed now that the tutorials have been updated to enroll with the CA.


nikhil550:
@mbwhite For the node application, the user can use [enrollUser.js](https://github.com/hyperledger/fabric-samples/blob/master/commercial-paper/organization/magnetocorp/application/enrollUser.js) to enroll their application, rather than use the addToWallet. I updated the readme and the tests as well. I kept the lines that copied the certificates into the locations that can be used by the addToWallet program so that the Java application could still use them.

lehors:
So @mbwhite what do you think of @nikhil550's answer? I think it's a bit unfortunate that we end with different ways of doing things depending on which language is used by I suppose the same change could eventually be done to the java app. It would be nice to dispose of this PR one way or another anyway.

nikhil550:
Closing this until the java app is updated as well seems ok to me. However, at the very least, the tests should be updated to match the tutorial.

PR #194, opened 2020-06-01
Add VerifyAssetProperties to marbles_transfer example

stephyee:
Signed-off-by: Tiffany Harris <tiffany.harris@ibm.com>

Verifies asset private data matches on chain hashes 

- Add query to verifies asset properties
- Delete price records from buyer and seller private data collections
    when transfer is complete
- Add 'receipt' for asset transfer

### DIFF
--- a/chaincode/marbles_transfer/marbles_transfer.go
+++ b/chaincode/marbles_transfer/marbles_transfer.go
@@ -24,6 +24,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/hyperledger/fabric-chaincode-go/pkg/statebased"
 	"github.com/hyperledger/fabric-chaincode-go/shim"
@@ -31,8 +32,10 @@ import (
 )
 
 const (
-	typeMarbleForSale = "S"
-	typeMarbleBid     = "B"
+	typeMarbleForSale     = "S"
+	typeMarbleBid         = "B"
+	typeMarbleSaleReceipt = "SR"
+	typeMarbleBuyReceipt  = "BR"
 )
 
 type SmartContract struct {
@@ -45,6 +48,11 @@ type Marble struct {
 	ID                string `json:"marble_id"`
 	OwnerOrg          string `json:"owner_org"`
 	PublicDescription string `json:"public_description"`
+}
+
+type receipt struct {
+	price     int
+	timestamp time.Time
 }
 
 // IssueAsset creates a marble and sets it as owned by the client's org
@@ -94,7 +102,7 @@ func (s *SmartContract) IssueAsset(ctx contractapi.TransactionContextInterface, 
 	}
 
 	// Persist private immutable marble properties to owner's private data collection
-	collection := "_implicit_org_" + clientOrgID
+	collection := buildCollectionName(clientOrgID)
 	err = ctx.GetStub().PutPrivateData(collection, marble.ID, []byte(immutablePropertiesJSON))
 	if err != nil {
 		return fmt.Errorf("failed to put Marble private details: %s", err.Error())
@@ -181,7 +189,7 @@ func agreeToPrice(ctx contractapi.TransactionContextInterface, marbleID string, 
 		return fmt.Errorf("marble_price key not found in the transient map")
 	}
 
-	collection := "_implicit_org_" + clientOrgID
+	collection := buildCollectionName(clientOrgID)
 
 	// Persist the agreed to price in a collection sub-namespace based on priceType key prefix,
 	// to avoid collisions between private marble properties, sell price, and buy price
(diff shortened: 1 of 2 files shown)


chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -198,11 +199,44 @@ func agreeToPrice(ctx contractapi.TransactionContextInterface, marbleID string,
...
+	if !ok {
+		return false, fmt.Errorf("marble_properties key not found in the transient map")
+	}
+
+	marble, err := s.GetAsset(ctx, marbleID)
+	if err != nil {
+		return false, fmt.Errorf("failed to get marble: %s", err.Error())
+	}
+
+	collectionSeller := "_implicit_org_" + marble.OwnerOrg
```
denyeart, line 220: It's technically not a sale yet... so let's call the variable collectionOwner.

And let's hide this ugly hand crafted internal implicit collection name behind a utility function that takes the owner org id (throughout). Eventually such a utility function may be included in the shim itself.

chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -361,9 +395,44 @@ func transferMarbleState(ctx contractapi.TransactionContextInterface, marble *Ma
...
+	}
+
+	err = ctx.GetStub().DelPrivateData(collectionBuyer, marblePriceKey)
+	if err != nil {
+		return fmt.Errorf("failed to delete marble price from implicit private data collection for buyer: %s", err.Error())
+	}
+
+	// Keep record for a 'receipt' in both buyer and seller private data collection to record the sales price and date
+	// Persist the agreed to price in a collection sub-namespace based on receipt key prefix
+	receipt, err := ctx.GetStub().CreateCompositeKey(typeMarbleReceipt, []string{marble.ID})
```
denyeart, line 422: It is possible that the org owns and then sells the same asset twice.
Therefore let's add one more part to the composite key to ensure the receipt is unique.
The txid is probably the best string to add to the composite key to ensure uniqueness. Additionally it will serve as a pointer to the transaction that transferred the asset in case somebody wants to explore the history.
You can get txid using shim's GetTxID().

If you think about it, you'll see that this key approach would enable somebody to query for all asset receipts, or query for the receipts for a specific asset id, depending on the partial composite key query used. Neat huh?

chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -361,9 +395,44 @@ func transferMarbleState(ctx contractapi.TransactionContextInterface, marble *Ma
...
+	if err != nil {
+		return fmt.Errorf("failed to create composite key: %s", err.Error())
+	}
+
+	err = ctx.GetStub().PutPrivateData(collectionBuyer, receipt, immutablePropertiesJSON)
```
denyeart, line 427: How about receiptKey instead of receipt?
And instead of saving the immutable properties JSON, I was thinking the sales price and sale date (GetTxTimestamp()) would get stored in a new receipt JSON.

chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -361,9 +395,44 @@ func transferMarbleState(ctx contractapi.TransactionContextInterface, marble *Ma
...
+	if err != nil {
+		return fmt.Errorf("failed to put Marble private properties for buyer: %s", err.Error())
```
denyeart, line 429: Wrong message...

chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -361,9 +395,44 @@ func transferMarbleState(ctx contractapi.TransactionContextInterface, marble *Ma
...
+	}
+
+	err = ctx.GetStub().PutPrivateData(collectionSeller, receipt, immutablePropertiesJSON)
```
denyeart, line 432: Same comments as buyer...

chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -198,11 +199,44 @@ func agreeToPrice(ctx contractapi.TransactionContextInterface, marbleID string,
 	return nil
 }
 
-// TODO implement function to verify marble properties
-// For example, Org1 may tell Org2 about the properties and salt.
-// Org2 would want to verify the properties before agreeing to buy.
-// Org2 would call a verify function on his peer.
-// The properties and salt would passed in, get hashed in the chaincode, and compared with the on-chain hash of the marble properties (queried via GetPrivateDataHash).
+// VerifyAssetProperties implement function to verify marble properties
```
denyeart, line 202: Since this is a learning sample, a more descriptive comment would be good, explaining that the hash is verified.

Also, for each new function like this, add a call in the readme showing how it would be used.

chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -33,6 +33,7 @@ import (
 const (
 	typeMarbleForSale = "S"
 	typeMarbleBid     = "B"
+	typeMarbleReceipt = "R"
```
denyeart, line 36: Let's use a different prefix for sales receipt versus buy receipt.

stephyee: How about `BR` and `SR`?

chaincode/marbles_transfer/marbles_transfer.go
stephyee: 📚

chaincode/marbles_transfer/marbles_transfer.go
stephyee: 🛠️

chaincode/marbles_transfer/marbles_transfer.go
stephyee: 🛠️

chaincode/marbles_transfer/marbles_transfer.go
stephyee: I like this. 🛠️

chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -348,22 +391,88 @@ func transferMarbleState(ctx contractapi.TransactionContextInterface, marble *Ma
...
+	}
+
+	err = ctx.GetStub().DelPrivateData(collectionBuyer, marblePriceKey)
+	if err != nil {
+		return fmt.Errorf("failed to delete marble price from implicit private data collection for buyer: %s", err.Error())
+	}
+
+	// Keep record for a 'receipt' in both buyer and seller private data collection to record the sales price and date
+	// Persist the agreed to price in a collection sub-namespace based on receipt key prefix
+	receiptBuyKey, err := ctx.GetStub().CreateCompositeKey(typeMarbleBuyReceipt, []string{ctx.GetStub().GetTxID(), marble.ID})
```
denyeart, line 430: Good, but I was thinking you should put marble.id before txid. This would enable partial composite key queries by marble id in the future. For example, you may want to find all sales receipts for "marble1".
In general, the 'logical' part of the key should be first, and the uniqueifying part of the key should be last.

chaincode/marbles_transfer/marbles_transfer.go
```diff
@@ -348,22 +391,88 @@ func transferMarbleState(ctx contractapi.TransactionContextInterface, marble *Ma
...
+	if err != nil {
+		return fmt.Errorf("failed to create composite key for receipt: %s", err.Error())
+	}
+
+	// Create the receipt
+	agreementJSON, err := getAssetPrice(ctx, marble.ID, typeMarbleForSale)
```
denyeart, line 436: This will retrieve the price json from the ledger. That seems like an unnecessary ledger hit, since the price json was already passed into TransferAsset(). I'd recommend simply pass it into transferMarbleState() instead of looking up from ledger again.

chaincode/marbles_transfer/marbles_transfer.go
stephyee: 🛠️

chaincode/marbles_transfer/marbles_transfer.go
stephyee: 🛠️

PR #195, opened 2020-06-02
[FAB-17952] Fix test-network to use the latest version of fabric-ca

satota2:
To get rid of hard-coded versioning and use the latest version of fabric-ca in test-network, this patch modifies the existence check logic of the fabric-ca-client binary in the same way as peer and orderer.

Also, this patch updates addOrg3.sh to use cai <ca_imagetag> option.

#### Type of change

- Bug fix

#### Description

Currently, test-network mainly uses the previous version (1.4.6) of fabric-ca.

Also, addOrg3.sh in test-network is not using the cai <ca_imagetag> option (This may use non-existent version or a hard-coded 1.4.4 version).

So, this should catch up with the latest version (latest or 1.4.7).

#### Related issues

- https://jira.hyperledger.org/browse/FAB-17952


nikhil550:
Lovely update. Thank you for this.

satota2:
@lehors @nikhil550 Thank you for your reviews.

lehors:
Thank you for your contribution!

PR #200, opened 2020-06-08
Optimize query conditions in fabcar chaincode

zhuzeyu:
optimize query conditions in fabcar chaincode

Signed-off-by: zhuzeyu <zhuzeyu0409@gmail.com>

### DIFF
--- a/chaincode/fabcar/external/fabcar.go
+++ b/chaincode/fabcar/external/fabcar.go
@@ -22,16 +22,16 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"strconv"
 	"os"
+	"strconv"
 
 	"github.com/hyperledger/fabric-chaincode-go/shim"
 	"github.com/hyperledger/fabric-contract-api-go/contractapi"
 )
 
 type ServerConfig struct {
-	CCID     string
-	Address  string
+	CCID    string
+	Address string
 }
 
 // SmartContract provides functions for managing a car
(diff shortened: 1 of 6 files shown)


zhuzeyu:
> Ok, could you please make the same change to chaincode/fabcar/external/fabcar.go so we have them all the same?

Done

chaincode/fabcar/external/fabcar.go
```diff
@@ -115,7 +115,7 @@ func (s *SmartContract) QueryCar(ctx contractapi.TransactionContextInterface, ca
 // QueryAllCars returns all cars found in world state
 func (s *SmartContract) QueryAllCars(ctx contractapi.TransactionContextInterface) ([]QueryResult, error) {
 	startKey := "CAR0"
-	endKey := "CAR99"
+	endKey := "CAR999"
```
denyeart, line 118: Sorry for the late review, but actually the startKey and endKey approach in Fabcar has always been wrong.
There is no need to create artificial startKey and endKey like this at all.
To do an open-ended query, empty string should be used for both startKey and endKey. This approach entirely avoids the problem of trying to decide which keys to use at the extreme ends.
It would be great if we could use the empty string approach in all chaincode languages.

lehors: Ah, good point. Thanks Dave!

zhuzeyu: > Sorry for the late review, but actually the startKey and endKey approach in Fabcar has always been wrong.
> There is no need to create artificial startKey and endKey like this at all.
> To do an open-ended query, empty string should be used for both startKey and endKey. This approach entirely avoids the problem of trying to decide which keys to use at the extreme ends.
> It would be great if we could use the empty string approach in all chaincode languages.

Thanks for review, your suggestion is right, I modify the query condition to empty string already.

sijocherian:
lgtm

PR #207, opened 2020-06-12
Add init function to avoid init function not found error

SamYuan1990:
Hi 

Recently I followed https://hyperledger-fabric.readthedocs.io/en/release-2.1/private_data_tutorial.html#commit-the-chaincode-definition
with master branch of fabric-sample.

An error happen when I attempted to init the chaincode with function init
```
 peer chaincode invoke -o localhost:7050 --ordererTLSHostnameOverride orderer.example.com --channelID mychannel --name marblesp --isInit --tls --cafile $ORDERER_CA --peerAddresses localhost:7051 --tlsRootCertFiles $ORG1_CA -c '{"Args":["Init"]}'
```
```
Error: endorsement failure during invoke. response: status:500 message:"error in simulation: transaction returned with failure: Function Init not found in contract SmartContract" 
```

Here is the fix for this with adding `function Init`, please verify.

### DIFF
--- a/chaincode/marbles02_private/go/marbles_chaincode_private.go
+++ b/chaincode/marbles02_private/go/marbles_chaincode_private.go
@@ -119,6 +119,9 @@ type SmartContract struct {
 	contractapi.Contract
 }
 
+func (s *SmartContract) Init(ctx contractapi.TransactionContextInterface) error {
+	return nil
+}
 
 // ============================================================
 // initMarble - create a new marble, store into chaincode state
@@ -217,7 +220,7 @@ func (s *SmartContract) InitMarble(ctx contractapi.TransactionContextInterface) 
 	indexName := "color~name"
 	colorNameIndexKey, err := ctx.GetStub().CreateCompositeKey(indexName, []string{marble.Color, marble.Name})
 	if err != nil {
-			return err
+		return err
 	}
 	//  Save index entry to state. Only the key name is needed, no need to store a duplicate copy of the marble.
 	//  Note - passing a 'nil' value will effectively delete the key from state, therefore we pass null character as value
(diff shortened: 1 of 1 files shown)


davidkhala:
If the definition of the chaincode is not init_required, then there is no need to call init function before regular usage. 

While I have not yet check the tutorial setting

SamYuan1990:
> If the definition of the chaincode is not `init_required`, then there is no need to call init function before regular usage.
> 
> I see in tutorial `--init-required` is used so Init function should be added.

So, I suppose either update the code or the tutorial,

as there is `--init-required` https://hyperledger-fabric.readthedocs.io/en/release-2.1/private_data_tutorial.html#commit-the-chaincode-definition
```
peer lifecycle chaincode commit -o localhost:7050 --ordererTLSHostnameOverride orderer.example.com --channelID mychannel --name marblesp --version 1.0 --sequence 1 --collections-config ../chaincode/marbles02_private/collections_config.json --signature-policy "OR('Org1MSP.member','Org2MSP.member')" --init-required --tls --cafile $ORDERER_CA --peerAddresses localhost:7051 --tlsRootCertFiles $ORG1_CA --peerAddresses localhost:9051 --tlsRootCertFiles $ORG2_CA
```

nikhil550:
Hi @SamYuan1990, thank you for finding this. The code was updated, but the `--init-required` was left in the tutorial. Since we are trying to move folks away from Init, I would rather update the tutorial rather than the code.

SamYuan1990:
> Hi @SamYuan1990, thank you for finding this. The code was updated, but the `--init-required` was left in the tutorial. Since we are trying to move folks away from Init, I would rather update the tutorial rather than the code.

Either fine for me. Pls let me know for pr of the tutorial upgrade. thx.

SamYuan1990:
> Hi @SamYuan1990, thank you for finding this. The code was updated, but the `--init-required` was left in the tutorial. Since we are trying to move folks away from Init, I would rather update the tutorial rather than the code.

https://github.com/hyperledger/fabric/pull/1405

mbwhite:
@nikhil550  I'm happy with this.. is ok from your perspective to merge?

nikhil550:
@mbwhite Dave and I have actually been trying to move people away from the Init behavior. The tutorial actually contains the changes required to run the chaincode as is in latest. However the cherry-picks of those changes into 2.1 and 2.0 are being held up by a CI problem:
https://github.com/hyperledger/fabric/pull/1402
https://github.com/hyperledger/fabric/pull/1401

Once those are merged, all will be good.

PR #208, opened 2020-06-15
Adding chaincode-js to asset-transfer-basic sample.

DereckLuo:
Signed-off-by: Chongxin Luo <Chongxin.Luo@ibm.com>

**Description**

* Adding javascript chaincode to `asset-transfer-basic` sample.

### DIFF
--- a/asset-transfer-basic/chaincode-javascript/.gitignore
+++ b/asset-transfer-basic/chaincode-javascript/.gitignore
@@ -1,0 +1,12 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+# Coverage directory used by tools like istanbul
+coverage
+
+# Dependency directories
+node_modules/
+jspm_packages/
+
+package-lock.json
--- a/asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
+++ b/asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
@@ -1,0 +1,159 @@
+/*
+ * Copyright IBM Corp. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict';
+
+const { Contract } = require('fabric-contract-api');
+
+class AssetTransfer extends Contract {
+
+    async initLedger(ctx) {
+        const assets = [
+            {
+                ID: 'asset1',
+                Color: 'blue',
+                Size: 5,
+                Owner: 'Tomoko',
+                AppraisedValue: 300,
+            },
+            {
+                ID: 'asset2',
+                Color: 'red',
+                Size: 5,
+                Owner: 'Brad',
+                AppraisedValue: 400,
+            },
+            {
+                ID: 'asset3',
+                Color: 'green',
+                Size: 10,
+                Owner: 'Jin Soo',
+                AppraisedValue: 500,
+            },
+            {
+                ID: 'asset4',
+                Color: 'yellow',
+                Size: 10,
+                Owner: 'Max',
+                AppraisedValue: 600,
+            },
+            {
+                ID: 'asset5',
+                Color: 'black',
+                Size: 15,
+                Owner: 'Adriana',
+                AppraisedValue: 700,
+            },
+            {
+                ID: 'asset6',
+                Color: 'white',
+                Size: 15,
+                Owner: 'Michel',
+                AppraisedValue: 800,
+            },
+        ];
+
+        for (let i = 0; i < assets.length; i++) {
+            assets[i].docType = 'asset';
+            await ctx.stub.putState(assets[i].ID, Buffer.from(JSON.stringify(assets[i])));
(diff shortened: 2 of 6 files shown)


asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
```diff
@@ -0,0 +1,159 @@
...
+        asset.Owner = newOwner;
+
+        await ctx.stub.putState(id, Buffer.from(JSON.stringify(asset)));
+    }
+
+    // getAllAssets returns all assets found in the world state.
+    async getAllAssets(ctx) {
+        const allResults = [];
+        // range query with empty string for startKey and endKey does an open-ended query of all assets in the chaincode namespace.
+        for await (const { key, value } of ctx.stub.getStateByRange("", "")) {
```
mbwhite, line 141: Just to confirm this range - does that work>

asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
```diff
@@ -0,0 +1,159 @@
...
+        let assetString = await this.readAsset(ctx, id);
+
+        let asset = JSON.parse(assetString);
+        asset.Owner = newOwner;
+
+        await ctx.stub.putState(id, Buffer.from(JSON.stringify(asset)));
+    }
+
+    // getAllAssets returns all assets found in the world state.
+    async getAllAssets(ctx) {
```
mbwhite, line 138: Suggestion:  Would suggest we don't include a 'getAll'  but rather always have it scoped down. 

getAll is fine when there are 10 records, but in reality it's going to be range query otherwise it'll take too long.

asset-transfer-basic/chaincode-javascript/.gitignore
```diff
@@ -0,0 +1,77 @@
+#
```
mbwhite, line 1: This is the standard .gitignore file .. but most is redundant for our needs.. perhaps worth slimming down.

asset-transfer-basic/chaincode-javascript/.editorconfig
```diff
@@ -0,0 +1,16 @@
+#
```
mbwhite, line 1: As this is specific to an editor, probably best not to include this.

asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
DereckLuo: confirmed with application-javascript and it works. :)

asset-transfer-basic/chaincode-javascript/.gitignore
DereckLuo: done.

asset-transfer-basic/chaincode-javascript/.editorconfig
DereckLuo: done.

asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
DereckLuo: This is a good point. 
Just currently `getAllAssets` are being used in multiple scripts. We can probably change this to get top 10 assets by assetID or something else.

lehors: OK but let's not have a built-in range either. I think the chaincode needs to allow retrieving a range specified by the app and the app needs to be written to allow specifying that range via command line arguments. I dislike the fact that fabcar has this built-in and hope we can do better this time around!

denyeart: Yeah, Fabcar used to hardcode an arbitrary startKey and endKey, we just got that fixed to query the full open-ended range. I'd suggest since these samples are small to keep getAllAssets as full open-ended range query. And I think we can address Matthew's point by putting a comment in the code and tutorial that you can also choose to do scoped queries or paginated queries in production chaincodes (paginated queries are demonstrated in the more advanced ledger-queries variant of the sample chaincode).

asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
denyeart: Not only does it work, it is the recommended way to do open-ended range queries in the shim doc :-)

asset-transfer-basic/chaincode-javascript/package-lock.json
```diff
@@ -0,0 +1,3246 @@
+{
```
denyeart, line 1: I've always assumed package-lock.json would be generated when user does 'npm install', rather than checking it in. Is there any advantage to checking it in?

DereckLuo: done. removed it from commit, and added it to .gitignore

PR #216, opened 2020-06-30
[Test] Using the Fabric test network

YaraiLee:


denyeart:
Not sure what this PR is... please add Description if there is something intended to change.

PR #217, opened 2020-06-30
fix up for test network

SamYuan1990:
replace duplicate peer version by which peer
remove un used config folder check

### DIFF
--- a/test-network/network.sh
+++ b/test-network/network.sh
@@ -88,10 +88,10 @@ BLACKLISTED_VERSIONS="^1\.0\. ^1\.1\. ^1\.2\. ^1\.3\. ^1\.4\."
 # of go or other items could be added.
 function checkPrereqs() {
   ## Check if your have cloned the peer binaries and configuration files.
-  peer version > /dev/null 2>&1
+  which peer
 
-  if [[ $? -ne 0 || ! -d "../config" ]]; then
-    echo "ERROR! Peer binary and configuration files not found.."
+  if [ "$?" -ne 0 ]; then
+    echo "ERROR! Peer binary not found..."
     echo
     echo "Follow the instructions in the Fabric docs to install the Fabric Binaries:"
     echo "https://hyperledger-fabric.readthedocs.io/en/latest/install.html"


SamYuan1990:
btw, ask for reviewer, I am also considering replace 
`  LOCAL_VERSION=$(peer version | sed -ne 's/ Version: //p')`
by
`  LOCAL_VERSION=$(cryptogen version | sed -ne 's/ Version: //p')`

as for my understanding, that in test network, we using `cryptogen` for peer, order container MSP right?
in a developer point of view, `peer version` doesn't means local `cryptogen` version.

test-network/network.sh
```diff
@@ -88,9 +88,9 @@ BLACKLISTED_VERSIONS="^1\.0\. ^1\.1\. ^1\.2\. ^1\.3\. ^1\.4\."
 # of go or other items could be added.
 function checkPrereqs() {
   ## Check if your have cloned the peer binaries and configuration files.
-  peer version > /dev/null 2>&1
+  which peer
 
-  if [[ $? -ne 0 || ! -d "../config" ]]; then
+  if [ "$?" -ne 0 ]; then
     echo "ERROR! Peer binary and configuration files not found.."
```
guoger, line 94: this should be updated as well to exclude `configuration files`, right?

SamYuan1990: right. will fix up.

SamYuan1990:
> Test network does in fact use config directory as the basis for FABRIC_CFG_PATH, so that check needs to remain.
> 
> I don't understand the motivation for this change. What's wrong with using "peer version"?

well, question on 
1. the dependence with FABRIC_CFG_PATH?
2. FABRIC_CFG_PATH doesn't mean `../config` right?

the reason for using `cryptogen` and `configtxgen`
`cryptogen` generate msp for peer and orderer, if miss match on version, peer container fails start.
`configtxgen` generate genesis block, if miss match on version, order/channel fails.

guoger:
indead... [here](https://github.com/hyperledger/fabric-samples/blob/adfef198df2a416629c6375e7632b553433456d2/test-network/scripts/createChannel.sh#L133) we need to config `FABRIC_CFG_PATH` for `peer` to execute

SamYuan1990:
> indead... [here](https://github.com/hyperledger/fabric-samples/blob/adfef198df2a416629c6375e7632b553433456d2/test-network/scripts/createChannel.sh#L133) we need to config `FABRIC_CFG_PATH` for `peer` to execute

k...
close this pr, thanks @guoger , @denyeart

PR #218, opened 2020-06-30
Fix fabcar Java chaincode test & coverage failures

jt-nti:

### DIFF
--- a/chaincode/fabcar/java/src/test/java/org/hyperledger/fabric/samples/fabcar/CarQueryResultTest.java
+++ b/chaincode/fabcar/java/src/test/java/org/hyperledger/fabric/samples/fabcar/CarQueryResultTest.java
@@ -1,0 +1,93 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.hyperledger.fabric.samples.fabcar;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+public final class CarQueryResultTest {
+
+    @Nested
+    class Equality {
+
+        @Test
+        public void isReflexive() {
+            CarQueryResult cqr = new CarQueryResult("CAR1", new Car("Toyota", "Prius", "blue", "Tomoko"));
+
+            assertThat(cqr).isEqualTo(cqr);
+        }
+
+        @Test
+        public void isSymmetric() {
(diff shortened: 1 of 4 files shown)


chaincode/fabcar/java/src/test/java/org/hyperledger/fabric/samples/fabcar/CarQueryResultTest.java
```diff
@@ -0,0 +1,93 @@
...
+        @Test
+        public void handlesNull() {
+            CarQueryResult cqr = new CarQueryResult("CAR1", new Car("Toyota", "Prius", "blue", "Tomoko"));
+
+            assertThat(cqr).isNotEqualTo(null);
+        }
+    }
+
+    @Test
+    public void toStringIdentifiesCarQueryResult() {
```
sijocherian, line 87: is a test on toString() really necessary?  specially asserting on the hash code

toString() is meant for display/logging. If you agree, then let us just remove it from CarTest as well.

chaincode/fabcar/java/src/test/java/org/hyperledger/fabric/samples/fabcar/FabCarTest.java
```diff
@@ -239,7 +239,7 @@ public void whenCarExists() {
 
             Car car = contract.changeCarOwner(ctx, "CAR0", "Dr Evil");
 
-            assertThat(car).isEqualTo(new CarQueryResult("CAR0", new Car("Toyota", "Prius", "blue", "Dr Evil")));
+            assertThat(car).isEqualTo(new Car("Toyota", "Prius", "blue", "Dr Evil"));
```
sijocherian, line 242: Thanks for fixing

Rest all Looks good !

chaincode/fabcar/java/src/test/java/org/hyperledger/fabric/samples/fabcar/CarQueryResultTest.java
mbwhite: I would disagree; toString() is part of the external API, so it's proper to have tests for it.  

There's nothing worth than debugging a log from a production system only to find you don't have the correct information.

jt-nti: It was also required for the coverage check to pass- it's a small sample so it seems like 100% coverage is reasonable to aspire to

PR #224, opened 2020-07-02
Update tests for commercial paper

nikhil550:
Update the commercial paper tests to use the test network certificate authorities and the `enrollUser.js` file, so we can test the same path used by the tutorial.

Also updating the go contract test, since it was testing the Java contract for some reason.


sijocherian:
lgtm

PR #229, opened 2020-07-06
Fix param sequence for javascript invoke function.

denali49:


sijocherian:
see comment at #227

denyeart:
LGTM, we'll standardize on this order:
id, color, size, owner, value.

PR #230, opened 2020-07-06
Refactor and Add Tests to FabCar Go Chaincode

btl5037:
Many chaincode developers raise questions on how to test chaincode. It used to be much easier with the old shim, as they could directly use the old mock stub in the shim. Now that it no longer exists the fabcar example can provide an example of how to test chaincode using our standard Go toolset and generated mocks.

Also worth noting is our current directory structure of our Go chaincodes prevents the creation of mocks due to import cycles. This change also pushes the chaincode logic down into a `contract` package.


btl5037:
I'd like to do an example in Node chaincode as well, and include a section in the developer doc on unit testing chaincode

chaincode/fabcar/go/contract/shim.go
```diff
@@ -0,0 +1,21 @@
+package contract
+
+import (
+	"github.com/hyperledger/fabric-chaincode-go/shim"
+	"github.com/hyperledger/fabric-contract-api-go/contractapi"
+)
+
+//go:generate counterfeiter -o mocks/transaction.go -fake-name TransactionContext . TransactionContext
```
sijocherian, line 8: This is totally new way to unit test . Thanks

I am not familiar with counterfeiter. As a developer, it will be helpful for now, to add a section/README
(in  fabcar directory somewhere) with bullet list of steps to use counterfeiter & generating mocks

btl5037:
Per a conversation with @denyeart I will do the testing on the `asset-transfer-basic` chaincode instead of the fabcar, as this is our strategic chaincode moving forward.

chaincode/fabcar/go/contract/shim.go
```diff
@@ -0,0 +1,21 @@
...
+type TransactionContext interface {
+	contractapi.TransactionContextInterface
+}
```
sijocherian, line 11: with this filename `shim.go`, I was confused.
It initially led me to think these types were used in main smart contract.

will removing this file, and moving all go:generate + types to file  `smartcontract_test.go` work or similar refactor? Similar to fabric repo tests. 

And make them non-exports of this package, will be better practice. so `type transactionContext `

chaincode/fabcar/go/go.mod
```diff
@@ -2,4 +2,11 @@ module github.com/hyperledger/fabric-samples/chaincode/fabcar/go
...
 go 1.13
 
-require github.com/hyperledger/fabric-contract-api-go v1.1.0
+require (
+	github.com/hyperledger/fabric-contract-api-go v1.1.0
+	github.com/golang/protobuf v1.3.2
+	github.com/hyperledger/fabric-chaincode-go v0.0.0-20200424173110-d7076418f212
+	github.com/hyperledger/fabric-contract-api-go v1.1.0
+	github.com/hyperledger/fabric-protos-go v0.0.0-20200424173316-dd554ba3746e
+	github.com/stretchr/testify v1.5.1
```
sijocherian, line 11: It will be helpful to  group the unit-test dependency towards the bottom and add a comment about test-dependency? 

wondering why there is no dependency on `counterfeiter` and if a `tools.go` file is needed to run test

chaincode/fabcar/go/contract/shim.go
btl5037: Correct, in the next update it was pushed into the test file itself. But then we decided to write the tests for a different chaincode so it was never pushed.

btl5037: And good call on the `unexported` interfaces

chaincode/fabcar/go/contract/shim.go
btl5037: Yes, as this was totally a spur of the moment thing without any input from the other maintainers, I decided to wait to add Doc until I gauged feedback and support. If the rest of the maintainers agree there is value in this, I plan to add a doc on unit testing in both Go and Node.

btl5037:
closing in favor of #231

chaincode/fabcar/go/contract/shim.go
btl5037: I'll also mention, this is the correct way to unit test chaincode now as I deprecated the MockStub last year that is part of the `fabric-chaincode-go` project, with the expectation that chaincode developers would use generated mocks instead.

https://jira.hyperledger.org/browse/FAB-15367
https://github.com/hyperledger/fabric/commit/dac782a804cf796721ca0f5383b9a6a8a5d83b9f

PR #232, opened 2020-07-07
run fabcar go 

Acmefocus:
add set host and env  to fabcar run go

### DIFF
--- a/fabcar/startFabric.sh
+++ b/fabcar/startFabric.sh
@@ -109,8 +109,25 @@ Go:
   Start by changing into the "go" directory:
     cd go
 
+  Next, set hosts:
+    sudo vim /etc/hosts
+
+  add Under the content into  /etc/hosts
+    ip	peer0.org1.example.com  
+    ip	peer0.org2.example.com  
+    ip	peer0.org2.example.com  
+
+    for example:
+    127.0.0.1       peer0.org1.example.com  peer0.org2.example.com  orderer.example.com
+
+  Next, set environment variable:
+    export DISCOVERY_AS_LOCALHOST=true
+
   Then, install dependencies and run the test using:
-    go run fabcar.go
+    1. export DISCOVERY_AS_LOCALHOST=true
+    2. go run fabcar.go
+  or
+    ./runfabcar.sh
 
   The test will invoke the sample client app which perform the following:
     - Import user credentials into the wallet (if they don't already exist there)
--- a/fabcar/startFabric.sh
+++ b/fabcar/startFabric.sh
@@ -109,12 +109,8 @@ Go:
   Start by changing into the "go" directory:
     cd go
 
-  Set environment variable DISCOVERY_AS_LOCALHOST to true, to indicate to ServiceDiscovery that all hostnames maps to localhost
-
   Then, install dependencies and run the test using:
     go run fabcar.go
-  or
-    ./runfabcar.sh
 
   The test will invoke the sample client app which perform the following:
     - Import user credentials into the wallet (if they don't already exist there)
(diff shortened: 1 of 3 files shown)


fabcar/startFabric.sh
```diff
@@ -109,8 +109,25 @@ Go:
...
+  Next, set hosts:
+    sudo vim /etc/hosts
+
+  add Under the content into  /etc/hosts
+    ip	peer0.org1.example.com  
+    ip	peer0.org2.example.com  
+    ip	peer0.org2.example.com  
+
+    for example:
+    127.0.0.1       peer0.org1.example.com  peer0.org2.example.com  orderer.example.com
```
sijocherian, line 121: what is the need to add this entry to /etc/hosts ?

Last time, I tried , `go run fabcar.go` works without this. ( still need `export DISCOVERY_AS_LOCALHOST=true`)

fabcar/startFabric.sh
```diff
@@ -109,8 +109,25 @@ Go:
...
+
+  Next, set environment variable:
+    export DISCOVERY_AS_LOCALHOST=true
+
   Then, install dependencies and run the test using:
-    go run fabcar.go
+    1. export DISCOVERY_AS_LOCALHOST=true
```
sijocherian, line 127: thanks for adding above step to this documentation.

My opinion is that the you made documentation in this script fairly clear, and we do not need alternate script `./runfabcar.sh`. So your instruction on setting env var is sufficient

fabcar/startFabric.sh
```diff
@@ -109,8 +109,25 @@ Go:
...
+  add Under the content into  /etc/hosts
+    ip	peer0.org1.example.com  
+    ip	peer0.org2.example.com  
+    ip	peer0.org2.example.com  
+
+    for example:
+    127.0.0.1       peer0.org1.example.com  peer0.org2.example.com  orderer.example.com
+
+  Next, set environment variable:
+    export DISCOVERY_AS_LOCALHOST=true
```
sijocherian, line 124: A description like this will inform the user, since you have a export command in line 127

```suggestion
  Set environment variable DISCOVERY_AS_LOCALHOST to true, to indicate to ServiceDiscovery that all hostnames maps to localhost
```

fabcar/startFabric.sh
matrixik: Personally I prefer to have scripts running everything like this. The first thing I did when started looking into running test-network and starting all services was to create script with all of the steps from documentation so other team members don't need to go through it again.

sijocherian:
Need input from core-maintainers on addition of this script just for the `go` chaincodes: runfabcar.sh
cc @andrew-coleman @denyeart

fabcar/startFabric.sh
sijocherian: the need for this in hosts entry , is still not clear to me. 

`127.0.0.1       peer0.org1.example.com  peer0.org2.example.com  orderer.example.com`

andrew-coleman:
I think to be consistent with the Java and JavaScript/TypeScript version of this sample, we should just hard code the setting of this directly into the application code.  I.e., put
`os.Setenv("DISCOVERY_AS_LOCALHOST", "true")`
somewhere near the top of `fabcar.go`

Acmefocus:
> I think to be consistent with the Java and JavaScript/TypeScript version of this sample, we should just hard code the setting of this directly into the application code. I.e., put
> `os.Setenv("DISCOVERY_AS_LOCALHOST", "true")`
> somewhere near the top of `fabcar.go`

good  suggestion

andrew-coleman:
@mbwhite I'm happy that you merge this :)

PR #235, opened 2020-07-07
Allow install of a chaincode

harrisob:
options on the network script to deploy chaincode
-ccn The name to be used as the deployed name and used as
     the short name of known chaincodes when the -ccp is not included.
   known short names
     'basic' - asset-transfer-basic
     'secure' - asset-transfer-secured-agreement
     'ledger' - asset-transfer-ledger-queries
     'private' - asset-transfer-private-data
-ccl the language of the chaincode
-ccv the version
-ccp [optional] the path to the chaincode, when provided
     the -ccn will be the deployed name
-cci [optional] the chaincode function to call during deployment
     that will perform an initialization of the channel state
     required for this chaincode

### DIFF
--- a/test-network/network.sh
+++ b/test-network/network.sh
@@ -20,13 +20,13 @@ export VERBOSE=false
 function printHelp() {
   echo "Usage: "
   echo "  network.sh <Mode> [Flags]"
-  echo "    <Mode>"
-  echo "      - 'up' - bring up fabric orderer and peer nodes. No channel is created"
-  echo "      - 'up createChannel' - bring up fabric network with one channel"
-  echo "      - 'createChannel' - create and join a channel after the network is created"
-  echo "      - 'deployCC' - deploy the fabcar chaincode on the channel"
-  echo "      - 'down' - clear the network with docker-compose down"
-  echo "      - 'restart' - restart the network"
+  echo "    Modes:"
+  echo "      "$'\e[0;32m'up$'\e[0m' - bring up fabric orderer and peer nodes. No channel is created
+  echo "      "$'\e[0;32m'up createChannel$'\e[0m' - bring up fabric network with one channel
+  echo "      "$'\e[0;32m'createChannel$'\e[0m' - create and join a channel after the network is created
+  echo "      "$'\e[0;32m'deployCC$'\e[0m' - deploy the asset transfer basic chaincode on the channel or specify
+  echo "      "$'\e[0;32m'down$'\e[0m' - clear the network with docker-compose down
+  echo "      "$'\e[0;32m'restart$'\e[0m' - restart the network
   echo
   echo "    Flags:"
   echo "    -ca <use CAs> -  create Certificate Authorities to generate the crypto material"
@@ -34,26 +34,31 @@ function printHelp() {
   echo "    -s <dbtype> - the database backend to use: goleveldb (default) or couchdb"
   echo "    -r <max retry> - CLI times out after certain number of attempts (defaults to 5)"
   echo "    -d <delay> - delay duration in seconds (defaults to 3)"
-  echo "    -l <language> - the programming language of the chaincode to deploy: go (default), java, javascript, typescript"
-  echo "    -v <version>  - chaincode version. Must be a round number, 1, 2, 3, etc"
+  echo "    -ccn <name> - the short name of the chaincode to deploy: basic (default),ledger, private, secured"
+  echo "    -ccl <language> - the programming language of the chaincode to deploy: go (default), java, javascript, typescript"
+  echo "    -ccv <version>  - chaincode version. 1.0 (default)"
+  echo "    -ccs <sequence>  - chaincode definition sequence. Must be an integer, 1 (default), 2, 3, etc"
+  echo "    -ccp <path>  - Optional, chaincode path. Path to the chaincode. When provided the -ccn will be used as the deployed name and not the short name of the known chaincodes."
+  echo "    -cci <fcn name>  - Optional, chaincode init required function to invoke. When provided this function will be invoked after deployment of the chaincode and will define the chaincode as initialization required."
   echo "    -i <imagetag> - the tag to be used to launch the network (defaults to \"latest\")"
   echo "    -cai <ca_imagetag> - the image tag to be used for CA (defaults to \"${CA_IMAGETAG}\")"
   echo "    -verbose - verbose mode"
-  echo "  network.sh -h (print this message)"
+  echo "    -h - print this message"
   echo
-  echo " Possible Mode and flags"
-  echo "  network.sh up -ca -c -r -d -s -i -verbose"
-  echo "  network.sh up createChannel -ca -c -r -d -s -i -verbose"
-  echo "  network.sh createChannel -c -r -d -verbose"
-  echo "  network.sh deployCC -l -v -r -d -verbose"
+  echo " Possible Mode and flag combinations"
+  echo "   "$'\e[0;32m'up$'\e[0m' -ca -c -r -d -s -i -verbose
+  echo "   "$'\e[0;32m'up createChannel$'\e[0m' -ca -c -r -d -s -i -verbose
+  echo "   "$'\e[0;32m'createChannel$'\e[0m' -c -r -d -verbose
+  echo "   "$'\e[0;32m'deployCC$'\e[0m' -ccn -ccl -ccv -ccs -ccp -cci -r -d -verbose
   echo
   echo " Taking all defaults:"
-  echo "	network.sh up"
+  echo "   network.sh up"
   echo
   echo " Examples:"
-  echo "  network.sh up createChannel -ca -c mychannel -s couchdb -i 2.0.0"
-  echo "  network.sh createChannel -c channelName"
-  echo "  network.sh deployCC -l javascript"
+  echo "   network.sh up createChannel -ca -c mychannel -s couchdb -i 2.0.0"
+  echo "   network.sh createChannel -c channelName"
+  echo "   network.sh deployCC -ccn basic -ccl javascript"
+  echo "   network.sh deployCC -ccn mychaincode -ccp ./user/mychaincode -ccv 1 -ccl javascript"
 }
 
 # Obtain CONTAINER_IDS and remove them
@@ -81,7 +86,7 @@ function removeUnwantedImages() {
 }
 
 # Versions of fabric known not to work with the test network
-BLACKLISTED_VERSIONS="^1\.0\. ^1\.1\. ^1\.2\. ^1\.3\. ^1\.4\."
+NONWORKING_VERSIONS="^1\.0\. ^1\.1\. ^1\.2\. ^1\.3\. ^1\.4\."
 
 # Do some basic sanity checking to make sure that the appropriate versions of fabric
 # binaries/images are available. In the future, additional checking for the presence
@@ -112,7 +117,7 @@ function checkPrereqs() {
     echo "==============================================="
   fi
 
-  for UNSUPPORTED_VERSION in $BLACKLISTED_VERSIONS; do
+  for UNSUPPORTED_VERSION in $NONWORKING_VERSIONS; do
     echo "$LOCAL_VERSION" | grep -q $UNSUPPORTED_VERSION
     if [ $? -eq 0 ]; then
       echo "ERROR! Local Fabric binary version of $LOCAL_VERSION does not match the versions supported by the test network."
@@ -205,7 +210,7 @@ function createOrgs() {
     res=$?
     set +x
     if [ $res -ne 0 ]; then
-      echo "Failed to generate certificates..."
+      echo $'\e[1;32m'"Failed to generate certificates..."$'\e[0m'
       exit 1
     fi
 
@@ -218,7 +223,7 @@ function createOrgs() {
     res=$?
     set +x
     if [ $res -ne 0 ]; then
-      echo "Failed to generate certificates..."
+      echo $'\e[1;32m'"Failed to generate certificates..."$'\e[0m'
       exit 1
     fi
 
@@ -231,7 +236,7 @@ function createOrgs() {
     res=$?
     set +x
     if [ $res -ne 0 ]; then
-      echo "Failed to generate certificates..."
+      echo $'\e[1;32m'"Failed to generate certificates..."$'\e[0m'
       exit 1
     fi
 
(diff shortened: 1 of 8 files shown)


test-network/scripts/deployCC.sh
```diff
@@ -1,47 +1,88 @@
...
+elif [ "$CC_SRC_LANGUAGE" = "javascript" ]; then
+	CC_RUNTIME_LANGUAGE=node
+
 elif [ "$CC_SRC_LANGUAGE" = "typescript" ]; then
-	CC_RUNTIME_LANGUAGE=node # chaincode runtime language is node.js
-	CC_SRC_PATH="../chaincode/fabcar/typescript/"
+	CC_RUNTIME_LANGUAGE=node
 
 	echo Compiling TypeScript code into JavaScript ...
 	pushd ../chaincode/fabcar/typescript
```
nikhil550, line 88: Path is still hard coded to fabcar

test-network/scripts/deployCC.sh
```diff
@@ -1,47 +1,88 @@
...
-elif [ "$CC_SRC_LANGUAGE" = "javascript" ]; then
-	CC_RUNTIME_LANGUAGE=node # chaincode runtime language is node.js
-	CC_SRC_PATH="../chaincode/fabcar/javascript/"
-
 elif [ "$CC_SRC_LANGUAGE" = "java" ]; then
 	CC_RUNTIME_LANGUAGE=java
-	CC_SRC_PATH="../chaincode/fabcar/java/build/install/fabcar"
 
 	echo Compiling Java code ...
 	pushd ../chaincode/fabcar/java
```
nikhil550, line 76: Path still hard coded to fabcar

test-network/scripts/deployCC.sh
```diff
@@ -1,47 +1,88 @@
...
+		CC_SRC_PATH="../asset-transfer-ledger-queries"
+	elif [ "$CC_NAME" = "private" ]; then
+		echo $'\e[0;32m'asset-transfer-private-data$'\e[0m' application
+		CC_SRC_PATH="../asset-transfer-private-data"
+	else
+		echo The chaincode name ${CC_NAME} is not supported by this script
+		echo Supported chaincode names are: basic, secure, and private
+		exit 1
+	fi
+
```
nikhil550, line 49: We will probably need to catch errors when a particular chaincode language is not available

test-network/network.sh
```diff
@@ -20,40 +20,44 @@ export VERBOSE=false
...
+  echo "      "$'\e[0;32m'deployCC$'\e[0m' - deploy the asset transfer basic chaincode on the channel or specify
+  echo "      "$'\e[0;32m'down$'\e[0m' - clear the network with docker-compose down
+  echo "      "$'\e[0;32m'restart$'\e[0m' - restart the network
   echo
   echo "    Flags:"
   echo "    -ca <use CAs> -  create Certificate Authorities to generate the crypto material"
   echo "    -c <channel name> - channel name to use (defaults to \"mychannel\")"
   echo "    -s <dbtype> - the database backend to use: goleveldb (default) or couchdb"
   echo "    -r <max retry> - CLI times out after certain number of attempts (defaults to 5)"
   echo "    -d <delay> - delay duration in seconds (defaults to 3)"
```
nikhil550, line 36: This is a matter of preference, but this list is getting quite long. My suggestion would be to remove the chaincode parameters and create a separate help text that would come up when you issue `./network.sh deployCC -h` That way the parameters for deployment of the network and deployment of a chaincode are not together. That would also make it easier to keep the flag names at one letter (which I would prefer but is not important one war or the other).

test-network/network.sh
```diff
@@ -20,40 +20,44 @@ export VERBOSE=false
...
   echo "Usage: "
   echo "  network.sh <Mode> [Flags]"
-  echo "    <Mode>"
-  echo "      - 'up' - bring up fabric orderer and peer nodes. No channel is created"
-  echo "      - 'up createChannel' - bring up fabric network with one channel"
-  echo "      - 'createChannel' - create and join a channel after the network is created"
-  echo "      - 'deployCC' - deploy the fabcar chaincode on the channel"
-  echo "      - 'down' - clear the network with docker-compose down"
-  echo "      - 'restart' - restart the network"
+  echo "    Modes:"
```
nikhil550, line 23: I like the colors.

test-network/network.sh
```diff
@@ -20,40 +20,44 @@ export VERBOSE=false
...
   echo "    Flags:"
   echo "    -ca <use CAs> -  create Certificate Authorities to generate the crypto material"
   echo "    -c <channel name> - channel name to use (defaults to \"mychannel\")"
   echo "    -s <dbtype> - the database backend to use: goleveldb (default) or couchdb"
   echo "    -r <max retry> - CLI times out after certain number of attempts (defaults to 5)"
   echo "    -d <delay> - delay duration in seconds (defaults to 3)"
-  echo "    -l <language> - the programming language of the chaincode to deploy: go (default), java, javascript, typescript"
-  echo "    -v <version>  - chaincode version. Must be a round number, 1, 2, 3, etc"
+  echo "    -ccl <language> - the programming language of the chaincode to deploy: go (default), java, javascript, typescript"
+  echo "    -ccn <name> - the short name of the chaincode to deploy: basic (default), secured, private"
```
nikhil550, line 38: -ccn basically serves two purposes. One is that to select from one of the asset transfer samples. It also provides the name of the smart contract if you provide the cc path. We should find a way to make that more clear.

test-network/scripts/deployCC.sh
```diff
@@ -52,204 +93,214 @@ elif [ "$CC_SRC_LANGUAGE" = "typescript" ]; then
...
+	peer chaincode invoke -o localhost:7050 --ordererTLSHostnameOverride orderer.example.com --tls --cafile $ORDERER_CA -C $CHANNEL_NAME -n ${CC_NAME} $PEER_CONN_PARMS --isInit -c ${fcn_call} >&log.txt
+	res=$?
+	set +x
+	cat log.txt
+	verifyResult $res "Invoke execution on $PEERS failed "
+	echo "===================== Invoke transaction successful on $PEERS on channel '$CHANNEL_NAME' ===================== "
+	echo
 }
 
 chaincodeQuery() {
```
nikhil550, line 272: The chaincodeQuery function is still here, but not used by the script. It was used to print the list of cars for fabcar. It would be good to print the list of assets for the basic asset if the InitLedger function is provided.

test-network/scripts/deployCC.sh
```diff
@@ -284,13 +335,13 @@ commitChaincodeDefinition 1 2
...
 
-## Invoke the chaincode
-chaincodeInvokeInit 1 2
-
-sleep 10
-
-# Query chaincode on peer0.org1
-echo "Querying chaincode on peer0.org1..."
-chaincodeQuery 1
+## Invoke the chaincode - this does require that the chaincode have the 'initLedger'
```
nikhil550, line 338: This switch conflates the --init-required flag with the idea of having a ledger initialization function, when the two are distinct. However, providing one flow is a demonstration of how it works, and having both of them as flags could make things even more confusing.

test-network/scripts/deployCC.sh
nikhil550: ie, basic asset transfer for java.

test-network/scripts/deployCC.sh
```diff
@@ -1,47 +1,88 @@
...
-: ${VERBOSE:="false"}
 CC_SRC_LANGUAGE=`echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:]`
 
 FABRIC_CFG_PATH=$PWD/../config/
 
-if [ "$CC_SRC_LANGUAGE" = "go" -o "$CC_SRC_LANGUAGE" = "golang" ] ; then
-	CC_RUNTIME_LANGUAGE=golang
-	CC_SRC_PATH="../chaincode/fabcar/go/"
+# User has not provided a path, therefore the cc name must
+# be the short name of known applications
```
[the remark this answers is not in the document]
harrisob: right, thanks

test-network/scripts/deployCC.sh
harrisob: fixed

test-network/scripts/deployCC.sh
harrisob: fixed

test-network/network.sh
harrisob: good idea, for now I will leave it all in one place

test-network/network.sh
harrisob: I agree

test-network/scripts/deployCC.sh
harrisob: I will remove the function. To use the script would need to work for all chaincodes which would require adding parameters for the function to call and arguments for the function. All of which the user should be doing with the execution of the sample applications.

test-network/scripts/deployCC.sh
harrisob: Yes that was the idea to combine the two , if you use one you have to do the other and keep it simple to use.  The user is free to run their own init function that is not tied to the lifecycle initialization tracking process by doing an invoke with the application.

test-network/scripts/deployCC.sh
```diff
@@ -1,47 +1,88 @@
...
+if [ "$CC_SRC_PATH" = "NA" ]; then
+	echo Determining the path to the application
```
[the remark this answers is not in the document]
harrisob: fixed

test-network/network.sh
```diff
@@ -20,40 +20,45 @@ export VERBOSE=false
...
   echo "    -ca <use CAs> -  create Certificate Authorities to generate the crypto material"
   echo "    -c <channel name> - channel name to use (defaults to \"mychannel\")"
   echo "    -s <dbtype> - the database backend to use: goleveldb (default) or couchdb"
   echo "    -r <max retry> - CLI times out after certain number of attempts (defaults to 5)"
   echo "    -d <delay> - delay duration in seconds (defaults to 3)"
-  echo "    -l <language> - the programming language of the chaincode to deploy: go (default), java, javascript, typescript"
-  echo "    -v <version>  - chaincode version. Must be a round number, 1, 2, 3, etc"
+  echo "    -ccn <name> - the short name of the chaincode to deploy: basic (default), secured, ledger, private"
+  echo "    -ccl <language> - the programming language of the chaincode to deploy: go (default), java, javascript, typescript"
+  echo "    -ccn <name> - the short name of the chaincode to deploy: basic (default), secured, private"
```
denyeart, line 39: -ccn is added twice here.

Also, let's keep the sample names in the same logical order as the directory structure - basic (default), ledger, private, secured

test-network/scripts/deployCC.sh
```diff
@@ -1,255 +1,307 @@
...
+	pushd $CC_SRC_PATH
 	npm install
 	npm run build
 	popd
 	echo Finished compiling TypeScript code into JavaScript
 
 else
 	echo The chaincode language ${CC_SRC_LANGUAGE} is not supported by this script
-	echo Supported chaincode languages are: go, java, javascript, and typescript
+	echo Supported chaincode languages are: go, golang, java, javascript, and typescript
```
denyeart, line 97: Why add 'golang'? why not just keep as 'go' everywhere?

test-network/scripts/deployCC.sh
```diff
@@ -1,255 +1,307 @@
...
 	exit 1
 fi
 
+INIT_REQUIRED="--init-required"
```
denyeart, line 101: --init-required is not the same thing as init function now used in the test network, let's not confuse the two.

test-network/scripts/deployCC.sh
```diff
@@ -1,255 +1,307 @@
...
+	res=$?
+	verifyResult $res "Invoke transaction failed on channel '$CHANNEL_NAME' due to uneven number of peer and org parameters "
+
+	# while 'peer chaincode' command can get the orderer endpoint from the
+	# peer (if join was successful), let's supply it directly as we know
+	# it using the "-o" option
+	set -x
+	fcn_call='{"function":"'${CC_INIT_FCN}'","Args":[]}'
+	echo invoke fcn call:${fcn_call}
+	peer chaincode invoke -o localhost:7050 --ordererTLSHostnameOverride orderer.example.com --tls --cafile $ORDERER_CA -C $CHANNEL_NAME -n ${CC_NAME} $PEER_CONN_PARMS --isInit -c ${fcn_call} >&log.txt
```
denyeart, line 264: We're not using init-required and --isInit now, right?

test-network/network.sh
denyeart: I'd suggest to print all the help at once, but it would be good to separate the functions and show the available flags per function, to make it clear that the -ccX flags are only available with 'deployCC'.

denyeart:
Overall these are great changes, thanks Bret.

test-network/network.sh
```diff
@@ -20,40 +20,45 @@ export VERBOSE=false
...
+  echo "    -ccv <version>  - chaincode version. Must be an integer, 1 (default), 2, 3, etc"
+  echo "    -ccp <path>  - Optional, chaincode path. Path to the chaincode. When provided the -ccn will be used as the deployed name and not the short name of the known chaincodes."
```
nikhil550, line 41: To support the asset transfer series, these flags would also need to support providing the endorsement policy of the chaincode, with a default of majority if one is not provided. This would support the "OR" endorsement policy used by the private data smart contract and the secured agreement smart contract.

For the private data smart contract, you also need to provide the path to the collection configuration file.

denyeart: Agreed. I think these changes could come in a subsequent PR though, I prefer to get the basic support merged for people to try out.

test-network/network.sh
harrisob: done

test-network/scripts/deployCC.sh
harrisob: removed

test-network/scripts/deployCC.sh
harrisob: if one defines a definition as init required then they would have to run the init function before it is usable, so I thought that we could combine the idea into just having the user define an init function and then set the init required and run the init for them.

Are you thinking that we should add another parameter for the init required ?

test-network/scripts/deployCC.sh
harrisob: we are using both if the user defines an init function

PR #237, opened 2020-07-07
Refactor Asset Secured Chaincode Into Idiomatic Go

btl5037:
Rewrites the chaincode in idiomatic Go and cleans up
the general implementation.

A future commit should push the chaincode logic itself
into a separate package as chaincode cannot be tested
when the logic is part of the main package.

### DIFF
--- a/asset-transfer-secured-agreement/chaincode-go/asset_transfer.go
+++ b/asset-transfer-secured-agreement/chaincode-go/asset_transfer.go
@@ -1,21 +1,6 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
- * regarding copyright ownership.  The ASF licenses this file
- * to you under the Apache License, Version 2.0 (the
- * "License"); you may not use this file except in compliance
- * with the License.  You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing,
- * software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


asset-transfer-secured-agreement/chaincode-go/asset_transfer.go
```diff
@@ -273,211 +252,228 @@ func (s *SmartContract) TransferAsset(ctx contractapi.TransactionContextInterfac
...
+		)
 	}
 
-	// since all checks passed, return without an error
 	return nil
 }
 
-// transferAssetState makes the public and private state updates for the transferred asset
+// transferAssetState performs the public and private state updates for the transferred asset
 func transferAssetState(ctx contractapi.TransactionContextInterface, asset *Asset, immutablePropertiesJSON []byte, clientOrgID string, buyerOrgID string, price int) error {
```
btl5037, line 372: This function seems very dangerous. It has many error paths, and after the initial PutState, any one of those error paths could leave the asset in a state where multiple orgs own it, multiple orgs can modify it, among other things. There is no logic to handle a recovery in the even an error occurs.

denyeart: > This function is very dangerous. It has many error paths, and after the initial PutState, any one of those error paths could leave the asset in a state where multiple orgs own it, multiple orgs can modify it, among other things. There is no logic to handle a recovery in the event an error occurs.

@btl5037 If it errors at any point, the error is returned to the client rather than a partial rwset. So the writes will be atomic, either they all succeed, or there is an error returned.  Maybe I'm not understanding the concern, can you elaborate?

PR #242, opened 2020-07-12
Fix param sequence in smartcontract and smartcontract_test.

denali49:
Fix param sequence, smart contract and smartcontract_test.


denali49:
Thanks, would be good to get PR #228 and #229 merged as well.  Both have been approved.

PR #243, opened 2020-07-15
Fix deployment of chaincode with version higher than 9 issue

Titasp:
The 'deployCC.sh' script fails if chaincode version is higher than 9 as 'grep' does not match version or sequence that is higher than 9.


test-network/scripts/deployCC.sh
```diff
@@ -185,7 +185,7 @@ queryCommitted() {
     peer lifecycle chaincode querycommitted --channelID $CHANNEL_NAME --name fabcar >&log.txt
     res=$?
     set +x
-		test $res -eq 0 && VALUE=$(cat log.txt | grep -o '^Version: [0-9], Sequence: [0-9], Endorsement Plugin: escc, Validation Plugin: vscc')
+		test $res -eq 0 && VALUE=$(cat log.txt | grep -o '^Version: [0-9]\{1,\}, Sequence: [0-9]\{1,\}, Endorsement Plugin: escc, Validation Plugin: vscc')
```
harrisob, line 188: checkout how the version is not a string, so this will not be needed, however the sequence is an integer, so maybe we could just use [0-9]+ to catch ints of all sizes ?

lehors: In resolving a conflict with master I made the change @harrisob suggested and successfully tested it deploying a chaincode all the way to v10.0 s10, and even v11.0 s11 for good measure. :-)

PR #246, opened 2020-07-15
Update CI For Asset Transfer

btl5037:
Pushes the scripting logic into a standalone script and adds tests for the applications which by side-effect also test the invoke and query functions on the chaincode.


ci/scripts/run-test-network.sh
```diff
@@ -0,0 +1,37 @@
...
+node enrollAdmin.js && node registerUser.js
+popd
+
+# Setup typescript application
+pushd ../asset-transfer-basic/application-typescript
+npm install -g typescript
+npm install
+cd src
+tsc enrollAdmin.ts invoke.ts query.ts registerUser.ts
+node enrollAdmin.js && node registerUser.js
```
harrisob, line 23: Not going to run this as the user has already been registered above.
Options are:
- do not run it and wait till I make enroll and register common
- stop and start the network before each application
- have each application create their own user with unique name

btl5037: I moved to just copying the wallet for now. Since enroll and register are SDK functionality no benefit to testing them twice

btl5037:
Converting back to draft for now. Because all of these invokes happen on the same asset `asset13` we can't invoke twice, and I don't want to keep restarting the network in between. My short answer was create a file for calling `deleteAsset` but I don't I imagine we wouldn't want to include a `delete.js` file alongside `invoke.js` and `query.js` let me think through this.

I may just end up splitting each test out into separate job executions

btl5037:
Here we go, this is a much more proper CI pipeline

ci/scripts/pullFabricImages.sh
```diff
@@ -1,15 +1,15 @@
 #!/bin/bash -e
 set -euo pipefail
 
-# FABRIC_VERSION is set in ci/azure-pipelines.yml
+FABRIC_VERSION=${FABRIC_VERSION:-2.2}
```
denyeart, line 4: Doesn't have to be addressed in this PR, but just curious if there is a way to test against the latest stable Fabric instead of hardcoding the version in several files, as we'll undoubtedly forget to update all the version numbers.

btl5037: There is. I'm only leaving it at 2.2 until we tag the 2.2.0 samples and then master can migrate to 2.3

ci/scripts/run-test-network-basic.sh
```diff
@@ -0,0 +1,60 @@
...
+npm install
+cd src
+print "Compiling Typescript"
+tsc enrollAdmin.ts registerUser.ts invoke.ts query.ts
+print "Executing enrollAdmin.js"
+node enrollAdmin.js
+print "Executing registerUser.js"
+node registerUser.js
+print "Executing invoke.js"
+node invoke.js
```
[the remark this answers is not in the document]
btl5037: I'm not sure if anyone is working on the typescript full fledged app. That wasn't something I intended to do. The nice thing about this PR is, if and when someone does create it, that CI will fail showing they aren't testing their new app, so we will catch it, and this file will get updated before a typescript app gets in

ci/scripts/run-test-network-basic.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Run Javascript application
+print "Initializing Javascript application"
+pushd ../asset-transfer-basic/application-javascript
+rm -rf wallet
+npm install
+print "Executing enrollAdmin.js"
+node enrollAdmin.js
+print "Executing registerUser.js"
+node registerUser.js
```
sijocherian, line 28: no need for
```
node enrollAdmin.js
node registerUser.js
```

since app.js calls above to add users.

harrisob: I like this where CI just makes one call to the application under test. It will decide what it needs to do.

btl5037: I didn't realize `app.js` was importing and calling enroll and register, removed the direct calls to enroll and register

ci/scripts/pullFabricImages.sh
btl5037: Since 2.2 is LTS I wanted to explicitly continue testing against 2.2 until we tag

btl5037:
Converting back to draft one more time, pending another discussion

btl5037:
Now that Typescript chaincode and apps are remove, going to simplify this greatly as well as introduce Javascript and Go Linters.

asset-transfer-basic/application-javascript/app.js
```diff
@@ -41,10 +41,10 @@ async function main() {
         // Steps:
```
sijocherian, line 41: Do you mind enhancing the func at line 18, as below, for cases when input is empty.

```
function prettyJSONString(inputString) {
    if(inputString) {
        return JSON.stringify(JSON.parse(inputString),null,2);
    }
    else {
        return inputString;
    }
}
```

ci/scripts/run-test-network-basic.sh
```diff
@@ -0,0 +1,56 @@
...
+	GREEN='\033[0;32m'
+  NC='\033[0m'
+  echo
+	echo -e "${GREEN}${1}${NC}"
+}
+
+print "Creating network"
+  ./network.sh up createChannel -ca -s couchdb -i "${FABRIC_VERSION}"
+print "Deploying ${CHAINCODE_NAME} chaincode"
+./network.sh deployCC -ccn "${CHAINCODE_NAME}" -ccv 1 -ccs 1 -ccl "${CHAINCODE_LANGUAGE}" -cci initLedger
```
sijocherian, line 17: to Davids point in line 26: the app.js calls initLedger. If we keep that, no need for `-cci initLedger`

asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
```diff
@@ -56,100 +56,93 @@ class AssetTransfer extends Contract {
...
-    // createAsset issues a new asset to the world state with given details.
-    async createAsset(ctx, id, color, size, owner, appraisedValue) {
+    // CreateAsset issues a new asset to the world state with given details.
+    async CreateAsset(ctx, id, color, size, owner, appraisedValue) {
         const asset = {
             ID: id,
             Color: color,
             Size: size,
             Owner: owner,
             AppraisedValue: appraisedValue,
```
sijocherian, line 73: can we make asset JSON compatible with that of go-chaincode ( where we have camelcase json keys like `appraisedValue`), so the application-* can work with any of these chaincodes.

specially because network.sh installs the chaincode-* using the same short name `basic` for all the langs.

asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
```diff
@@ -56,100 +56,93 @@ class AssetTransfer extends Contract {
...
+            const strValue = Buffer.from(result.value.value.toString()).toString('utf8');
             let record;
             try {
                 record = JSON.parse(strValue);
             } catch (err) {
                 console.log(err);
                 record = strValue;
             }
-            allResults.push({ Key: key, Record: record });
+            allResults.push({ Key: result.value.key, Record: record });
```
sijocherian, line 143: btw in chaincode-go , getALL returns []asset,  no Key & Record

asset-transfer-basic/chaincode-typescript/src/asset.ts
```diff
@@ -1,12 +1,26 @@
...
- */
+  SPDX-License-Identifier: Apache-2.0
+*/
 
+import {Object, Property} from 'fabric-contract-api';
+
+@Object()
 export class Asset {
+    @Property()
     public docType?: string;
```
sijocherian, line 10: if we want to keep compatiblity with that of go-chaincode :  remove docType

sijocherian:
what is typescript-application being removed, but typescript CC being kept? 

btw
#268 had similar updates

btl5037:
> what is typescript-application being removed, but typescript CC being kept?
> 
> btw
> #268 had similar updates

After speaking with the chaincode maintainers, there are typescript specific constructs, specifically the annotations that need to be demoed via chaincode. So they elected to keep typescript chaincode to demo these typescript specific constructs. None of them apply to applications however. So we will keep one example of typescript chaincode (basic).

asset-transfer-basic/chaincode-typescript/src/asset.ts
btl5037: This is an optional property, Typescript is more strongly typed than Node, so it is present, but not necessary.

asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
btl5037: Will update

asset-transfer-basic/chaincode-javascript/lib/assetTransfer.js
btl5037: Will update in a follow on PR

btl5037: Though this doesn't break anything, the application works just fine with all of the chaincodes

asset-transfer-basic/application-javascript/package.json
```diff
@@ -15,8 +15,8 @@
     "author": "Hyperledger",
     "license": "Apache-2.0",
     "dependencies": {
-        "fabric-ca-client": "^2.1.0",
-        "fabric-network": "^2.1.0"
+        "fabric-ca-client": "2.2.0",
```
denyeart, line 18: We usually keep the ^ prefix so that it will use the latest v2.x.x version, rather than v2.2.0 specifically.

PR #247, opened 2020-07-16
Add endorsement policy to chaincode deployment script

nikhil550:
Add endorsement policy and collection configuration parameters to the deployCC script.

Also add let users know if they provide a bad chaincode path, or if the language and sample are not yet available


harrisob:
Looks good, thanks

PR #257, opened 2020-07-22
Update secured asset sample tutorial link and remove README

stephyee:
Signed-off-by: Tiffany Harris <tiffany.harris@ibm.com>


asset-transfer-secured-agreement/chaincode-go/README.md
```diff
@@ -1,342 +0,0 @@
-# Secured private asset transfer scenario
```
nikhil550, line 1: I would keep this REAMDE, but just include a link to the tutorial. That would provide some instructions just in case users do not land on the front REAMDE.

PR #258, opened 2020-07-22
Remove Typescript References

btl5037:
As the Typescript examples are essentially just reiterations of the same Javascript code, there is no benefit to having providing examples in both languages. The functional code was exactly the same in both languages. On the contrary it meant we widened the surface of maintenance and thus we are removing due to the limited benefit provided by the example chaincode and applications.


lehors:
I found the following remaining references you might want to also delete:

`
chaincode/abstore/javascript/.gitignore:# TypeScript v1 declaration files
chaincode/fabcar/javascript/.gitignore:# TypeScript v1 declaration files
chaincode/marbles02/javascript/.gitignore:# TypeScript v1 declaration files
commercial-paper/organization/digibank/contract/.npmignore:# TypeScript v1 declaration files
commercial-paper/organization/magnetocorp/contract/.npmignore:# TypeScript v1 declaration files
fabcar/javascript/.gitignore:# TypeScript v1 declaration files
`

btl5037:
Thanks Arnaud, when I did the search I didnt hit it with case sensitivity, let me grab those as well quick

btl5037:
> I found the following remaining references you might want to also delete:
> 
> ```
> chaincode/abstore/javascript/.gitignore:# TypeScript v1 declaration files
> chaincode/fabcar/javascript/.gitignore:# TypeScript v1 declaration files
> chaincode/marbles02/javascript/.gitignore:# TypeScript v1 declaration files
> commercial-paper/organization/digibank/contract/.npmignore:# TypeScript v1 declaration files
> commercial-paper/organization/magnetocorp/contract/.npmignore:# TypeScript v1 declaration files
> fabcar/javascript/.gitignore:# TypeScript v1 declaration files
> ```

Removed these references as well

btl5037:
I did discuss this removal with Simon and Dave as well before doing this PR. Just wanted to clarify this wasn't out of the blue.

mbwhite:
Sorry I was away, and not able to comment on this; however, with the removal of the TypeScript examples, we are not showing how to use the typescript annotations. These provide much valuable metadata that can be used by client applications and tooling. 

I would rather remove the JS and keep the TS.  TS offers more to the contract developer over and above just being 'typed'.

Without knowledge of the annotations, developers are missing out.

lehors:
Ok, sorry we missed that. We can revert this change.

denali49:
+1

lehors:
I'm not sure about removing javascript... There are far more people that use javascript than that use typescript.

btl5037:
So my middle ground here would be, we leave the typescript and javascript chaincode, but don't do a typescript application. The annotation model would exist for the chaincode and thats the only thing we demo when it comes to typescript

lehors:
That sounds good to me. We should also limit this to one of the chaincode samples. We don't need to get a typescript version for every sample chaincode.

btl5037:
> That sounds good to me. We should also limit this to one of the chaincode samples. We don't need to get a typescript version for every sample chaincode.

Agreed, my biggest concern with providing so many samples and in so many languages, is that over time the community will contribute updates and fixes to the applications and chaincodes. It's unlikely a user will make the change in the node, java, go and typescript chaincode, and we will need to be on the look out for changes that are made to one and ask the user to add the same changes to the other to keep them all consistent. Having more languages just increases the surface area for maintenance making the job of the maintainers more difficult to keep everything up to date.

lehors:
Indeed.

PR #259, opened 2020-07-23
Update deployCC.sh

tyanj:
res=$? missing


btl5037:
Thank you for the update @tyanj.

We do require all commits to be signed-off by the author, if you could checkout your code and do a `git commit -s --amend --no-edit && git push origin master -f` where the `-s` flag will add your Git signature.

Thank you again for the contribution.

tyanj:
already re-pr

PR #266, opened 2020-07-24
Update PNG images to remove reference to 'marble'.

denali49:


nikhil550:
This images should actually no longer be in the samples, since the README was removed. I would just remove the images folder, and move the changes to the tutorial in the Fabric docs

denali49:
> This images should actually no longer be in the samples, since the README was removed. I would just remove the images folder, and move the changes to the tutorial in the Fabric docs

Will do.

PR #276, opened 2020-08-03
asset-transfer-basic java application

r2roC:
Tested using the test network along with go chaincode. 
Instructions for running are at the top of TestApp.java.

Signed-off-by: r2roC <arturo@IBM.com>


asset-transfer-basic/application-java/src/main/java/application/java/ClientUtilities.java
```diff
@@ -0,0 +1,177 @@
...
+			Network network = gateway.getNetwork("mychannel");
+			Contract contract = network.getContract("basic");
+
+			contract.submitTransaction("CreateAsset", assetID, color, size, owner, appraisedValue);
+			System.out.println("Created asset: [assetID=" + assetID + ", color="+ color + ", size=" + size 
+								+ ", owner=" + owner + ", appraisedValue=" + appraisedValue +"]");
+			gateway.close();
+		}
+		catch(Exception e){
+			System.out.println("Error connecting to gateway: " + e);
```
sijocherian, line 74: System.err.println

asset-transfer-basic/application-java/src/main/java/application/java/ClientUtilities.java
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+
+	// connects to the chaincode and calls CreateAsset()
+	public static void CreateAsset(final String assetID, final String color, final String size,
+        final String owner, final String appraisedValue) throws Exception {
+
+		// create a gateway connection
+		try (Gateway gateway = connect()) {
```
sijocherian, line 62: Call connect only once in a sample transaction sequence. see my comment in the unit test file

calling connect() is expensive, since it reads wallet & config files possibly.
Avoid calling it  in each txn util  like CreateAsset , ReadAsset

asset-transfer-basic/application-java/src/test/java/application/java/TestApp.java
```diff
@@ -0,0 +1,64 @@
...
+	// Creates asset7, modify the parameters to create other assets
+	@Test
+	public void CreateAssetTest() throws Exception {
+		ClientUtilities.CreateAsset("asset7", "magenta", "20", "kenysha", "800");
+	}
+
+	// Reads asset4, modify the parameters to read other assets
+	@Test
+	public void ReadAssetTest() throws Exception {
+		ClientUtilities.ReadAsset("asset4");
```
sijocherian, line 39: If we go with unit test to demo the app, then we should add assert statements

asset-transfer-basic/application-java/src/test/java/application/java/TestApp.java
```diff
@@ -0,0 +1,64 @@
...
+// gradle test --tests TestApp.TransferAssetTest
+// gradle test --tests TestApp.GetAllAssetsTest
+
+package application.java;
+
+import org.junit.Test;
+import static org.junit.Assert.*;
+
+
+public class TestApp {
```
sijocherian, line 22: I think a main method is better to demo an app, than a unit test, since each test is disconnected.
Not sure if we decided on one way or other, during the brief time in the call this morning.

My suggestion is to have a main() method (not in the src/test), that calls a sequence of invoke & query with some Println like in js app.

and, Use  connect() only once in a sample transaction sequence.

asset-transfer-basic/application-java/src/main/java/application/java/ClientUtilities.java
r2roC: Sounds good, I updated the exceptions to be outputted to err

asset-transfer-basic/application-java/src/main/java/application/java/ClientUtilities.java
r2roC: Good point, I restructured the app so that connect() is only called once.

asset-transfer-basic/application-java/src/test/java/application/java/TestApp.java
r2roC: I switched directions and removed the unit tests, a main file will be used instead to showcase the app.

asset-transfer-basic/application-java/src/test/java/application/java/TestApp.java
r2roC: I restructured it in this exact way, a main file will be used instead to showcase the app.

asset-transfer-basic/application-java/src/main/java/application/java/EnrollAdmin.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hyperledger.fabric.sdk.security.CryptoSuite;
+import org.hyperledger.fabric.sdk.security.CryptoSuiteFactory;
+import org.hyperledger.fabric_ca.sdk.EnrollmentRequest;
+import org.hyperledger.fabric_ca.sdk.HFCAClient;
+
+public class EnrollAdmin {
+
+	static {
+		System.setProperty("org.hyperledger.fabric.sdk.service_discovery.as_localhost", "true");
+	}
```
harrisob, line 26: Please remove

asset-transfer-basic/application-java/src/main/java/application/java/RegisterUser.java
```diff
@@ -0,0 +1,111 @@
...
+import org.hyperledger.fabric.sdk.security.CryptoSuite;
+import org.hyperledger.fabric.sdk.security.CryptoSuiteFactory;
+import org.hyperledger.fabric_ca.sdk.HFCAClient;
+import org.hyperledger.fabric_ca.sdk.RegistrationRequest;
+
+public class RegisterUser {
+
+	static {
+		System.setProperty("org.hyperledger.fabric.sdk.service_discovery.as_localhost", "true");
+	}
```
harrisob, line 28: Please remove

asset-transfer-basic/application-java/src/main/java/application/java/RegisterUser.java
```diff
@@ -0,0 +1,111 @@
...
+
+		};
+
+		// Register the user, enroll the user, and import the new identity into the wallet.
+		RegistrationRequest registrationRequest = new RegistrationRequest("appUser");
+		registrationRequest.setAffiliation("org1.department1");
+		registrationRequest.setEnrollmentID("appUser");
+		String enrollmentSecret = caClient.register(registrationRequest, admin);
+		Enrollment enrollment = caClient.enroll("appUser", enrollmentSecret);
+		Identity user = Identities.newX509Identity("Org1MSP", adminIdentity.getCertificate(), adminIdentity.getPrivateKey());
```
harrisob, line 106: Why does the user have the admin identity certificate and private key ?

r2roC: I'm not sure, I didn't touch enrollAdmin and registerUser when I grabbed it from fabcar. registerUser.js in the js app, is formatted the same way. I'm open to alternatives.

PR #297, opened 2020-08-13
Upgrade the chaincode docker devmode sample to fabric 2.x capability

jchen8tw:
Issue:
- The new Fabric chaincode lifecycle won't allow instantiating  chaincode.

- The original channel artifacts didn't support contract with private data.

Therefore, I've upgraded the channel artifacts to v2_0 capability and also modify the ReadMe with steps to start a
chaincode devmode in fabric 2.2.
There is also a `installChaincode.sh` for users to easily package, install chaincode, approve chaincode definition, commit chaincode definition in one script.


lehors:
I have to ask: what does devmode bring that we don't now have with the external chaincode. It is fairly straightforward to debug your chaincode when running as an external chaincode - I've done it!
Admittedly this may be more of a question for fabric than fabric-samples but I think it could be argued that we could reduce our technical debt by removing all the devmode code at this point.

jchen8tw:
> I have to ask: what does devmode bring that we don't now have with the external chaincode? It is fairly straightforward to debug your chaincode when running as an external chaincode - I've done it!
> Admittedly this may be more of a question for fabric than fabric-samples but I think it could be argued that we could reduce our technical debt by removing all the devmode code at this point.

You're right, I think I'll just close this pull request.
Maybe we should come up with another example of debugging external chaincode in the future.

PR #300, opened 2020-08-14
Update asset-transfer-basic external chaincode readme

denyeart:
Minor clarifications to external chaincode readme.


lehors:
> Do we want to say something about the location of the volumes being added to the docker compose need to be listed within the 'file sharing' of the docker resources under preferences.

I don't think so. For most people this should work out of the box with the default docker settings. And if not, we're probably talking about someone who knows what they are doing with docker.

PR #303, opened 2020-08-17
fix ccp template and generator to include orderer and channel

nasermirzaei89:
Update ccp template and generator to include orderer and channel.
needed for `asset-transfer-basic/application-go/assetTransfer.go`

Signed-off-by: Naser Mirzaei <nasermirzaei89@gmail.com>

### DIFF
--- a/test-network/organizations/ccp-generate.sh
+++ b/test-network/organizations/ccp-generate.sh
@@ -5,26 +5,35 @@ function one_line_pem {
 }
 
 function json_ccp {
-    local PP=$(one_line_pem $4)
-    local CP=$(one_line_pem $5)
+    local OP=$(one_line_pem $5)
+    local PP=$(one_line_pem $6)
+    local CP=$(one_line_pem $7)
     sed -e "s/\${ORG}/$1/" \
-        -e "s/\${P0PORT}/$2/" \
-        -e "s/\${CAPORT}/$3/" \
+        -e "s/\${ORDERERPORT}/$2/" \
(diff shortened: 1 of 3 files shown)


nikhil550:
The simplified connection profile is best practice when using service discovery, which we encourage. May be better to create another ccp specifically for the go application, rather than update the connection profile used by all of them.

nasermirzaei89:
> This should not be required when using the gateway GO.
> 
> Is there a failure that needs to addressed in this sample or the gateway GO package ?

when I try to run `asset-transfer-basic/application-go/assetTransfer.go` it tells:
`SendTransaction failed: orderers is nil`

More info at https://stackoverflow.com/questions/62927393/hyperledger-2-2-fabcar-transaction-issue

andrew-coleman (changes requested):
This should not be needed for any of the 'gateway' SDKs.
Do you have the DISCOVERY_AS_LOCALHOST env-var set to "TRUE"?

nikhil550:
I think we have agreed that this pull request could be closed. To fix this bug, rather than changing the main connection profiles, we should
1) update the Go code or SDK to allow it to make use of the connection profiles used by the other SDK's
or 
2) If that is not possible, create a new connection profile that can be used by go applications.

andrew-coleman:
Just to clarify, the Go SDK gateway package _does_ support the same connection profile as the other SDKs.  The fabcar sample (now deprecated) used this same network as this and works.  The fabcar sample gets run as part of the CI pipeline.

PR #290, opened 2020-08-10
Asset transfer private-data go-chaincode - Add unit tests

sijocherian:
- refactored to chaincode package as best practice for organization
- Added unit tests for query & transfer asset chaincode functions, with counterfeiter mocks

PR #305, opened 2020-08-17
Re-enable CI tests for asset-transfer-private-data JS app

denyeart:
Re-enable CI tests for asset-transfer-private-data JavaScript application.

Also re-add the error when private asset details are not found in the collection,
so that the CLI instructions work as desired, and to demonstrate error
handing in the client application.

### DIFF
--- a/asset-transfer-private-data/chaincode-go/private_asset_queries.go
+++ b/asset-transfer-private-data/chaincode-go/private_asset_queries.go
@@ -147,7 +147,6 @@ func (s *SmartContract) QueryAssetByOwner(ctx contractapi.TransactionContextInte
 	return queryResults, nil
 }
 
-
 // QueryAssets uses a query string to perform a query for assets.
 // Query string matching state database syntax is passed in and executed as is.
 // Supports ad hoc queries that can be defined at runtime by the client.
--- a/ci/scripts/run-test-network-private.sh
+++ b/ci/scripts/run-test-network-private.sh
@@ -15,7 +15,7 @@ function createNetwork() {
   print "Creating network"
   ./network.sh up createChannel -ca -s couchdb -i "${FABRIC_VERSION}"
   print "Deploying ${CHAINCODE_NAME} chaincode"
-  ./network.sh deployCC -ccn "${CHAINCODE_NAME}" -ccv 1 -ccs 1 -ccl "${CHAINCODE_LANGUAGE}"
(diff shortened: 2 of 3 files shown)


ci/scripts/run-test-network-private.sh
```diff
@@ -25,10 +25,10 @@ function stopNetwork() {
 
 # Run Javascript application
 createNetwork
-#print "Initializing Javascript application"
-#pushd ../asset-transfer-private-data/application-javascript
-#npm install
-#print "Executing app.js"
-#node app.js
-#popd
```
sijocherian, line 33: In #290 the CI is already enabled & passing

https://github.com/hyperledger/fabric-samples/blob/fe04cdcb65fc50d22fcb31b8a045e2ae12ee9e61/ci/scripts/run-test-network-private.sh

asset-transfer-private-data/chaincode-go/private_asset_queries.go
```diff
@@ -47,8 +47,9 @@ func (s *SmartContract) ReadAssetPrivateDetails(ctx contractapi.TransactionConte
 		return nil, fmt.Errorf("failed to read from asset details %v", err)
 	}
 	if assetDetailsJSON == nil {
-		log.Printf("AssetPrivateDetails for %v does not exist in collection %v", assetID, collection)
-		return nil, nil
+		message := fmt.Sprintf("AssetPrivateDetails for %s does not exist in collection %s", assetID, collection)
+		log.Print(message)
+		return nil, fmt.Errorf(message)
```
sijocherian, line 52: In the last PR, while developing JS app, I switched to the pattern of reads to return empty data when Data_Not_Found: so, all the ReadXXX functions  return empty data , instead of throwing Error. As a client, I believe that was more intuitive.

IIRC when the read call access a collection that it does not have access to, the error is returned to client, to indicate Access error.

if we really need this pattern to throw Err, then the change should be for all ReadXX and the unit test in #290 need rework.

ci/scripts/run-test-network-private.sh
btl5037: Correct, a later update I added erroneously removed the the collection config, Dave fixed it for me

asset-transfer-private-data/chaincode-go/private_asset_queries.go
denyeart: I do prefer to return an error if the asset is not read (all the other samples do this). If nothing else, showing how to handle an expected error in the application, will help users to start thinking about how to handle all errors.

But I agree that the change is out of scope of this PR. I've removed it for now, I'll defer to Sijo since he has been doing the majority of the work for this sample.

denyeart: Sounds good. There is not an 'update' per se in this sample, but you could test failure of any of the conditions validated against in verifyAgreement() or verifyClientOrgMatchesPeerOrg().

PR #309, opened 2020-08-22
PrivateData App : Added two client operations where chaincode is expected to return error

sijocherian:


denyeart:
The 1st commit adds two try/catch blocks. The 2nd commit removes one of the try/catch blocks.
Note - it is usually a good practice to squash the commits in the PR, so that the maintainer understands the final intent of the PR.
Some good squash guidance can be found here: https://hyperledger-fabric.readthedocs.io/en/latest/github/github.html#updating-a-pull-request

sijocherian:
thanks. #290 included both the try catch blocks, so this is not needed anymore

PR #310, opened 2020-08-26
add utils.sh and colorize messages

nasermirzaei89:
In this commit, I added a new file `utils.sh` that includes some exported functions for echo colorized messages.
the functions are:
1. `println` echos normal messages
1. `infoln` echos info messages in blue color
1. `successln` echos success messages in green color
1. `warnln` echos warn messages in yellow color
1. `errorln` echos error messages in red color
1. `fatalln` echos error messages in red color and exits with fail.

All scripts (except org3 scripts) uses these functions to echo their messages.

With these colors, we don't need to make important messages more visible with `##` or `!!!!!!!!`


test-network/utils.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
nikhil550, line 1: I like this Idea. May rename this file scriptUtils so that folks are aware at first glance that there is no blockchain content in this file

nasermirzaei89: Done

PR #315, opened 2020-08-28
Readme links fix - point to latest

rijulaggarwal16:
The links in the README redirect to a 404. The valid documentation links point to `release-2.2` instead of `master`.

Signed-off-by: Rijul Aggarwal <rijul.aggarwal@ibm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,18 +2,18 @@
 
 # Hyperledger Fabric Samples
 
-You can use Fabric samples to get started working with Hyperledger Fabric, explore important Fabric features, and learn how to build applications that can interact with blockchain networks using the Fabric SDKs. To learn more about Hyperledger Fabric, visit the [Fabric documentation](https://hyperledger-fabric.readthedocs.io/en/master).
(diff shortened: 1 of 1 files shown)


lehors (changes requested):
Thanks for your contribution. I didn't know the fabric doc master branch was no longer released so that's a good catch. However, I think unless we have a reason to link to a specific version we're better off linking to the latest version of the doc. Otherwise we're going to keep having to change this every time there is a new fabric release.

harrisob:
The sample code points to a specific code version, so we must point to the matching doc. No changes will be needed when this code base is finally merged into a dedicated branch for release-2.2.

harrisob:
However since the release-2.2 is not as up to date as latest, for example the https://hyperledger-fabric.readthedocs.io/en/latest/secured_asset_transfer/secured_private_asset_transfer_tutorial.html is not in release-2.2

I think we are forced to use "latest" on the links

lehors (approved):
LGTM. Thanks.

PR #318, opened 2020-09-01
Add account-based token sample (go chaincode)

denyeart:
Add an account-based token sample go chaincode, as well as an associated readme tutorial.


token-account-based/README.md
```diff
@@ -0,0 +1,200 @@
+# Account-based token scenario
+
```
nikhil550, line 2: Given that this is a new topic, should we add a discussion on the advantages of account based vs UTXO based tokens?

token-account-based/README.md
```diff
@@ -0,0 +1,200 @@
...
+The account-based token smart contract demonstrates how to create and transfer fungible tokens using an account-based model. In an account-based model, there is an account for each participant that holds a balance of tokens.
+A mint transaction creates tokens in an account, while a transfer transaction debits the caller's account and credits another account.
+
+In this sample it is assumed that only one organization (played by Org1) is in a central banker role and can mint new tokens into their account, while any organization can transfer tokens from their account to a recipient's account.
+Accounts could be setup at the organization level or client identity level. In this sample accounts are setup at the client identity level, where every authorized client with an enrollment certificate from their organization has an account keyed by their client ID.
+The client ID is simply a base64-encoded concatenation of the issuer distinguished name and subject distinguished name from the client identity's enrollment certificate. The client ID can therefore be considered the payment address of a recipient.
```
nikhil550, line 8: It might be worth it to try to explain distinguished name and subject in more plain language.

token-account-based/README.md
```diff
@@ -0,0 +1,200 @@
...
+The -ca flag is used to deploy the network using certificate authorities. This allows you to use each organization's CA to register and enroll new users for this tutorial.
+
+## Deploy the smart contract to the channel
+
+You can use the test network script to deploy the account-based token contract to the channel that was just created. Deploy the smart contract to `mychannel` using the following command:
+```
+./network.sh deployCC -ccn token_account -ccp ../token-account-based/chaincode-go/
+```
+
+The above command deploys the go chaincode with short name `token_account`. The smart contract will utilize the default endorsement policy of majority of channel members.
```
nikhil550, line 37: utilize > use

token-account-based/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+The smart contract supports accounts owned by individual client identities from organizations that are members of the channel. In our scenario, the minter of the tokens will be a member of Org1, while the recipient will belong to Org2. To highlight the connection between the `GetClientIdentity().GetID()` API and the information within a user's certificate, we will register two new identities using the Org1 and Org2 Certificate Authorities (CA's), and then use the CA's to generate each identity's certificate and private key.
+
+First, we need to set the following environment variables to use the the Fabric CA client (and subsequent commands).
+```
+export PATH=${PWD}/../bin:${PWD}:$PATH
+export FABRIC_CFG_PATH=$PWD/../config/
+```
+
+We will use the Org1 CA to create the minter identity. Set the Fabric CA client home to the MSP of the Org1 CA admin (this identity was generated by the test network script):
```
nikhil550, line 52: It would provide more set up work, but would it be worth it to add another recipient from org2 and allow recipient 1 to transfer to some tokens to recipient 2. It would further demonstrate that the token is being held at the individual level, and not the org level.

nikhil550 (commented):
Ran the sample on my local machine. Looks good, but left some nits on the readme.

token-account-based/chaincode-go/chaincode/token_contract.go
```diff
@@ -0,0 +1,173 @@
...
+	clientCurrentBalanceBytes, err := ctx.GetStub().GetState(clientID)
+	if err != nil {
+		return fmt.Errorf("failed to read client account %s from world state: %v", clientID, err)
+	}
+
+	if clientCurrentBalanceBytes == nil {
+		return fmt.Errorf("client account %s has no balance", clientID)
+	}
+
+	clientCurrentBalance, _ := strconv.Atoi(string(clientCurrentBalanceBytes))
```
sijocherian, line 87: good to add error handling here and line 103, safety check for cases when value is not int

token-account-based/chaincode-go/chaincode/token_contract.go
```diff
@@ -0,0 +1,173 @@
...
+		return fmt.Errorf("failed to read minter account %s from world state: %v", minter, err)
+	}
+
+	var currentBalance int
+
+	// If minter current balance doesn't yet exist, we'll create it with a current balance of 0
+	if currentBalanceBytes == nil {
+		currentBalance = 0
+	} else {
+		currentBalance, _ = strconv.Atoi(string(currentBalanceBytes))
```
sijocherian, line 49: good to add error handling on Atoi return

sijocherian (changes requested):

token-account-based/chaincode-go/chaincode/token_contract.go
lehors: True but trying to mint a non integer amount generates an error so I don't think the extra check is really necessary:

`Error: endorsement failure during invoke. response: status:500 message:"Error managing parameter param0. Conversion error. Cannot convert passed value 1.5 to int" `

token-account-based/chaincode-go/chaincode/token_contract.go
lehors: Same here: Trying to transfer a non integer amount generates an error so I don't think the extra check is really necessary:

`Error: endorsement failure during invoke. response: status:500 message:"Error managing parameter param1. Conversion error. Cannot convert passed value 1.5 to int"`

token-account-based/chaincode-go/chaincode/token_contract.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	balance, _ := strconv.Atoi(string(balanceBytes))
+
+	return balance, nil
+}
+
+// ClientID returns the client id of the transaction creator
+// Users can use this function to get their own client id, which they can then give to others as the payment address
+func (s *SmartContract) ClientID(ctx contractapi.TransactionContextInterface) (string, error) {
```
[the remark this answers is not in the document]
lehors: I think there is value in that suggestion. The fact that the client ID is used as the account ID can be considered as an implementation detail. From an application point of view, it is the account ID. The README might need to be updated accordingly to make that clear though.

lehors (commented):
LGTM overall. Just a few suggestions for improvement along those made by other reviewers.

denyeart:
Thanks folks. I'll make the updates here. And I will also add the remaining ERC20 functions to make it functionally identical with the Javascript PR: https://github.com/hyperledger/fabric-samples/pull/327

token-account-based/README.md
nikhil550: in the private data tutorial, we refer to the common name and issuer of the identity certificate.

token-account-based/chaincode-go/chaincode/token_contract.go
```diff
@@ -0,0 +1,173 @@
...
+		return fmt.Errorf("failed to read client account %s from world state: %v", clientID, err)
+	}
+
+	if clientCurrentBalanceBytes == nil {
+		return fmt.Errorf("client account %s has no balance", clientID)
+	}
+
+	clientCurrentBalance, _ := strconv.Atoi(string(clientCurrentBalanceBytes))
+
+	if clientCurrentBalance < amount {
```
sijocherian, line 89: ERC20 allows transfer amount of 0 ("Transfers of 0 values MUST be treated as normal transfers and fire the Transfer event."), so this should allow even if clientCurrentBalance==0

sijocherian (commented):
ERC20 interface signature sounds better for adoption.

We may want to be conscious of the Terminology we use in the sample: like AccountID, ClientID , Owner , User, Recipient

ERC20 uses the terminology "balance of an account with address _owner"
https://eips.ethereum.org/EIPS/eip-20

Account & Balance seem like the most compatible ones. I am on the fence about "address"

token-account-based/README.md
denyeart: Technically it is the distinguished name that is used, not the common name, but I agree we don't have to get technical here, removed 'distinguished'.

token-account-based/README.md
denyeart: Sure, will do in another PR.

token-account-based/chaincode-go/chaincode/token_contract.go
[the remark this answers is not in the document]
denyeart: How about ClientAccountID.  I do want to drive home the fact that the chaincode gets the calling client identity, otherwise "My" may not be clear.

token-account-based/chaincode-go/chaincode/token_contract.go
```diff
@@ -0,0 +1,173 @@
...
+		return 0, fmt.Errorf("the account %s does not exist", account)
+	}
+
+	balance, _ := strconv.Atoi(string(balanceBytes))
+
+	return balance, nil
+}
+
+// ClientBalance returns the balance of the requesting client
+func (s *SmartContract) ClientBalance(ctx contractapi.TransactionContextInterface) (int, error) {
```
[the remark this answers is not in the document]
denyeart: How about ClientAccountBalance.

token-account-based/chaincode-go/chaincode/token_contract.go
denyeart: I was trying to keep it as compact as possible so that it can be easily read without distracting error handling.
I've added a comment as a compromise:
// Error handling not needed since Itoa() was used when setting the account balance, guaranteeing it was an integer.

token-account-based/chaincode-go/chaincode/token_contract.go
denyeart: I was trying to keep it as compact as possible so that it can be easily read without distracting error handling.
I've added a comment as a compromise:
// Error handling not needed since Itoa() was used when setting the account balance, guaranteeing it was an integer.

token-account-based/chaincode-go/chaincode/token_contract.go
denyeart: ok, I changed line 68 from `if amount <= 0` to `if amount < 0`

token-account-based/README.md
```diff
@@ -0,0 +1,200 @@
...
+```
+fabric-ca-client enroll -u https://minter:minterpw@localhost:7054 --caname ca-org1 -M ${PWD}/organizations/peerOrganizations/org1.example.com/users/minter@org1.example.com/msp --tls.certfiles ${PWD}/organizations/fabric-ca/org1/tls-cert.pem
+```
+
+Run the command below to copy the Node OU configuration file into the minter identity MSP folder.
+```
+cp ${PWD}/organizations/peerOrganizations/org1.example.com/msp/config.yaml ${PWD}/organizations/peerOrganizations/org1.example.com/users/minter@org1.example.com/msp/config.yaml
+```
+
+We can now use the Org2 CA to create the Org2 recipient identity. Set the Fabric CA client home the Org2 CA admin:
```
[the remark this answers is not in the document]
denyeart: good idea, done.

lehors (approved):
LGTM!

PR #323, opened 2020-09-02
Readme updates

nikhil550:
Updating the Readme to reflect new application and smart contract samples

Also getting off the fence to say we deprecate interest rate samples. With its use of the BYFN network, tools container, low level programming smart contract model, I think we should steer folks to the new sample.


lehors (approved):
LGTM.

PR #332, opened 2020-09-15
Fix help messages

ikegawa-koshi:
In the past, there have been committed to change the output method of help messages from "echo" to "println”.
Dou to that change, network.sh not shown help messages correctly.
When changing the method of output help message "echo" to "println", the message format should be fixed.

Signed-off-by: ikegawa-koshi <koshi.ikegawa.mf@hitachi.com>


nikhil550 (approved):
Much better!

PR #338, opened 2020-09-18
Move off chain data sample to basic chaincode

nikhil550:
Move off chain data sample off the marbles chaincode to the basic chaincode. One less dependency one the older samples, and simplifying the script.

One problem is that the couchDB component does not seem to work any more, probably with changes in couch. I am going to investigate whether any of my changes contributed but I am skeptical. I would be happy to have any couch experts pitch in.


off_chain_data/README.md
```diff
@@ -25,7 +25,7 @@ Use the following command to start the sample network:
 This command will deploy an instance of the Fabric test network. The network
 consists of an ordering service, two peer organizations with one peers each, and
 a CA for each org. The command also creates a channel named `mychannel`. The
-marbles chaincode will be installed on both peers and deployed to the channel.
+assets chaincode will be installed on both peers and deployed to the channel.
```
nikhil550, line 28: Wrong chaincode name, will fix.

harrisob: Should we say the full name of the chaincode, "asset transfer basic"

off_chain_data/addAssets.js
```diff
@@ -0,0 +1,118 @@
...
+
+        // Configure a wallet. This wallet must already be primed with an identity that
+        // the application can use to interact with the peer node.
+        const walletPath = path.resolve(__dirname, 'wallet');
+        const wallet = await Wallets.newFileSystemWallet(walletPath);
+
+        // Create a new gateway, and connect to the gateway peer node(s). The identity
+        // specified must already exist in the specified wallet.
+        const gateway = new Gateway();
+        await gateway.connect(ccp, { wallet, identity: 'appUser', discovery: { enabled: true, asLocalhost: true } });
```
harrisob, line 83: I do not see where the user gets added to the wallet, we should indicate what needs to run first to get the user added to the wallet directory

off_chain_data/README.md
nikhil550: Yup!

off_chain_data/addAssets.js
nikhil550: There is an `enrollAdmin.js` and `registerUser.js` in the directory. Those identities are used for this application and the listener.

off_chain_data/README.md
nikhil550: done.

nikhil550:
Tested the sample on master, and it has the same problems with couchDB as I encountered here. However, I would welcome those who know couch well to look at what the problem is, and fix in this pr or on a follow on.

The problem is that the database is not created, so when you travel to http://127.0.0.1:5990/mychannel_basic/_all_docs, the database does not exist. As a result, the curl commands provided do not work. My guess this is a couchDB version issue. Note that I pull 2.3.1 since 3.0 requires operating as an admin.

harrisob (approved):

denyeart (approved):
Thanks Nik, I'll merge and make a note to check the CouchDB issue.

PR #339, opened 2020-09-19
asset transfer basic - application typescript

rijulaggarwal16:
run this application by running `npm start` in the `application-typescript` directory inside `asset-transfer-basic`

Signed-off-by: Rijul Aggarwal <rijul.aggarwal@ibm.com>


denyeart:
https://github.com/hyperledger/fabric-samples/pull/246 previously removed the application for typescript because it was essentially the same as the javascript application, and therefore the conclusion was it did not add value, but added overhead.

Let's hear the arguments for and against having an application typescript sample again... what do you think @btl5037 @harrisob ?

btl5037:
This code is exactly the same as the javascript sample. Nearly line for line. If we were going to maintain a typescript version, I'd like to hear an argument for how the typescript version differs from the javascript version itself. Since typescript is just a compiler (that compiles typescript to javascript), the result of compiling this typescript file is the same code as the javascript implementation.

rijulaggarwal16:
I agree. Although if needed, we can add definitions for the object interfaces of wallet, ccp, connection profile etc. to strongly type this. However, I'm not sure even in those cases if it makes sense if those type definitions are arbitrary like a json/key-value dictionary.  If not, do we have those definitions available in the imported packages or do we have to build those definitions somehow?

harrisob (commented):
I think we should add a typescript sample if it includes all the setup written in typescript. Would you write this sample to NOT use the common javascript setup of the user and gateway and include how it would be done using types with typescript. Thanks.

harrisob:
Also the code should be run during build, please update the "fabric-samples/ci/scripts/run-test-network-basic.sh" script to include the compile and execution of the new sample.

rijulaggarwal16:
The PR is ready for review. Changes:
1. Added to CI script
2. Strongly typed code with its own setup.

asset-transfer-basic/application-typescript/package.json
```diff
@@ -0,0 +1,62 @@
...
+        "prepublishOnly": "npm run build"
+    },
+    "engineStrict": true,
+    "author": "Hyperledger",
+    "license": "Apache-2.0",
+    "dependencies": {
+        "fabric-ca-client": "^2.2.0",
+        "fabric-network": "^2.2.0"
+    },
+    "devDependencies": {
```
harrisob, line 27: Could we clean this up to be just what is needed, I do not see any "mocha" test ?

asset-transfer-basic/application-typescript/src/app.ts
```diff
@@ -0,0 +1,169 @@
...
+        // a user that has been verified.
+        const gateway = new Gateway();
+
+        try {
+            // setup the gateway instance
+            // The user will now be able to create connections to the fabric network and be able to
+            // submit transactions and query. All transactions submitted by this gateway will be
+            // signed by this user using the credentials stored in the wallet.
+            await gateway.connect(ccp, {
+                wallet,
```
harrisob, line 94: Could we create the "GatewayOptions" object before calling the "connect", I think it would be good to show how to set the different options in TypeScript

export interface GatewayOptions {
	identity: string | Identity;
	wallet?: Wallet;
	identityProvider?: IdentityProvider;
	clientTlsIdentity?: string;
	tlsInfo?: {
		certificate: string;
		key: string;
	};
	discovery?: DiscoveryOptions;
	eventHandlerOptions?: DefaultEventHandlerOptions;
	queryHandlerOptions?: DefaultQueryHandlerOptions;
	'connection-options'?: any;
}

export interface ConnectedGatewayOptions extends GatewayOptions {
	discovery: DiscoveryOptions;
	eventHandlerOptions: DefaultEventHandlerOptions;
	queryHandlerOptions: DefaultQueryHandlerOptions;
}

export interface DiscoveryOptions {
	asLocalhost?: boolean;
	enabled?: boolean;
}

export interface DefaultEventHandlerOptions {
	commitTimeout?: number;
	endorseTimeout?: number;
	strategy?: TxEventHandlerFactory | null;
}

export interface DefaultQueryHandlerOptions {
	strategy?: QueryHandlerFactory;
	timeout?: number;
}

PR #341, opened 2020-09-22
Fix Error message application-go

kgeonw:
[FAB-18236] The description of the error is incorrect. 
An error must be issued according to the transaction type.

* fix Error message
* add additional logs to transaction

Signed-off-by: KIM GEONWOO <kgw8919@gmail.com>


lehors (approved):
LGTM. Thanks!

PR #349, opened 2020-10-14
Fix token-utxo-based README terminal 2 setup

lehors:
The user's PATH must be set on the first use of Terminal 2 for the
call to the fabric-ca-client command to succeed. The current README
only sets it later. This fix changes that.


denyeart (approved):
Thanks!

PR #351, opened 2020-10-14
Add type parameters

suraneti:
If you accept this PR please add `hacktoberfest-accepted` label on this PR.

Signed-off-by: Suraneti Rodsuwan <suraneti.rod@gmail.com>

### DIFF
--- a/asset-transfer-ledger-queries/chaincode-go/asset_transfer_ledger_chaincode.go
+++ b/asset-transfer-ledger-queries/chaincode-go/asset_transfer_ledger_chaincode.go
@@ -112,7 +112,7 @@ type PaginatedQueryResult struct {
 }
 
 // CreateAsset initializes a new asset in the ledger
-func (t *SimpleChaincode) CreateAsset(ctx contractapi.TransactionContextInterface, assetID, color string, size int, owner string, appraisedValue int) error {
+func (t *SimpleChaincode) CreateAsset(ctx contractapi.TransactionContextInterface, assetID string, color string, size int, owner string, appraisedValue int) error {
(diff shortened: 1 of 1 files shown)


lehors:
Hi,
Can you please explain what the point of that change is? (other than trying to land a PR for the Hacktoberfest ;-)
The way the code is currently written is perfectly legal Go code and doesn't even trigger any warning or similar from golint.

suraneti:
Hello,
I agree with you that the code is working perfectly but I think it would be good to make all function look the same code style, 
 some function parameters have a type but some haven't.

Ex.1 this func `assetID` parameter has type.
```Go
func (t *SimpleChaincode) AssetExists(ctx contractapi.TransactionContextInterface, assetID string) (bool, error) {
...
}
```

Ex.2 this func `assetID` parameter doesn't have type.
```Go
func (t *SimpleChaincode) TransferAsset(ctx contractapi.TransactionContextInterface, assetID, newOwner string) error {
...
}
```

P.S. I'm starting studying HyperLedger Fabric tutorial with Golang.

denyeart:
Thank you for the contribution, but all parameters have a type already - in Go it is idiomatic to not repeat the parameter type for multiple parameters in a row that have the same type.
Therefore let's Close this PR.

denyeart:
I just checked asset-transfer-basic and I forgot that it actually does enumerate the type for each of the parameters, so this PR would make the samples consistent. My apologies @suraneti ... I do agree it would be good to be consistent and I'm open to discussion on which approach is best.

suraneti:
@denyeart Thank you for the explanation and rechecking, Have a great day.

PR #353, opened 2020-10-16
update high throughput sample to use deployCC script

nikhil550:
Update the high throughput sample to use the deployCC script, rather than its own scripts. This will cut down on duplicate code and make the sample easier to maintain going forward.

Signed-off-by: Nikhil Gupta <ngupta@symbridge.com>


harrisob (approved):

lehors (approved):
LGTM.

PR #325, opened 2020-09-04
Add Asset-Transfer-Events

harrisob:
Add new chaincode and javascript application to demostrate
the use of events and events with private data.

Signed-off-by: Bret Harrison <beharrison@nc.rr.com>

PR #354, opened 2020-10-18
Java chaincode for asset transfer events sample

sijocherian:
- Java chaincode for Event sample ( use of events and events with private data).
- Complements #325


denyeart (approved):
Great contribution Sijo. Thanks for adding a README as well!

PR #364, opened 2020-10-27
Move high throughput sample to go application

nikhil550:
The High Throughput sample currently uses scripts to push transactions to the network. The peer CLI cannot push concurrent transactions, and so does not make good use of the smart contract.

Moving this sample to an application that is built using the Go SDK, allowing the users to push concurrent transactions using a GO routine.


lehors (approved):
Well done, this is a nice improvement.
This said, I can't seem to get any updates through at all using the traditional method, while with the scripts we typically got some through. I take it that this is because with the app the transactions are fired much closer to one another than with the script, making the possibility of a conflict very high. This makes me wonder whether we lost some educational aspect because the result now seems to be entirely predictable. Should we consider adding a delay parameter that would allow the user to simulate some latency and which would lead to some transactions making it through for a less predictable result?
Otherwise, I suggest adding a log of the peer to show the reason for the lack of udpate.
But none of this is mandatory in my view so I'll approve anyway so it can be merged as is if this is felt like the right thing to do.
Thanks.

nikhil550:
When I run the original scripts, it provides the same result that all changes are rejected:
```
./scripts/get-invoke.sh testvar
Error: endorsement failure during invoke. response: status:500 message:"No variable by the name testvar exists" 
```

The suggestion to add the peer logs was a good one and I included in this PR.

lehors:
> When I run the original scripts, it provides the same result that all changes are rejected:
> 
> ```
> ./scripts/get-invoke.sh testvar
> Error: endorsement failure during invoke. response: status:500 message:"No variable by the name testvar exists" 
> ```
> 
> The suggestion to add the peer logs was a good one and I included in this PR.

Ok, I haven't recently tried the scripts but it used to work that way for sure. That's why the README said that the results were undefined rather than give a specific value. But I guess that's ok, let's go with this. I'll merge it.

lehors (approved):

PR #370, opened 2020-11-11
Add dutch auction sample

nikhil550:
Add the dutch auction to the auction sample, alongside the simple blind auction


nikhil550:
The dutch auction uses many of the same features as the simpler auction, with a somewhat more complicated smart contract. However, I think adding a more complicated auction could help drive interest for developing examples that are common blockchain use cases such as online auctions (EV charging) or double auctions.

denyeart:
I like the approach of having both, but haven't had time to review yet.

nikhil550:
> I like the approach of having both the simple auction and dutch auction, but haven't had time to review yet.

We also discussed adding and auditor to the auction using the longer signature syntax. It might be best to add an auditor to this auction to keep the first auction simple, and provide another technical feature to this one.

nikhil550:
Closing this to do some code cleanup.

PR #371, opened 2020-11-16
Rename token-account-based Directory to token-erc20

jcastrence:
#### Type of change
- Documentation update
#### Description
- Changed name of sample directory token-account-based to token-erc20
#### Related issues
- [FAB-18275](https://jira.hyperledger.org/browse/FAB-18275)


lehors (dismissed):
Although ERC20 is widely in use, the name of the standard is ERC-20. We should use that consistently throughout the sample. Right now we are mixing the two which is the worst.

jcastrence:
> Although ERC20 is widely in use, the name of the standard is ERC-20. We should use that consistently throughout the sample. Right now we are mixing the two which is the worst.

denyeart (approved):
Nice work - I've tested locally and everything looks good. I'll update the erc-20 readme text in a next PR to match the changes here.

PR #376, opened 2020-11-20
Mount configuration files in the test network

nikhil550:
When going through the new channel participation tutorial, I realize that we do not mount the configuration file in the test network containers, instead using the files that are inside the container. In addition to not being best practice, it makes it hard to update the containers.

Signed-off-by: Nikhil Gupta <ngupta@symbridge.com>

### DIFF
--- a/test-network/docker/docker-compose-test-net.yaml
+++ b/test-network/docker/docker-compose-test-net.yaml
@@ -20,6 +20,7 @@ services:
     image: hyperledger/fabric-orderer:$IMAGE_TAG
     environment:
       - FABRIC_LOGGING_SPEC=INFO
+      - FABRIC_CFG_PATH=/var/hyperledger/orderer
       - ORDERER_GENERAL_LISTENADDRESS=0.0.0.0
       - ORDERER_GENERAL_LISTENPORT=7050
       - ORDERER_GENERAL_GENESISMETHOD=file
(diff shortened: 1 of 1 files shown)


test-network/docker/docker-compose-test-net.yaml
```diff
@@ -76,6 +79,7 @@ services:
       - CORE_PEER_LOCALMSPID=Org1MSP
     volumes:
         - /var/run/:/host/var/run/
+        - ../../config/core.yaml:/etc/hyperledger/fabric/core.yaml
```
sijocherian, line 82: if you mount the same ../../config/core.yaml file, to both org1 & org2 peer, will it work?

```
../../config/core.yaml file has a id as jdoe & listenAddress at port 7051

peer:
    id: jdoe
    listenAddress: 0.0.0.0:7051

```

nikhil550: Hi Sijo, The important fields would be overridden by environment variables, which are different between the two orgs. The config files in the two peers are the same for both orgs in the current deployment as well.

sijocherian (approved):
lgtm

lehors (approved):
I tested it and it ran without problems.

SamYuan1990:
is this changes supports fabric v2.2?

denyeart:
@SamYuan1990 You're right, this change will break things if you use the Fabric download script (which would currently give you 2.3 config) and then try to run the v2.2 images using the test-network -i flag.  It is always best to run images with the same version of the config, that's why the version-specific images have the version-specific config built in. I'm going to open a PR to revert this change.

PR #380, opened 2020-11-25
Revert "Mount configuration files in the test network"

denyeart:
Reverts hyperledger/fabric-samples#376

The change will break things if you use the Fabric download script (which would currently give you 2.3 config) and then try to run the v2.2 images using the test-network -i flag. Unfortunately the orderer crashes if there is config it doesn't understand. While that's not good, regardless it is always best to run images with the same version of the config, that's why the version-specific images have the version-specific config built in. The test-network should use the config built into the images as the starting defaults so that things always work by default.

@nikhil550 @lehors FYI


harrisob (approved):

nikhil550:
Understood. Thank you for cleaning this up.

PR #382, opened 2020-12-01
Enable test-network/network.sh to easily run docker-compose commands

g2flyer:
Small patch which makes it much easier to trouble-shoot issues in the containers by easily be able to run docker(-compose) commands, e.g., `ps` or `logs` on the started containers or even logging into them.

### DIFF
--- a/test-network/network.sh
+++ b/test-network/network.sh
@@ -289,6 +289,16 @@ function networkUp() {
   fi
 }
 
+function dockerComposeCommand() {
+
+  if [ "$CRYPTO" == "Certificate Authorities" ]; then
+      IMAGE_TAG=${CA_IMAGETAG} docker-compose -f $COMPOSE_FILE_CA -f $COMPOSE_FILE_CA_ORG3 "${ARGS[@]}"
+  else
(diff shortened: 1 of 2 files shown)


denyeart:
@g2flyer thanks for the contribution. I'm trying to understand the benefit... I frequently do "docker ps" or "docker logs" in a separate terminal while the network is up and it meets all my needs. What additional benefit does this change provide?

g2flyer:
> @g2flyer thanks for the contribution. I'm trying to understand the benefit... I frequently do "docker ps" or "docker logs" in a separate terminal while the network is up and it meets all my needs. What additional benefit does this change provide?

it allows you to use the symbolic names from docker-compose files to address  containers rather than having to look up the container id. It also allows you to filter on just relevant containers for a network instance rather than seeing also all kind of other containers which are running ...

denyeart:
@g2flyer you don't actually have to find the container id. You can use the name from the docker compose files. For example I frequently do:
`docker logs peer0.org1.example.com`

nikhil550:
From a user perspective, I also think there is a benefit from users learning how to use docker more directly to inspect the nodes and get logs.

g2flyer:
> @g2flyer you don't actually have to find the container id. You can use the name from the docker compose files. For example I frequently do:
> `docker logs peer0.org1.example.com`

Oops, how embarrassing.  I've overlooked that you define the container_name to be identical to the symbolic name.  The patch would also help to get things like IMAGE_TAGs consistent and allow network-wide operations (e.g., logs of all containers but ignoring any other container running) without having to know (or at least specify) the set of relevant yml files but that is certainly a very minor benefit.
=>
I'll close the PR ...

PR #383, opened 2020-12-01
Fix test-network chaincode packaging log

lehors:
The chaincode packaging function wrongly implies that the packaging
operation is performed on a peer when it is in fact done locally.
This change fixes that.


nikhil550 (approved):
Good update. The MSP is no longer required to package a smart contract

PR #388, opened 2020-12-07
Remove Short Name Support in Scripts

jcastrence:
Signed-off-by: Julian Castrence <juliancastrence@ibm.com>


nikhil550:
@jcastrence If you want to go this route, we would need to fix the samples, backing tests, and the docs. 

On the substance, I do not have a strong opinion either way, but when the deployment script was re-written, which samples had short paths and which ones did not was bit less random than it is today.

PR #389, opened 2020-12-09
Add support for HSM based identities

harrisob:
Add support to the basic application to check for the HSM environment and
enroll and store the certificate and HSM handle of the  HSM identity in the
wallet. The identity will then be retrieved form the wallet providing access
for signing.

### DIFF
--- a/asset-transfer-basic/README.md
+++ b/asset-transfer-basic/README.md
@@ -1,0 +1,142 @@
+# Asset Transfer Basic javascript Sample
+
+The asset transfer basic sample demonstrates submitting transactions
+and queries.
+
+
+## About the Sample
+
+This javascript sample application is able to use any of the asset transfer basic
+chaincode samples. It will show how to enroll an admin identity. The admin will then
+register and enroll a user identity to work with the Fabric network.
+The application will demostrate the basic operations of create, update, delete,
+and query on an asset.
+
+
+## Running the sample
+
+Use the Fabric test network utility (`network.sh`) to start a Fabric network and deploy one
+one of the sample chaincodes.
+
+Follow these step in order.
+- Start the test network with a Certificate Authority and create a channel
+```
+cd test-network
+./network.sh up createChannel -c mychannel -ca
+```
+
+- Deploy the chaincode (smart contract)
+```
+# to deploy the javascript version
+./network.sh deployCC -ccs 1 -ccv 1 -ccep "OR('Org1MSP.peer','Org2MSP.peer')" -ccl javascript -ccp ./../asset-transfer-basic/chaincode-javascript/ -ccn basic
+```
+
+- Run the application
+```
+cd asset-transfer-basic/application-javascript
+npm install
+node app.js
+```
+
+If you see an error:
+```
+    2020-08-07T20:23:17.590Z - error: [DiscoveryService]: send[mychannel] - Channel:mychannel received discovery error:access denied
+    ******** FAILED to run the application: Error: DiscoveryService: mychannel error: access denied
+```
+   or
+```
+   Failed to register user : Error: fabric-ca request register failed with errors [[ { code: 20, message: 'Authentication failure' } ]]
+   ******** FAILED to run the application: Error: Identity not found in wallet: appUser
+```
+Delete the `/fabric-samples/asset-transfer-basic/application-javascript/wallet` directory
+and retry this application.
+
+The certificate authority must have been restarted and the saved certificates for the
+admin and application user are not valid. Deleting the wallet store will force these to be reset
+with the new certificate authority.
+
+To see the SDK workings, try setting the logging to show on the console before running
+```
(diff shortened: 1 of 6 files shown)


btl5037 (changes requested):
Can you clear up what this is trying to achieve. You initialize the HSM crypto suite, but this isn't an HSM-enabled CA, so the CA exhibits it's default behavior and doesn't actually use the HSM to do anything, the private key is generated and then lost except to the CA client, the CA no longer has a reference to it itself.

I think if your purpose here is to demo the use of the HSM library in the SDK, then it should be a documentation sample, and not a sample in the test network. If you really want to demo the HSM usage it should be with an HSM enabled network, you can use SoftHSM to do that.

The fact that the "simulator" can only be run once (because there really is no private key stored in the HSM), and the fact that the private key doesn't exist in the HSM is only going to cause more confusion. One of the most asked support questions we get asked in RocketChat is how to configure SoftHSM for Fabric. If we put this example that doesn't actually use the SoftHSM other than to trick the CA Client into thinking there is an HSM, people are going to start using this example thinking they are doing something with the HSM enabled and start seeking more help on where the private keys are and why they can't see them in the HSM.

I already publish HSM enabled versions of all of the Fabric components using SoftHSM and have even rewritten the test-network to use SoftHSM for real: https://github.com/hyperledger/fabric-test/tree/master/regression/hsm

I don't think this does anything that good documentation couldn't do, other than confuse people with an already difficult concept.

harrisob:
To address the comment from "lindluni" 
The CA does not need and will not have access to the  private key in any form, HSM or not. We create a separate instance of the FabricCA service class, one that has crypto code to work with an HSM locally on the client, however this instance will still use the same CA as the instance of a non-HSM instance. 

This sample code is not meant to show how to run an entire network on an HSM, just how to setup and run from a client.

asset-transfer-basic/README.md
```diff
@@ -0,0 +1,142 @@
...
+variable to point to a config file before running the javascript application.
+When the javascript application sees the following environment variable, it will
+attempt to connect to the HSM simulator.
+
+```bash
+export SOFTHSM2_CONF="../test-network/hsm/softhsm2.conf"
+```
+
+### Create a token to store keys in the HSM
+
```
nikhil550, line 126: We may need more instructions on how to use softHSM. For instance, I issued the following command: 
```
Available slots:
Slot 742340698
    Slot info:
        Description:      SoftHSM slot ID 0x2c3f385a                                      
        Manufacturer ID:  SoftHSM project                 
        Hardware version: 2.6
        Firmware version: 2.6
        Token present:    yes
    Token info:
        Manufacturer ID:  SoftHSM project                 
        Model:            SoftHSM v2      
        Hardware version: 2.6
        Firmware version: 2.6
        Serial number:    5e1731c1ac3f385a
        Initialized:      yes
        User PIN init.:   yes
        Label:            ForFabric                       
Slot 1
    Slot info:
        Description:      SoftHSM slot ID 0x1                                             
        Manufacturer ID:  SoftHSM project                 
        Hardware version: 2.6
        Firmware version: 2.6
        Token present:    yes
    Token info:
        Manufacturer ID:  SoftHSM project                 
        Model:            SoftHSM v2      
        Hardware version: 2.6
        Firmware version: 2.6
        Serial number:                    
        Initialized:      no
        User PIN init.:   no
        Label:                                  
```

but was still able to use slot 0

asset-transfer-basic/README.md
```diff
@@ -0,0 +1,142 @@
...
+ledger data that you created:
+
+```
+./network.sh down
+```
+Be sure to delete the wallet directory create by the application.
+The stored identities will no longer be valid when restarting the network.
+
+## Configuring and running a Hardware Security Module
+
```
nikhil550, line 79: I would split this readme into two, and quickly restate the commands above so that folks are sure when to complete the steps below.

test-network/hsm/softhsm2.conf
```diff
@@ -0,0 +1,7 @@
+# SoftHSM v2 configuration file
```
nikhil550, line 1: I think putting this in the test-network directory would just cause more confusion, given that this is used by the application and not the infrastructure.

This can be moved to the app directory, or users can even create this file themselves given how small it is.

nikhil550 (commented):
I ran through and got it to work. Having an application sample that uses an HSM to sign is a good add. However, I think it would be better if it was a separate application to reduce the clutter in the basic application. This is the first sample many people will use. I think having lines that setup an HSM option would only add confusion. I think a separate `appHSM.json` app would be a better user experience. Users could create or add the configuration file in this directory.

davidkel:
This PR is replaced by #455

nikhil550:
Going to close as stale. Replaced by https://github.com/hyperledger/fabric-samples/pull/455

PR #390, opened 2020-12-11
Completely Remove Short Name Support

jcastrence:
#### Type of change

#### Description

- Short names are a crutch that can prevent someone from learning how to deploy their own chaincode. Short names have been replaced with their full path name. Short names are no longer recognized by shell scripts.


test-network/scripts/deployCC.sh
```diff
@@ -33,51 +33,8 @@ CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
 
 FABRIC_CFG_PATH=$PWD/../config/
 
-# User has not provided a path, therefore the CC_NAME must
```
nikhil550, line 36: This may be personal preference, but I would keep the default option of using deploying the basic chaincode in go if no other default is provided. This would make basic testing easier. Users would need to supply name and path for all other chaincodes

ci/scripts/run-test-network-basic.sh
```diff
@@ -3,6 +3,7 @@ set -euo pipefail
 FABRIC_VERSION=${FABRIC_VERSION:-2.2}
 CHAINCODE_LANGUAGE=${CHAINCODE_LANGUAGE:-go}
 CHAINCODE_NAME=${CHAINCODE_NAME:-basic}
+CHAINCODE_PATH=${CHAINCODE_PATH:-../asset-transfer-basic/chaincode-go/}
```
nikhil550, line 6: The tests are failing because you need to pass the paths of the other chaincode languages, javascript, java, typescript, to this script.

nikhil550 (approved):

test-network/scripts/deployCC.sh
jcastrence: This makes sense to me, @denyeart thoughts on this?

ci/scripts/run-test-network-basic.sh
jcastrence: Fixed

test-network/scripts/deployCC.sh
denyeart: @nikhil550 The whole reason we are removing the short name support is so that users learn a single way to deploy chaincode using the test network from the beginning, so that they can easily understand how to deploy their own chaincodes. I think most users will copy/paste the entire command when running the tutorials, so it doesn't seem to save typing. And once they get familiar with the command, it is an easy mental leap to replace the name/path/language with their own chaincode. Does that make sense?

test-network/scripts/deployCC.sh
```diff
@@ -33,51 +33,8 @@ CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
...
-# be the short name of a known chaincode sample
-if [ "$CC_SRC_PATH" = "NA" ]; then
```
denyeart, line 38: Let's see what @nikhil550 thinks... if we decide to remove all the chaincode defaults in favor of client passing them, then the defaults in lines 6-8 should be removed and we'd need to validate that client passed in the values.

test-network/scripts/deployCC.sh
nikhil550: The change makes sense. My only thought is that a default would make it easier to test a new version or code change, ie with:
```
./network.sh up createChannel
./network.sh deploy
```

and that it would not be a problem if only one path was provided as a default and users had to learn the others, but if it would only add confusion it would be better to remove. If the default is removed, this PR removes the only use of the shortnames in the samples: https://github.com/hyperledger/fabric-samples/pull/392

denyeart: @nikhil550 Thanks for fixing that last remaining default usage #392 , I've merged it.

My other concern with leaving the defaults in, is if somebody forgets to pass the name, path, or language, it will simply take the basic sample default and they may be confused when something goes wrong. If there are no defaults, then we can provide a clear message indicating which property is missing. My suggestion is to remove the defaults and make it explicit so that there is no chance for confusion.

lehors: I was leaning towards having the default but Dave's idea sounds pretty good. In case nothing is specified the error message could literally indicate the command to type to have "the default".

nikhil550: sounds good to me

test-network/scripts/deployCC.sh
```diff
@@ -33,54 +33,31 @@ CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
...
 fi
 
+# User has not provided a language
+if [ -z "$CC_SRC_LANGUAGE" ]; then
+  fatalln "No chaincode language was provided. Valid call example: ./network.sh deployCC -ccn \"basic\" -ccp \"../asset-transfer-basic/chaincode-go\" -ccl \"go\""
+fi
+
+# check that the language is available for the sample chaincode
+if [ ! -d "$CC_SRC_PATH" ]; then
+  fatalln "The smart contract language \"$CC_SRC_LANGUAGE\" is not yet available for the \"$CC_NAME\" sample smart contract"
```
nikhil550, line 58:
```
if [ ! -d "$CC_SRC_PATH" ]; then
  fatalln "The smart contract language \"$CC_SRC_LANGUAGE\" is not yet available for the \"$CC_NAME\" sample smart contract"
```

This is duplicative with the if statement above and should be removed. The statement that the language is not available is only relevant for the asset-trasnfer samples that could be called with short names.

test-network/scripts/deployCC.sh
```diff
@@ -33,54 +33,31 @@ CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
...
-    println $'\e[0;32m'asset-transfer-private-data$'\e[0m' chaincode
-    CC_SRC_PATH="../asset-transfer-private-data"
-  elif [ "$CC_NAME" = "sbe" ]; then
-    println $'\e[0;32m'asset-transfer-sbe$'\e[0m' chaincode
-    CC_SRC_PATH="../asset-transfer-sbe"
-  else
-    fatalln "The chaincode name ${CC_NAME} is not supported by this script. Supported chaincode names are: basic, events, ledger, private, sbe, secured"
-  fi
+#User has not provided a name
+if [ -z "$CC_NAME" ]; then
```
nikhil550, line 37: When I issued a command without any parameters, the script was only printing: ` fatalln "Path to chaincode does not exist. Please provide different path"` without the call example. I would make these series of If statements into a series of If then statements so you can have more control over what is printed.

jcastrence: I'm not as familiar with shell scripts, could you please elaborate on your suggestion?

nikhil550:
I was playing around with the error text,a nd I realized that the defaults are still set in the main `network.sh` script:
```
# chaincode name defaults to "basic"
CC_NAME="basic"
# chaincode path defaults to "NA"
CC_SRC_PATH="NA"
# use go as the default language for chaincode
CC_SRC_LANGUAGE="go"
```

I would remove those from that script, and then set the defaults to "NA" in the first lines of the deployCC script. You can then set the following if statements accordingly

jcastrence:
@nikhil550 I noticed that when I was playing with the script but couldn't figure out why it was still setting them to the defaults. Thanks for pointing that out.

nikhil550:
> I was playing around with the error text,a nd I realized that the defaults are still set in the main `network.sh` script:
> 
> ```
> # chaincode name defaults to "basic"
> CC_NAME="basic"
> # chaincode path defaults to "NA"
> CC_SRC_PATH="NA"
> # use go as the default language for chaincode
> CC_SRC_LANGUAGE="go"
> ```
> 
> I would remove those from that script, and then set the defaults to "NA" in the first lines of the deployCC script. You can then set the following if statements accordingly

Actually, playing around, removing them is the wrong thing to do. If you remove the defaults, then the variables do not exist. Rather, set them to NA in the network.sh script instead.

jcastrence:
> > I was playing around with the error text,a nd I realized that the defaults are still set in the main `network.sh` script:
> > ```
> > # chaincode name defaults to "basic"
> > CC_NAME="basic"
> > # chaincode path defaults to "NA"
> > CC_SRC_PATH="NA"
> > # use go as the default language for chaincode
> > CC_SRC_LANGUAGE="go"
> > ```
> > 
> > 
> > I would remove those from that script, and then set the defaults to "NA" in the first lines of the deployCC script. You can then set the following if statements accordingly
> 
> Actually, playing around, removing them is the wrong thing to do. If you remove the defaults, then the variables do not exist. Rather, set them to NA in the network.sh script instead.

Would we still need to check if the variables are empty then in the deployCC script, since our variables will now always be set to NA by default?

test-network/scripts/deployCC.sh
nikhil550: Sorry, meant if else. Was playing around and got engrossed. Try this:
```
#User has not provided a name
if [ "$CC_NAME" = "NA" ]; then
  fatalln "No chaincode name was provided. Valid call example: ./network.sh deployCC -ccn basic -ccp ../asset-transfer-basic/chaincode-go -ccl go"

# User has not provided a path
elif [ "$CC_SRC_PATH" = "NA" ]; then
  fatalln "No chaincode path was provided. Valid call example: ./network.sh deployCC -ccn basic -ccp ../asset-transfer-basic/chaincode-go -ccl go"

# User has not provided a language
elif [ "$CC_SRC_LANGUAGE" = "NA" ]; then
  fatalln "No chaincode language was provided. Valid call example: ./network.sh deployCC deployCC -ccn basic -ccp ../asset-transfer-basic/chaincode-go -ccl go"

## Make sure that the path to the chaincode exists
elif [ ! -d "$CC_SRC_PATH" ]; then
  fatalln "Path to chaincode does not exist. Please provide different path"
fi

# do some language specific preparation to the chaincode before packaging

CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
```
Moving the last line down does not shift "NA" to "na"

nikhil550:
> > > I was playing around with the error text,a nd I realized that the defaults are still set in the main `network.sh` script:
> > > ```
> > > # chaincode name defaults to "basic"
> > > CC_NAME="basic"
> > > # chaincode path defaults to "NA"
> > > CC_SRC_PATH="NA"
> > > # use go as the default language for chaincode
> > > CC_SRC_LANGUAGE="go"
> > > ```
> > > 
> > > 
> > > I would remove those from that script, and then set the defaults to "NA" in the first lines of the deployCC script. You can then set the following if statements accordingly
> > 
> > 
> > Actually, playing around, removing them is the wrong thing to do. If you remove the defaults, then the variables do not exist. Rather, set them to NA in the network.sh script instead.
> 
> Would we still need to check if the variables are empty then in the deployCC script, since our variables will now always be set to NA by default?

You can check for "NA". see comment above.

test-network/scripts/deployCC.sh
```diff
@@ -33,52 +33,24 @@ CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
...
-    CC_SRC_PATH="../asset-transfer-private-data"
-  elif [ "$CC_NAME" = "sbe" ]; then
-    println $'\e[0;32m'asset-transfer-sbe$'\e[0m' chaincode
-    CC_SRC_PATH="../asset-transfer-sbe"
-  else
-    fatalln "The chaincode name ${CC_NAME} is not supported by this script. Supported chaincode names are: basic, events, ledger, private, sbe, secured"
-  fi
+#User has not provided a name
+if [ -z "$CC_NAME" ] || [ "$CC_NAME" = "NA" ]; then
+  fatalln "No chaincode name was provided. Valid call example: ./network.sh deployCC -ccn \"basic\" -ccp \"../asset-transfer-basic/chaincode-go\" -ccl \"go\""
```
denyeart, line 38: Good.
But I think you can remove all the double quotes. It is only there with the variables as a good pactice, but shouldn't be needed when people make their own calls.

test-network/scripts/deployCC.sh
```diff
@@ -33,52 +33,24 @@ CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
...
-  elif [ "$CC_SRC_LANGUAGE" = "java" ]; then
-    CC_SRC_PATH="$CC_SRC_PATH/chaincode-java/"
-  elif [ "$CC_SRC_LANGUAGE" = "javascript" ]; then
-    CC_SRC_PATH="$CC_SRC_PATH/chaincode-javascript/"
-  elif [ "$CC_SRC_LANGUAGE" = "typescript" ]; then
-    CC_SRC_PATH="$CC_SRC_PATH/chaincode-typescript/"
-  fi
+# User has not provided a path
+if [ -z "$CC_SRC_PATH" ] || [ "$CC_SRC_PATH" = "NA" ]; then
+  fatalln "No chaincode path was provided. Valid call example: ./network.sh deployCC -ccn \"basic\" -ccp \"../asset-transfer-basic/chaincode-go\" -ccl \"go\""
```
denyeart, line 43: Same.

test-network/scripts/deployCC.sh
```diff
@@ -33,52 +33,24 @@ CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
...
-## Make sure that the path the chaincode exists if provided
-elif [ ! -d "$CC_SRC_PATH" ]; then
-  fatalln "Path to chaincode does not exist. Please provide different path"
+## Make sure that the path to the chaincode exists
+if [ ! -d "$CC_SRC_PATH" ]; then
+  fatalln "Path to chaincode does not exist. Please provide different path. ./network.sh deployCC -ccn \"basic\" -ccp \"../asset-transfer-basic/chaincode-go\" -ccl \"go\""
+fi
+
+# User has not provided a language
+if [ -z "$CC_SRC_LANGUAGE" ] || [ "$CC_SRC_LANGUAGE" = "NA" ]; then
```
denyeart, line 52: Line 32 has lowercased "NA" to "na", so this check doesn't work (I tried it). You'll need to check for "na" or better yet check for case insensitive "NA".

test-network/scripts/deployCC.sh
```diff
@@ -33,52 +33,24 @@ CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])
...
+  fatalln "No chaincode language was provided. Valid call example: ./network.sh deployCC -ccn \"basic\" -ccp \"../asset-transfer-basic/chaincode-go\" -ccl \"go\""
```
denyeart, line 53: Same.

denyeart (commented):
I grepped fabric-samples and fabric for occurrences of "deployCC". I see a few more occurrences where not all three flags are provided (ccn, ccp, ccl).

test-network/scripts/deployCC.sh
nikhil550: Another way would be to move line 32: `CC_SRC_LANGUAGE=$(echo "$CC_SRC_LANGUAGE" | tr [:upper:] [:lower:])` down and to the case conversion after this check.

test-network/scripts/deployCC.sh
```diff
@@ -29,58 +29,27 @@ println "- DELAY: ${C_GREEN}${DELAY}${C_RESET}"
...
-  fi
-## Make sure that the path the chaincode exists if provided
+# User has not provided a language
+elif [ -z "$CC_SRC_LANGUAGE" ] || [ "$CC_SRC_LANGUAGE" = "NA" ]; then
+  fatalln "No chaincode language was provided. Valid call example: ./network.sh deployCC -ccn basic -ccp ../asset-transfer-basic/chaincode-go -ccl go"
+
+## Make sure that the path to the chaincode exists
 elif [ ! -d "$CC_SRC_PATH" ]; then
-  fatalln "Path to chaincode does not exist. Please provide different path"
+  fatalln "Path to chaincode does not exist. Please provide different path. ./network.sh deployCC -ccn basic -ccp ../asset-transfer-basic/chaincode-go -ccl go"
```
nikhil550, line 48: You can probably remove the valid call example from this message. (If you keep it, add the text `Valid call example:`

nikhil550:
Looks good, just a couple nits. 

You will need to update the help text to state that go is no longer he default chaincode language:

` -ccl <language> - Programming language of the chaincode to deploy: go (default), java, javascript, typescript`

nikhil550 (approved):
LGTM. I will leave open in case anyone else has any comments

SamYuan1990:
Hi guys,

Please correct me for any mis understanding.
In this pr, we are removed short names and replaced by full name. as 

Short names are a crutch that can prevent someone from learning how to deploy their own chaincode. Short names have been replaced with their full path name. Short names are no longer recognized by shell scripts.

but ... why removed the default values?
is the default values prevent someone from learning or the short name prevents?

PR #393, opened 2020-12-16
Add go smart contract language to deploy commands

nikhil550:
With https://github.com/hyperledger/fabric-samples/pull/390 requiring the -ccl path, we need to add it to some samples that used the go default

Signed-off-by: Nikhil Gupta <nikhilgupta@macbook-air.lan>


lehors (approved):
LGTM

PR #394, opened 2020-12-17
Stop using deprecated outputAnchorPeersUpdate in test-network

wlahti:
#### Type of change

- Improvement

#### Description

First commit: 
- `configtxgen`'s `outputAnchorPeersUpdate` function has been deprecated for a long time now. This commit removes its usage in the test-network, using the config update flow instead to set the anchor peers for each organization. 

Second commit:
- Be more consistent in the output we generate in the test-network.

#### Related issues

[FAB-18381](https://jira.hyperledger.org/browse/FAB-18381)


wlahti:
I initially attempted a much bigger refactor of things to allow more reuse of code between the main test-network and the addOrg3 functions but I'm not particularly good at scripting and hit some issues in a few areas. This is good enough for now and moves us off the deprecated `outputAnchorPeersUpdate` function (and cleans up some of the output which was all over the place).

test-network/docker/docker-compose-test-net.yaml
```diff
@@ -123,3 +123,25 @@ services:
       - 9051:9051
     networks:
       - test
+  
+  cli:
```
denyeart, line 127: Why start the CLI fabric-tools container with the network?
We worked hard to remove the CLI container from test-network, since we want to show people how to use commands from their local environment rather than exec into a CLI container.

denyeart (commented):
Overall I like it and it works well!  Just one question on the CLI...

test-network/scripts/addAnchorPeer.sh
```diff
@@ -0,0 +1,64 @@
...
+#
+
+# import utils
+. scripts/envVar.sh
+. scripts/configUpdate.sh
+
+
+# NOTE: this must be run in a CLI container since it requires jq and configtxlator 
+createAnchorPeerUpdate() {
+  ORG=$1
```
nikhil550, line 15: Think you forgot a CHANNEL_NAME=$2 here

test-network/scripts/addAnchorPeer.sh
```diff
@@ -0,0 +1,64 @@
...
+  setGlobalsCLI $ORG
+
+  peer channel update -o orderer.example.com:7050 --ordererTLSHostnameOverride orderer.example.com -c $CHANNEL_NAME -f ${CORE_PEER_LOCALMSPID}anchors.tx --tls --cafile $ORDERER_CA >&log.txt
+  res=$?
+  cat log.txt
+  verifyResult $res "Anchor peer update failed"
+  successln "Anchor peers updated for org '$CORE_PEER_LOCALMSPID' on channel '$CHANNEL_NAME'"
+}
+
+ORG=$1
```
nikhil550, line 59: May be easier to consolidate these four lines into 

```
createAnchorPeerUpdate $1 $2

updateAnchorPeer $1 $2
```

rather than re-declare these variables so many times but your taste may vary.

nikhil550 (commented):
Overall the new process works well and is quite fast. The output cleanup is welcome and things look much better.

Bringing back the CLI container container is unwelcome, but the since the known process for updating the channel config requires `jq`, which is bundled with the container, not using the CLI container would create another dependency for getting started.

If we do keep the cli container, we may want to get rid of the org3 cli and only use one in the future.

wlahti:
Thanks for the reviews! Nik did a great job summarizing the conversation that a few of us had in the fabric documentation workgroup call last week. It comes down to what we want to do with the `jq` dependency and whether we keep a CLI container around to avoid that for end users or force them to install it locally as a prereq.

wlahti:
> If we do keep the cli container, we may want to get rid of the org3 cli and only use one in the future.

Note that I did remove the `org3cli` container and we reuse a single `cli` container now for all orgs.

test-network/scripts/addAnchorPeer.sh
wlahti: I don't have a strong preference here. I coded this both ways at various times but couldn't decide which I preferred or if one would be considered "best practice."

test-network/scripts/addAnchorPeer.sh
wlahti: Done

test-network/scripts/addAnchorPeer.sh
wlahti: Updated.

wlahti: Actually, I changed this up a bit. Consolidated down to setting the ORG and CHANNEL_NAME from $1 and $2 here, set the globals once for the script, and then avoid duplicating that code in each of the functions.

test-network/addOrg3/docker/docker-compose-org3.yaml
```diff
@@ -50,33 +50,3 @@ services:
       - 11051:11051
     networks:
       - test
-
```
nikhil550, line 53: This CLI container is used by the adding an org to the channel tutorial. As a result, I think it should be left in to not break the existing tutorial, even if it is not used by the script. We can update the tutorial to use the new CLI container, and then remove this redundant container afterwards

wlahti: I was planning to update the tutorial immediately afterwards, but I guess there's no reason I can't push a PR now even before this is merged.

wlahti: Actually, you're right, there are more moving updates to the doc than I realized. Reintroduced the Org3cli container temporarily until we're able to update the tutorials.

nikhil550: Tested locally, and ran through a bit of the Org3 tutorial and everything works. 

I think this is good enough to merge, and includes a lot of overdue clean up, once we have come to a decision on whether it is worth bringing back the Org3 container or not.

wlahti: Interestingly, the create channel tutorial actually already documents the correct way to set the anchor peers for each organization and avoids using a `fabric-tools` cli container. Maybe with that in mind we should just move away from the `fabric-tools` containers entirely and just require `jq` be installed as a prereq for the `test-network` at this point?

denyeart: I don't like the CLI container, but I also don't like dependencies for people just kicking the tires. My opinion is to keep CLI container so that people just kicking the tires with test-network, but not doing any config transactions, wouldn't require the jq dependency. It would only be required by people doing config transactions themselves, which is a whole different level of tire kicking which could have more dependencies. But let's get some other opinions as well.

wlahti: Discussed this with the crowd on today's fabric contributors meeting. Everyone present (that voiced an opinion) agreed that sticking with the CLI container for the automated scripts (`./network.sh createChannel` and `./addOrg3.sh up`) to set the anchor peers is ideal to keep the barrier to entry low for new users. For the tutorials that demonstrate running each command manually, we'll make sure all of those run on the local machine and avoid a CLI container. The original tutorial for creating a channel already does this while the adding an org to a channel tutorial still uses the org3cli container. I'll also add a note regarding the CLI container and why it's used in the scripts but not the manual steps.

nikhil550: Sounds like a good plan.

wlahti:
I'll put together a PR over in the `fabric` repo today to update the tutorials to reflect the recommended config update process for setting anchor peers. 

I think all of the review comments have been addressed at this point. Any last requests/suggestions or can we go ahead with a merge here?

nikhil550 (approved):
LGTM

PR #397, opened 2021-01-04
Explicitly Mount the Docker Unix Socket

lindluni:
Instead of mounting the entire `/var/run` directory explicitly mount the unix socket. Mounting the entire directory causes issues with future versions of Docker on some platforms.


nikhil550 (approved):
I saw this as well but Brett got here first. Thanks Brett!

PR #399, opened 2021-01-05
Fix error that occurs with go CC and javascript app

nikhil550:
Unlike the Javascript smart contract, the go version of the smart contract does not return any result when an asset is created. However, the application expects that a JSON of the asset being created is returned. As a result, the application quits with the following error:
```
******** FAILED to run the application: SyntaxError: Unexpected end of JSON input
```

This error is actually generated by the CI, but does not register as an error.

This PR updates that Javascript smart contract to bring the CreateAsset transaction into line with the other functions and smart contract languages and not return the value of the asset created.


lehors (dismissed):
I don't think this is the right way to fix this. The smart contract implementations should be aligned instead.

harrisob (changes requested):
We had a customer question at one time to see how to have results from the submit from the chaincode. Maybe we could we check for a result before doing the pretty print ?

nikhil550:
That would not be difficult. However, I wanted to check what we wanted to demonstrate before updating the smart contract

nikhil550:
@lehors @harrisob I checked the other smart contract languages and functions, and this was the only one that returned a value. As a result, I aligned the smart contract by not returning a the value from the JavaScript `CreateAsset` function, and by updating the javascript application to not expect a result.

denyeart:
Actually the Go application expects a result and prints it. It just happens to be nil since the Go chaincode doesn't return it:
https://github.com/hyperledger/fabric-samples/blob/master/asset-transfer-basic/application-go/assetTransfer.go#L81-L85

So I agree with Bret, for CreateAsset() I'd rather write the apps in a way that demonstrates results can be returned and handled, in this case printed. And check for a nil return in case the result is not actually returned by the chaincode (although we could update the chaincodes to return the JSON string similar to the node chaincode).

nikhil550:
@denyeart @harrisob In the interest of trying to get this bug fixed before the new releases are tagged, I updated the javascript application to check if the return is empty before printing the response. Now it runs for all languages. This is the only update.

harrisob (approved):

harrisob (approved):

PR #403, opened 2021-01-13
Update NodeJS Deps

lindluni:


nikhil550:
This sample is actually deprecated, and I wondered when it would be removed, since it is no longer pointed to by the docs. I assumed the 2.3 branch?

lindluni:
> This sample is actually deprecated, and I wondered when it would be removed, since it is no longer pointed to by the docs. I assumed the 2.3 branch?

I updated the title, originally I was just updating the fabcar deps, as I noticed they were out of date, but then I noticed everything was out of date, so I updated everything. But that is a good question...

Dave does plan to tag the release-2.2 and v2.2.x this afternoon, maybe thats a good time to remove it afterwards

nikhil550 (approved):

lindluni:
Also, just for the record should someone wonder why the package-lock files were removed: I did talk to the Node SDK maintainers and confirmed they would like to remove the package-lock files. We had some applications that had it, and some that did not. For now they said lets remove it (since in production you would correctly want to add the lock and properly shrink-wrap the application, which we don't do at all), and if they decide later we can add them all back.

PR #409, opened 2021-01-27
Rename --channel-id to --channelID in osadmin

jcastrence:
#### Type of change

#### Description

- Rename --channel-id flag in osadmin to --channelID to match the rest of fabric flags


lindluni:
Currently an outage at Google Cloud where our Artifactory instance is hosted, which is causing the failure here as the Maven repo is timing out trying to fetch the artifact, will retry later when the outage is resolved:

https://status.jfrog.io/

denyeart:
@jcastrence I figured out the test failures...
fabric-samples currently uses Fabric 2.3-stable images from artifactory.
But fabric master publishes as Fabric 2.4-stable currently (in prep for a 2.4.0 release).
Since the channelID flag change is only in 2.4-stable, the change in this PR still gets tested with the prior flag from fabric 2.3.
The fix is to move fabric-samples ci up to the latest 2.4-stable images, so that fabric-samples master gets tested against fabric master. This should be done regardless so that we always know that the latest of each works.

I'd recommend make the ci change in this PR itself so that we can verify the tests work.
Simply grep for "2.3" in the /ci directory and update all references to "2.4". You can see an example of what I mean in commit:
https://github.com/hyperledger/fabric-samples/commit/622cf5281459967b9263c6e83879381f9bbf8d22

lindluni:
Dave is correct, but you have an old base, this has been updated in master. You need to do: `git fetch upstream && git rebase upstream/master`

Then fix any conflicts and force push to update the branch

denyeart (approved):

PR #410, opened 2021-01-28
Fix crypto materials copy in commercial-paper network-starter

lebdron:
Fix usage of `*` which was treated as literal symbol instead of wildcard.


lehors (approved):
Yup, missed that when reviewing the last change.
Thanks.

PR #412, opened 2021-02-06
try to fix CI error

SamYuan1990:
Hi fabric-sample-maintainers,

I noticed that CI has error and I am try to fix it via this pr.


lindluni (commented):
This seems reasonable, not sure why CI all of a sudden is encountering this, but you can see in the log that the variable is not being set (even though its supposed to parse and set it for you by default in compose if you don't set it yourself) but it resolves the problem and has no side-effect as this variable is use for it's intended purpose

lindluni (changes requested):
Can we change the project name to `fabric` so it's obvious the network belongs to fabric. People have issues very frequently cleaning up things created by fabric-samples, the network artifacts left behind could be confused with something created by docker rather than fabric.

SamYuan1990:
> Can we change the project name to `fabric` so it's obvious the network belongs to fabric. People have issues very frequently cleaning up things created by fabric-samples, the network artifacts left behind could be confused with something created by docker rather than fabric.

I am not sure in which docker compose version changed this behavior.
it should be/used to be reading `.env` file with data configured below.
```
COMPOSE_PROJECT_NAME=net
IMAGE_TAG=latest
SYS_CHANNEL=system-channel
```
so that
```
      - CORE_VM_DOCKER_HOSTCONFIG_NETWORKMODE=${COMPOSE_PROJECT_NAME}_test
```
equals to 
```
networks:
  test:
```
as `net_test`, for peer starts from compose file in `net_test` docker network and keep chain code container in same network.
so I am upgrade code `.env` file to `fabric` too, it looks like some kind of redundancy... but keep shell logic and composer logic in same way.

lindluni:
Ah, nice catch on the `.env` file, that would explain why I couldn't find it using `grep` as I wasn't traversing hidden files. That makes it easier to find why the behaviour change as well. Thank you again for the contribution while we sort through figuring out the breaking change!

test-network/network.sh
```diff
@@ -479,6 +479,7 @@ else
   exit 1
 fi
```
nikhil550, line 481: Does COMPOSE_PROJECT_NAME=fabric need to be in both places? Could it be just in the env file?

Wouldn't this update change the name of the docker composer network? The docs refer to the `net_test` docker compose network in quite a few places?

https://hyperledger-fabric.readthedocs.io/en/latest/search.html?q=net_test&check_keywords=yes&area=default

nikhil550 (dismissed):

test-network/network.sh
SamYuan1990: let's see this CI log https://dev.azure.com/Hyperledger/Fabric-Samples/_build/results?buildId=29892&view=logs&j=98b5b0e0-fc73-50d8-cdbe-b7a1b46de137&t=e8322841-a67d-5ec3-ae89-37f2233483b3&l=513 with our code
https://github.com/hyperledger/fabric-samples/blob/master/test-network/.env#L1

ref here:
https://dev.azure.com/Hyperledger/TWGC/_build/results?buildId=29851&view=logs&j=079c46b8-677c-5016-d763-eb78926de79e&t=42fc6774-e221-5249-c003-7cb87b0a6fd3&l=886
the on release-1.4 branch, `net_byfn` is the network name and works.

wait a min....

SamYuan1990:
why release-1.4 works, but master doesn't?

SamYuan1990:
it's nearly mid night for me.
ref: https://docs.docker.com/compose/env-file/
it seems our .env file mis match with docker compose version?
@nikhil550, @lindluni I hope by remove the file to `docker` dir can make it works.

SamYuan1990:
otherwise, I will follow up and check reviewer's feedback tomorrow local time.

lindluni:
Whats even more interesting is...some of the tests work, some do not, so we need to unpack the difference there

lindluni:
> it's nearly mid night for me.
> ref: https://docs.docker.com/compose/env-file/
> it seems our .env file mis match with docker compose version?
> @nikhil550, @lindluni I hope by remove the file to `docker` dir can make it works.

There we go, that worked

denyeart (approved):

denyeart:
Thanks @SamYuan1990 !

PR #415, opened 2021-02-08
Make lint.sh Self-Detecting and Cleanup All Linting Issues

lindluni:
The current `lint.sh` script requires users to create a large matrix of information to lint everything and its possible to not check in your code to the linting matrix and thus not lint new code added. This change detects all chaincodes and applications automatically as long as they conform to our naming standard and lint them appropriately.

As part of that it cleans up all of the code that was incorrectly linted, or not linted at all.


commercial-paper/organization/digibank/application-java/pom.xml
```diff
@@ -14,7 +14,7 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
     <!-- fabric-chaincode-java -->
-    <fabric-chaincode-java.version>[2.0,3.0)</fabric-chaincode-java.version>
+    <fabric-chaincode-java.version>2.1.0</fabric-chaincode-java.version>
```
denyeart, line 17: I don't liking pinning to what will become an outdated java chaincode version, but if it is required, let's add a comment explaining the motivation for future generations.

denyeart (commented):
Just one nit.

commercial-paper/organization/digibank/application-java/pom.xml
lindluni: Something actually broke in the dependency chain. You actually cant compile this example anymore, using any version other than `2.1.0` has dependency conflicts, several people have noted this in RocketChat this past week. I need to open a Jira to have the Java team investigate this

denyeart: Ok, let's go ahead and merge. FYI @lesleyannjordan @bestbeforetoday this could use some investigation.

denyeart (approved):

denyeart (approved):

PR #416, opened 2021-02-10
Fix unread .env file

naonishijima:
Deploying chaincode failed due to missing .env file.
The previous commit moves .env file to docker directory from
test-network directory. Therefore, .env file doesn't exist
current working directory.
This PR adds --env-file option to docker-compose command to
read .env file.


lindluni:
While this change is valid and what should be done, instead I superseded it with #417 and removed the .env file altogether and expliclty just named the network.

naonishijima:
I got it. Thanks.

PR #418, opened 2021-02-11
Add Service Labels and Cleanup Using Standard Tools  and Remove Unused Vars

lindluni:
Add a service label to all containers launched by compose. This allows us to filter on the label for removing running containers.
    
Use filtering for querying running containers and images to make sure we only target the containers and images we want to remove.

Removes the `OS_ARCH` var as it was unused.

### DIFF
--- a/test-network/network.sh
+++ b/test-network/network.sh
@@ -19,27 +19,19 @@ export VERBOSE=false
 . scripts/utils.sh
 
 # Obtain CONTAINER_IDS and remove them
-# TODO Might want to make this optional - could clear other containers
 # This function is called when you bring a network down
 function clearContainers() {
-  CONTAINER_IDS=$(docker ps -a | awk '($2 ~ /dev-peer.*/) {print $1}')
-  if [ -z "$CONTAINER_IDS" -o "$CONTAINER_IDS" == " " ]; then
-    infoln "No containers available for deletion"
-  else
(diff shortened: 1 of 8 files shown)


test-network/network.sh
```diff
@@ -19,27 +19,19 @@ export VERBOSE=false
...
 # This function is called when you bring the network down
 function removeUnwantedImages() {
-  DOCKER_IMAGE_IDS=$(docker images | awk '($1 ~ /dev-peer.*/) {print $3}')
-  if [ -z "$DOCKER_IMAGE_IDS" -o "$DOCKER_IMAGE_IDS" == " " ]; then
-    infoln "No images available for deletion"
-  else
-    docker rmi -f $DOCKER_IMAGE_IDS
-  fi
+  infoln "Removing generated chaincode docker images"
+  docker image rm -f $(docker images -aq --filter reference='dev-peer*') 2>/dev/null || true
```
denyeart, line 34: this should be `rmi` rather than `rm` right?

denyeart (approved):

test-network/network.sh
lindluni: Just to close this for whoever might visit this in the future, `--filter` is not available on `rmi` which is why I used `image rm`

lehors:
That's a welcome change! Thank you.

PR #419, opened 2021-02-11
Remove Image Tag Override

lindluni:
The `-i` flag was originally added to support an upgrade sample. Since that sample is no longer available remove the `-i` flag to clean up the network.sh options and avoid confusion as it's possible now to specify an image version that is no longer backwards compatible with the new test-network with which uses `osnadmin`.

Signed-off-by: Brett Logan <lindluni@github.com>


denyeart (commented):
I grepped for "IMAGETAG" and found a few more that I think can be removed.

lindluni:
> I grepped for "IMAGETAG" and found a few more that I think can be removed.

Fixed, was grepping for `IMAGE_TAG` but we renamed it to `IMAGETAG` in some places...

denyeart (approved):

PR #426, opened 2021-02-24
Add dutch auction sample with auditor

nikhil550:
This PR uses the dutch auction sample to provide an example of how an auditor organization can be added to a smart contract. The protobuf signature policy syntax is used to assign a complex signature policy to an asset using state based endorsement, which is something that users have been asking for help on. This allows the auditor organization to "overrule" the normal endorsement policy if needed.

The simple blind auction is moved into a separate folder inside `auction`


lindluni:
I know it's still in draft, but I wanted to make a comment so you could address it while still iterating on this:

In Go, variables starting with Capital letters have a specific semantic meaning, i.e., are exported non-private variables. There are lots of places where you used capital letters for variable names unintentionally, these should be lowercased to provide proper meaning (and intention).

Beyond that, this looks really interesting, I've never seen a dutch auction before so enjoying going through this. Thanks Nik

lindluni (changes requested):
.

nikhil550:
> I know it's still in draft, but I wanted to make a comment so you could address it while still iterating on this:
> 
> In Go, variables starting with Capital letters have a specific semantic meaning, i.e., are exported non-private variables. There are lots of places where you used capital letters for variable names unintentionally, these should be lowercased to provide proper meaning (and intention).
> 
> Beyond that, this looks really interesting, I've never seen a dutch auction before so enjoying going through this. Thanks Nik

Thank you for pointing that out. I have addressed that and fixed the variables cases in all sets of smart contracts.

lehors (changes requested):
Apologies for not getting to this sooner. I like the auction and everything worked fine when I tried it. However, I don't think we should add another directory for this. I'd rather we simply have both auction samples at the top, which is consistent with what we have done for tokens. For that matter, I suggest using the names "auction-simple" and "auction-dutch" to keep them together.
Thanks.

nikhil550:
> Apologies for not getting to this sooner. I like the auction and everything worked fine when I tried it. However, I don't think we should add another directory for this. I'd rather we simply have both auction samples at the top, which is consistent with what we have done for tokens. For that matter, I suggest using the names "auction-simple" and "auction-dutch" to keep them together.
> Thanks.

Good suggestion. I moved the samples into two different directories.

lehors (approved):
LGTM. Thank you!

nikhil550:
> LGTM. Thank you!

@lehors Looks like I need a re-review after a rebase

lehors (approved):
LGTM. Thanks for your determination and patience on this one!

PR #435, opened 2021-03-30
abstore contract - Aval, Bval bad validation if values are not a number

mirooon:
Standard validation for numbers does not work.

`if (typeof parseInt(Aval) !== 'number' || typeof parseInt(Bval) !== 'number') {
      return shim.error('Expecting integer value for asset holding');
    }`

Good:

`if (isNaN(parseInt(Aval)) || isNaN(parseInt(Bval))) {
      return shim.error('Expecting integer value for asset holding');
    }`

Otherwise contract accepts no number values as below:
image

### DIFF
--- a/chaincode/abstore/javascript/abstore.js
+++ b/chaincode/abstore/javascript/abstore.js
@@ -25,7 +25,7 @@ var ABstore = class {
     let Aval = args[1];
     let Bval = args[3];
 
-    if (typeof parseInt(Aval) !== 'number' || typeof parseInt(Bval) !== 'number') {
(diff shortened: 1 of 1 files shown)


nikhil550:
You need to sign off your commit in order to contribute.
You can do that using `git commit --amend -s` if you have configured your github email.

nikhil550:
Still no singoff. Maybe you did not push. You also need to squash all of the commits in this PR into one commit, including the merge. It may be easier to close this and create a clean PR with the same updates.

PR #434, opened 2021-03-26
Fixed information leak

kurtseifried:
Fixed leak as per request https://hackerone.com/bugs?report_id=1124005&subject=user

PR #440, opened 2021-04-19
Fixed information leak

ryjones:
Replaces PR #434 

Signed-off-by: Kurt Seifried <kurt@seifried.org>
Signed-off-by: Ry Jones <ry@linux.com>

### DIFF
--- a/commercial-paper/organization/digibank/digibank.sh
+++ b/commercial-paper/organization/digibank/digibank.sh
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
+TMPFILE=`mktemp`
 shopt -s extglob
 
 function _exit(){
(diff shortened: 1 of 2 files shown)


ryjones:
@lehors ?

lehors (approved):
Thanks for the contribution although I wouldn't say that just obscuring the name of the file avoids leaking info per se. I agree it's a step in the right direction though so I'll approve but I would suggest you also trap any signal to delete the temp file in case of interruption.

kurtseifried:
The tmp file has safe ownership defaults so deletion is not a problem, anyone that can read the file can by definition read the file while it exists. Trapping signals to delete it if interrupted prevents the file from hanging which is nice, but 1) tmp gets cleaned (e.g. system reboot) and 2) you can't reliably trap and catch all this in the script and 3) it still wouldn't solve the problem, that anyone that can read the file, can read it while it exists.

kurtseifried:
FYI this is CVE-2021-1000010

PR #444, opened 2021-04-30
Disable Go pprof profiling in test network

denyeart:
Peers and Orderers by default disable Go pprof profiling, however test
network was enabling pprof in peers by default. As a good practice,
pprof should be disabled by default and only enabled when needed
for debug profiling.


lehors (approved):
LGTM!

PR #447, opened 2021-06-04
Support docker.sock in rootless mode

naonishijima:
Docker version 20.10 support rootless mode. That features
changes a mount path of docker.sock.
This PR loads a mount path from DOCKER_HOST environment,
and if not set, a mount path will be /var/run/docker.sock

FAB-18481


lehors (approved):
LGTM. Thanks!

PR #448, opened 2021-06-05
change docker-compose version from 2 to 2.4  on folder addOrg3/docker

s4l10u:
Following the instructions to set an organization 3, the docker-compose file (test-network/addOrg3/docker/docker-compose-org3.yaml) was invalid because it didn't accept additional properties in the network declaration (test_network) with docker-compose version 2. I was able to solve the problem by setting the version to 2.4.

### DIFF
--- a/test-network/addOrg3/docker/docker-compose-ca-org3.yaml
+++ b/test-network/addOrg3/docker/docker-compose-ca-org3.yaml
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-version: '2'
+version: '2.4'
 
 networks:
   test:
--- a/test-network/addOrg3/docker/docker-compose-couch-org3.yaml
+++ b/test-network/addOrg3/docker/docker-compose-couch-org3.yaml
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-version: '2'
+version: '2.4'
 
 networks:
   test:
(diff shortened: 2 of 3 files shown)


indranil32:
@s4l10u you have to do this for all the compose files under addOrg3/docker.
$ ls -ltr addOrg3/docker/
total 9
-rw-r--r-- 1 XXXX 1049089  666 Jun  4 19:30 docker-compose-ca-org3.yaml
-rw-r--r-- 1 XXXX 1049089 1919 Jun  4 19:30 docker-compose-org3.yaml
-rw-r--r-- 1 XXXX 1049089 1386 Jun  9 15:04 docker-compose-couch-org3.yaml

`The Compose file '.\addOrg3/docker/docker-compose-couch-org3.yaml' is invalid because:
networks.test value Additional properties are not allowed ('name' was unexpected)
`

Also please fix the commit signoff for checks to turn green.

lindluni:
Presumably you were seeing this because you were on an old version of fabric-samples (maybe v2.2 which our current LTS). This was fixed in main already as we no longer use `test_network` parsed from .dotfiles, instead we hardcode the network to `fabric_test`

s4l10u:
Hi I made my last clone from the  2.3 release and I had the same issue.

indranil32:
$ ./addOrg3.sh up -ca
Adding org3 to channel 'mychannel' with '10' seconds and CLI delay of '3' seconds and using database 'leveldb'

Generating certificates using Fabric CA
**Version "2.4.4.4.4"** in ".\docker/docker-compose-ca-org3.yaml" **is invalid**.
Creating Org3 Identities
Enrolling the CA admin

lehors (approved):
LGTM. Thanks!

PR #451, opened 2021-06-16
Add process.exit(0) to sample applications

denyeart:
Add process.exit(0) so that sample javascript and typescript
applications exit with success upon completion instead of hanging.


nikhil550 (approved):
LGTM

denyeart:
While this fix makes CI pass, after talking to others, it sounds like process.exit(0) should not be needed and the hang has some underlying root cause that needs to be investigated rather than worked around.

mbwhite:
The SDK has now been updated to include fixes from the grpc-js library that should resolve the issue

PR #453, opened 2021-06-29
expose operations endpoint

varadgit:

PR #454, opened 2021-06-29
expose operations endpoint

varadgit:


nikhil550:
Is this a dup of https://github.com/hyperledger/fabric-samples/pull/453

varadgit:
@nikhil550 this is against 2.2 branch.. (just fixed the target!)

nikhil550 (dismissed):
LGTM. no sure about the pipeline problems

varadgit:
@nikhil550 can you trigger the pipeline again to see if the the last set of changes resolve it?

varadgit:
@nikhil550 one other question? Does this need to change to `test-network`?
https://github.com/hyperledger/fabric/blob/main/scripts/bootstrap.sh#L52

nikhil550:
> @nikhil550 one other question? Does this need to change to `test-network`?
> https://github.com/hyperledger/fabric/blob/main/scripts/bootstrap.sh#L52

That does need to be changed. Good catch. I will put in the PR.

test-network/docker/docker-compose-test-net.yaml
```diff
@@ -57,7 +59,7 @@ services:
       # the following setting starts chaincode containers on the same
       # bridge network as the peers
       # https://docs.docker.com/compose/networking/
-      - CORE_VM_DOCKER_HOSTCONFIG_NETWORKMODE=${COMPOSE_PROJECT_NAME}_test
+      - CORE_VM_DOCKER_HOSTCONFIG_NETWORKMODE=docker_test
```
nikhil550, line 62: I think the network should be called net_test. I ran the sample locally and got the following error
```
  peer=peer0.org1.example.com:7051, status=500, message=error in simulation: failed to execute transaction 2471796d567ee3d268c83e8c2bd889a8c30fbb7294a33726592db3c5b6fdf996: could not launch chaincode basic_1.0:4ec191e793b27e953ff2ede5a8bcc63152cecb1e4c3f301a26e22692c61967ad: error starting container: error starting container: API error (404): network docker_test not found
```

nikhil550 (approved):
Got the pipeline to run but found an error when I ran locally

test-network/docker/docker-compose-test-net.yaml
varadgit: In main it is called `fabric`
https://github.com/hyperledger/fabric-samples/blob/main/test-network/docker/docker-compose-test-net.yaml#L71

I left it as `docker` since it defaults the network to the folder name it ran from... I just made another commit that hopefully helps keep it consistent...

nikhil550: Main uses networks:
```
  test:
    name: fabric_test
```

and then 
```
    networks:
      - test
```

Maybe we should use those just to keep the samples and docs consistent between versions.

varadgit: sounds good! PR updated...

mbwhite (approved):

PR #457, opened 2021-07-02
Fix off_chain_data to put data to off-chain database

yukiknd:
The off_chain_data sample fails to put data to the off-chain
database. The application does not read fetched blocks because
it uses the old interface of `addBlockListener()` to handle
block events.

This PR fixes the application to use the latest block listener
and build the off-chain database.

Signed-off-by: Yuki Kondo <yuki.kondo.ob@hitachi.com>


nikhil550 (approved):
Ran through, and this looks good to me. Thank you for putting in this fix.

PR #460, opened 2021-07-16
Update CreateAsset txn in basic js and ts sample

jt-nti:
The Golang and Java sample chaincode returned an error when trying to create an asset which already exists

JavaScript and TypeScript samples should now do the same


nikhil550 (approved):
LGTM

PR #462, opened 2021-07-25
Genesis Block error, fixed.

ksanjaykumar1:
Genesis Block error due  to  missing profile in configtx.yaml file.
When you run the present program, it creates a genesis block which doesn't  look right, plus I couldn't find where the genesis block was being created in the present network.sh file in test-network folder.


nikhil550:
I do not think that this PR is necessary. If you use the install scripts: https://hyperledger-fabric.readthedocs.io/en/latest/install.html these files should be installed on your machine.

lehors:
I agree with @nikhil550 so I'm closing this. If you believe we're missing something please, reopen and explain.
Thanks.

PR #465, opened 2021-07-26
changed all 'complileOnly' to 'implementation' on all fabric-samples for the fabric dependency

fraVlaca:
spotted that on the dependencies of all chaincode-java we were doing 'compileOnly 'org.hyperledger.fabric-chaincode-java:fabric-chaincode-shim:2.+'' instead of 'implementation'
This was preventing all java smart contracts to run on VSCode: microfab (used on vscode) does not import fabric while in cli (which uses test-network) it is embeded


lehors:
This is no longer relevant, PR #481 took care of it.

PR #467, opened 2021-08-12
Fixed state deserialize which is an uint enum in contract definition.

lucribas:
Signed-off-by: Luciano da Silva Ribas <luciano.ribas+git@gmail.com>


lehors:
Sorry but our CI was broken. It is now fixed, please, rebase to pick up the fix. The checks should then be able to pass. Thanks.

mbwhite (approved):

PR #468, opened 2021-08-18
Fix: Cannot launch chaincode in nano bash in Linux

naonishijima:
CHAINCODEADDRESS is set to host.docker.internal which is
only available in MAC and Windows environments.
For that reason, it cannot launch chaincode on Linux.
This PR sets CHAINCODEADDRESS to 127.0.0.1 instead of
host.docker.internal.

### DIFF
--- a/test-network-nano-bash/peer1.sh
+++ b/test-network-nano-bash/peer1.sh
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+if [ "$(uname)" == "Linux" ] ; then
+  CCADDR="127.0.0.1"
+else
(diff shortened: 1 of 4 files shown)


mbwhite:
@naonishijima yes I agree that this needs to be changed to work with Linux.

Though I *think* that this then stops it working in Mac.. @denyeart ?

denyeart:
Right, I couldn't get loopback to work on Mac, and I believe host.docker.internal only works on Mac. So I guess we'll need to add a OS check to the script and set the environment variable accordingly.

Thank you @naonishijima for finding this, do you want to attempt the OS check?

PR #470, opened 2021-08-19
Fix: logspec output

naonishijima:
If ORDERER_KAFKA_VERBOSE is true, the output of /logspec will contain
the following kafka information even if it is raft orderer.
{"spec":"orderer.consensus.kafka.sarama=debug:info"}
This PR deletes environment valiable for kafka in docker-compose.


nikhil550 (approved):
LGTM. I will leave open in case there is a reason we left the kafka varables in.

PR #478, opened 2021-08-30
Update first-network to explicity mount docker unit socket

nikhil550:
Explicitly mount docker.sock on the old first network sample. The sample is deprecated. However, this update will allow the sample to run on MacOS and allow me and others to run migration tests.
'

Signed-off-by: Nikhil Gupta <ngupta@triplepointliquidity.com>


mbwhite:
So the problem here is with the wrong version of Java been pulled in... will get a separate PR to fix this..

mbwhite:
PR that needs to go in first.. https://github.com/hyperledger/fabric-samples/pull/483

PR #479, opened 2021-08-30
Rm docker files mount for CLI container

nikhil550:
The CLI container mounts `- /var/run/:/host/var/run/`. Not sure the reason for this, and I assume that it was left over from the peer container. I removed and tested and test network ran fine.

Signed-off-by: Nikhil Gupta <ngupta@triplepointliquidity.com>


mbwhite:
@nikhil550 when your ran the network, did it start chaincode without any problems.

Not certain but I think that the /var/run is to let the peer access the docker daemon

nikhil550:
I re-tested, deployed chaincode and submitted transactions. The peer needs the access to the deamon to launch chaincode, but I do not think that the CLI container needs access, unless I am mistaken.

lehors (approved):
Works for me as well so, LGTM.

PR #487, opened 2021-09-07
Add in a setEnvOrg script

mbwhite:
Helper script to set environment variables, so make it eaiser to use the peer command
from the shell command line


test-network/README.md
```diff
@@ -3,3 +3,22 @@
 You can use the `./network.sh` script to stand up a simple Fabric test network. The test network has two peer organizations with one peer each and a single node raft ordering service. You can also use the `./network.sh` script to create channels and deploy chaincode. For more information, see [Using the Fabric test network](https://hyperledger-fabric.readthedocs.io/en/latest/test_network.html). The test network is being introduced in Fabric v2.0 as the long term replacement for the `first-network` sample.
 
 Before you can deploy the test network, you need to follow the instructions to [Install the Samples, Binaries and Docker Images](https://hyperledger-fabric.readthedocs.io/en/latest/install.html) in the Hyperledger Fabric documentation.
+
+## Using the Peer commands
```
denyeart, line 7: I think this is a good enhancement.
However, we previously decided to put all test network doc in the Fabric ReadTheDocs tutorial, rather than having instructions spanning both the README and tutorial, so that we can have a single place for readers and for our own maintenance sanity. Do you want to move the README content to the test network docs tutorial?

test-network/setOrgEnv.sh
```diff
@@ -0,0 +1,59 @@
...
+elif [[ ${ORG,,} == "org2" || ${ORG,,} == "magnetocorp" ]]; then
+
+   CORE_PEER_LOCALMSPID=Org2MSP
+   CORE_PEER_MSPCONFIGPATH=${DIR}/test-network/organizations/peerOrganizations/org2.example.com/users/Admin@org2.example.com/msp
+   CORE_PEER_ADDRESS=localhost:9051
+   CORE_PEER_TLS_ROOTCERT_FILE=${DIR}/test-network/organizations/peerOrganizations/org2.example.com/peers/peer0.org2.example.com/tls/ca.crt
+
+else
+   echo "Unknown \"$ORG\", please choose Org1/Digibank or Org2/Magnetocorp"
+   echo "For example to get the environment variables to set upa Org2 shell environment run:  ./setOrgEnv.sh Org2"
```
lehors, line 40:
```suggestion
   echo "For example to get the environment variables to set up a Org2 shell environment run:  ./setOrgEnv.sh Org2"
```

lehors (approved):
I quite like it too but note that there is quite a bit of overlap with scripts/envVar.sh. Ideally we would reduce that overlap. This could be done in another PR though.

mbwhite:
@lehors thanks - from a functional point of view yes... envvar.sh is very similar.. .though envvars.sh I've never been able to properly run  that to set up a local shell;  primary reasons I think is that if you're NOT running bash yourself (I'm zsh) then just sourcing the script has gotchas.

test-network/README.md
mbwhite: @denyeart I'll put in a separate PR for the fabric docs and update this  to refer back. when it's avilable.

PR #493, opened 2021-09-20
System Channel missing

ksanjaykumar1:
Added missing function to generate genesis block. As far I understand , Orderer use genesis.block to start the network  but in the code genesis.block is not being generated. 

Further more when I clone this code and try to create a application channel it is falling.

### DIFF
--- a/test-network/configtx/configtx.yaml
+++ b/test-network/configtx/configtx.yaml
@@ -300,6 +300,20 @@ Channel: &ChannelDefaults
 ################################################################################
 Profiles:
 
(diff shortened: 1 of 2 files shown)


denyeart:
Thanks for the contribution but in main branch system channel is not used anymore and therefore the system channel genesis block has been removed. You can see the switch away from system channel in this commit:
https://github.com/hyperledger/fabric-samples/commit/aa7287e1e38d99cf305b53443b8239b82e1371f9

If you want to use test-network with system channel, please reference release-2.2 branch which still uses system channel.

PR #494, opened 2021-09-22
Basic assert transfer typescript app migration to gateway

sapthasurendran:


asset-transfer-basic/application-typescript/src/app.ts
```diff
@@ -3,28 +3,66 @@
...
+    'ca.crt',
+);
+const peerEndpoint = 'localhost:7051';
 
 // pre-requisites:
-// - fabric-sample two organization test-network setup with two peers, ordering service,
-//   and 2 certificate authorities
+// - fabric-sample two organization test-network setup with two peers and ordering service
 //         ===> from directory /fabric-samples/test-network
-//         ./network.sh up createChannel -ca
```
denyeart, line 21: Why remove the CA option and related register/enroll utilities?
Won't we still want to demonstrate how to use fabric-ca-client to interact with CAs?

asset-transfer-basic/application-typescript/package.json
```diff
@@ -20,11 +20,13 @@
     "author": "Hyperledger",
     "license": "Apache-2.0",
     "dependencies": {
-        "fabric-ca-client": "^2.2.4",
-        "fabric-network": "^2.2.4"
+        "fabric-gateway": "0.1.0-dev.20210921.3"
```
denyeart, line 23: Can we use a latest or unstable instead of pinning to a point in time version?

denyeart (commented):
I thought we talked about ADDING gateway application SDK samples rather than REPLACING the legacy SDK samples right away.
I was thinking we'd add a folder like /application-gateway-typescript.
That being said, we'll still have the release-2.2 branch with the legacy application samples if we prefer to replace right away. Thoughts?

denyeart:
Typically we ask contributors to squash commits when submitting PRs. Many people find it easier to amend commits rather than creating/squashing commits. Either way is fine, but ultimately we'll want the change to show up as a single commit in the branch history.
See suggestions here: https://hyperledger-fabric.readthedocs.io/en/latest/github/github.html#updating-a-pull-request

asset-transfer-basic/application-typescript/src/app.ts
```diff
@@ -108,64 +105,130 @@ async function main() {
...
+            );
             console.log('*** Result: committed');
 
-            console.log('\n--> Evaluate Transaction: ReadAsset, function returns an asset with a given assetID');
+            console.log(
+                '\n--> Evaluate Transaction: ReadAsset, function returns an asset with a given assetID',
+            );
             result = await contract.evaluateTransaction('ReadAsset', 'asset13');
-            console.log(`*** Result: ${prettyJSONString(result.toString())}`);
+            console.log(`*** Result: ${Buffer.from(result).toString()}`);
```
davidkel, line 141: For the result, you probably want to use something like
```
Buffer.from(result, 'hex').toString()
```
ie add the hex parameter otherwise the output won't be easily understood

asset-transfer-basic/application-typescript/package.json
bestbeforetoday: Ideally we would use **latest** now that (as of yesterday) **latest** does actually point to the most recent build rather than one from 6 months ago! However... until we reach a first release version, this may cause breakages if any API changes are made in the development stream, so we might be safer picking a specific build version for now, _as long as_we remember to update again (and change to use **latest** tag) when we get to release time._

bestbeforetoday:
> I thought we talked about ADDING gateway application SDK samples rather than REPLACING the legacy SDK samples right away.
> I was thinking we'd add a folder like /application-gateway-typescript.
> That being said, we'll still have the release-2.2 branch with the legacy application samples if we prefer to replace right away. Thoughts?

I don't know (or don't remember) what discussion was had prior to this work on replacing vs. augmenting the samples, but... I suspect users just visit the default (main) branch and expect to see usable samples for any current version of Fabric. So for now I suspect we are better augmenting (adding new) samples rather than replacing existing ones that are still useful.

Beyond this PR, maybe we should think about providing better pointers in the README to specific samples for specific Fabric versions and client APIs. They could point to other branches if that makes sense.

Some of the samples in this repository are definitely past their best though, and probably cause more harm than help. Perhaps for the future we could remove old samples that have been superseded by more up-to-date ones?

asset-transfer-basic/application-typescript/src/app.ts
bestbeforetoday: I agree that demonstrating interaction with the CA is useful. I'm not sure if the mechanism we should be promoting is the SDKs or if it would be better to promote the use of the CLI. The Fabric Gateway API that this PR is demonstrating does not have any API for interacting with the CA.

One thing I really dislike about the current samples is how the user identity is enrolled with the CA as part of the same code (and often the same function) as the transactional logic. This is absolutely not how real applications should be written and actively promotes bad practices. Countless times I have seen application code that does enrolment for client identities every time the application is run. In other words, create new client credentials for every invocation of the application. Identities should be created as a separate administrative step (and probably by an administrator) prior running client applications that make use of those client identities.

asset-transfer-basic/application-typescript/src/app.ts
bestbeforetoday: Is the JSON returned by the chaincode really hex encoded? I thought it would be UTF-8 encoded to byte data.

asset-transfer-basic/application-typescript/src/app.ts
denyeart: ok, i'm good with removing CA register/enrollment from most of the samples, as long as we have at least one sample that focuses on register/enroll with fabric-ca-client.

denyeart:
> > I thought we talked about ADDING gateway application SDK samples rather than REPLACING the legacy SDK samples right away.
> > I was thinking we'd add a folder like /application-gateway-typescript.
> > That being said, we'll still have the release-2.2 branch with the legacy application samples if we prefer to replace right away. Thoughts?
> 
> I don't know (or don't remember) what discussion was had prior to this work on replacing vs. augmenting the samples, but... I suspect users just visit the default (main) branch and expect to see usable samples for any current version of Fabric. So for now I suspect we are better augmenting (adding new) samples rather than replacing existing ones that are still useful.
> 
> Beyond this PR, maybe we should think about providing better pointers in the README to specific samples for specific Fabric versions and client APIs. They could point to other branches if that makes sense.
> 
> Some of the samples in this repository are definitely past their best though, and probably cause more harm than help. Perhaps for the future we could remove old samples that have been superseded by more up-to-date ones?

Ok @sapthasurendran , please re-work this PR to ADD a new gateway app sample instead of REPLACE the existing sample.

@bestbeforetoday we actually did refresh and standardize on asset transfer samples last year. We said we would go with a bit of overlap time and then remove most of the samples under /chaincode folder, which I think we can do now.  The biggest question is what to do about Commercial Paper. We can have that discussion as part of the Gateway doc discussion since it ties so heavily to the Developing Applications doc.

asset-transfer-basic/application-typescript/src/app.ts
sapthasurendran: @davidkel I don't think this operation is possible on unit8array. Also output looks good with default  UTF-8 encoding .

sapthasurendran:
Thanks for the inputs. Updated the PR with the changes suggested.

asset-transfer-basic/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,235 @@
...
+                arguments: ['asset1', 'blue', '5', 'Tomoko', '400'],
+            });
+
+            console.log('*** Waiting for transaction commit');
+
+            const successful = await commit.isSuccessful();
+            if (!successful) {
+                const status = await commit.getStatus();
+                throw new Error(`Transaction ${commit.getTransactionId()} failed to commit with status code: ${status}`);
+            }
```
bestbeforetoday, line 176: This flow has just changed. Sorry! It would be worth updating to the latest published version of _fabric-gateway_ and reworking to use the current flow, which is something like:
```typescript
const status = await commit.getStatus();
if (!status.successful) {
    throw new Error(`Transaction ${status.transactionId} failed to commit with status code ${status.code}`);
}
```

bestbeforetoday (changes requested):
A few things I think are worth addressing in comments.

Also... I wonder if the `main()` is reducing readability by doing too much. It seems to contain multiple examples of each invocation type, and multiple invocations of many chaincode functions:

1. submitTransaction('InitLedger')
2. evaluateTransaction('GetAllAssets')
3. submitTransaction('CreateAsset')
4. evaluateTransaction('ReadAsset')
5. evaluateTransaction('AssetExists')
6. submitTransaction('UpdateAsset')
7. submitAsync('UpdateAsset')
8. evaluateTransaction('ReadAsset')
9. Failing submitTransaction('UpdateAsset')

Could this be reduced to a smaller set of interactions so it still demonstrates all the transaction functions and invocation styles, but there is less code for people to get their heads around?

asset-transfer-basic/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,246 @@
...
+        await createAsset(contract);
+
+        //Update an existing asset asynchronously.
+        await updateAssetAsync(contract);
+
+        //Get the asset details by assetID
+        await readAssetByID(contract);
+
+        //Update an asset which does not exist.
+        await updateAsset(contract)
```
bestbeforetoday, line 124: It's a really small niggle but, for clarity, the function name could either reflect that it's going to fail, or that it's updating a non-existent asset

asset-transfer-basic/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,246 @@
...
+}
+
+async function updateAssetAsync(contract:Contract):Promise<void> {
+
+    // Submit transaction asynchronously, blocking until the transaction has been sent to the orderer, and allowing
+    // this thread to process the chaincode response (e.g. update a UI) without waiting for the commit notification
+
+    const commit = await contract.submitAsync('UpdateAsset', {
+        arguments: ['asset1', 'blue', '5', 'Tomoko', '400'],
+    });
```
bestbeforetoday, line 201: Does this (or any other) transaction that we submit actually return a value? If so, it might be good to print the return value (`commit.getResult()`) immediately after the `submitAsync()` call to demonstrate that we can do work with the result after successfully sending to the orderer but before we wait for the commit

bestbeforetoday (commented):
Generally looks good to me. A couple of comments where it might be possible to improve. Otherwise needs a review from a maintainer like @mbwhite

asset-transfer-basic/application-gateway-typescript/src/app.ts
sapthasurendran: I don't see any submit  transaction returning value in the chaincode .

bestbeforetoday (dismissed):
Looks OK to me but will need a maintainer review

PR #495, opened 2021-09-23
Pipeline changes for Linting

sapthasurendran:
- Removed global install of modules. 
- 
- Project can now chose it’s own linting framework, versions and execution pattern
- 
- Fixed linting errors for chaincode-typescript

### DIFF
--- a/ci/scripts/lint.sh
+++ b/ci/scripts/lint.sh
@@ -29,13 +29,11 @@ for dir in $dirs; do
         print "The following files contain import errors, please run 'goimports -l -w <path>' to fix these issues:"
         echo "${output}"
       fi
-    elif [[ "$dir" =~ "-javascript" ]]; then
-      print "Running ESLint"
-      if [[ "$dir" =~ "chaincode" ]]; then
-        eslint *.js */**.js
-      else
-        eslint *.js
-      fi
+    elif [[ "$dir" =~ "-javascript" || "$dir" =~ "-typescript" ]]; then
+      print "Installing node modules"
+      npm install
+      print "Running Lint"
+      npm run lint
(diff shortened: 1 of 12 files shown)


denyeart:
It's good that  asset-transfer-basic chaincode and application javascript has been updated. But should we update all the javascript assets in fabric-samples to be consistent?

Let's have a javascript expert such as @bestbeforetoday or @mbwhite comment on the actual changes.

sapthasurendran:
> It's good that asset-transfer-basic chaincode and application javascript has been updated. But should we update all the javascript assets in fabric-samples to be consistent? Based on the new lint error in the checks, I think so.
> 
> Let's have a javascript expert such as @bestbeforetoday or @mbwhite comment on the actual changes.

@bestbeforetoday @mbwhite Could you please confirm on this ?

ci/scripts/lint.sh
```diff
@@ -45,8 +43,10 @@ for dir in $dirs; do
         ./gradlew build
       fi
     elif [[ "$dir" =~ "-typescript" ]]; then
-      print "Running TSLint"
-      tslint --project .
+        print "Installing node modules"
+        npm install
+        print "Running Lint"
+        npm run lint
```
bestbeforetoday, line 49: It would be good to combine this into a single test with "-javascript" since exactly the same operation should be carried out for both project types

bestbeforetoday (changes requested):
I think we should also remove the global install of tslint and eslint in the `ci/azure-pipelines.yml` lint job if it's no longer necessary:
```yaml
      - script: npm install -g typescript eslint tslint
        displayName: Install Javascript Linting Deps
```

bestbeforetoday (approved):
👍

PR #509, opened 2021-10-22
Get label for chaincode from metadata.json

mbwhite:

### DIFF
--- a/test-network-k8s/scripts/chaincode.sh
+++ b/test-network-k8s/scripts/chaincode.sh
@@ -157,7 +157,9 @@ function set_chaincode_id() {
   local cc_package=build/chaincode/${CHAINCODE_NAME}.tgz
   cc_sha256=$(shasum -a 256 ${cc_package} | tr -s ' ' | cut -d ' ' -f 1)
 
-  CHAINCODE_ID=${CHAINCODE_LABEL}:${cc_sha256}
+  label=$( jq -r '.label' chaincode/${CHAINCODE_NAME}/metadata.json)
+
+  CHAINCODE_ID=${label}:${cc_sha256}
 }
 
 # Package and install the chaincode, but do not activate.


denyeart:
Can we use the new `peer lifecycle chaincode calculatepackageid`?
https://hyperledger-fabric.readthedocs.io/en/latest/commands/peerlifecycle.html#peer-lifecycle-chaincode-calculatepackageid
A PR should be coming into Fabric in next days to return the package id without any prefix that needs to be stripped off.

mbwhite:
> Can we use the new `peer lifecycle chaincode calculatepackageid`? https://hyperledger-fabric.readthedocs.io/en/latest/commands/peerlifecycle.html#peer-lifecycle-chaincode-calculatepackageid A PR should be coming into Fabric in next days to return the package id without any prefix that needs to be stripped off.

Yes certainly - when that it's available it's a better approach...

denyeart:
Ok, https://github.com/hyperledger/fabric/pull/3009 is merged to return the packageid.

jkneubuh (changes requested):
@mbwhite can you please remove the `local` keyword from line #142?

I discovered that bash will set $? to the return code from `local`, overriding the error returned by a command within a $(shell eval).  So in this case if the user does not have jq installed, the label variable will be set to the empty string, and `errexit` will ignore the 127 error code from exec-ing jq.

jkneubuh:
@denyeart for the new peer command, it's actually more difficult in this scope to use the new CLI option to generate the chaincode ID.  When this routine runs, it is being executed on the host OS, which may **not** actually have the peer binaries installed locally.  While this may sound draconian, it supports a much simpler workflow if we can keep the fabric binaries as optional -- all of the interaction with the Fabric network should go through the Kube APIs.

We can certainly exec into a pod running in the cluster, issue the `peer lifecycle .... calculatepackageid` command, grab the output, and ferret the results back out to the script running on the host OS.  But in this case I don't believe it's worth the benefit to justify forking the repository to retain backwards compatibility with the 2.3 line.  For this scenario I think it's more appropriate to delay the upgrade until the 2.3 / 2.4 release branches are established.

denyeart:
Sounds good.

mbwhite:
@jkneubuh good point.... didn't know that!

I've updated the script (and also edited the line number in your comment as Github treated it as a link to an issue :-))

jkneubuh (approved):

PR #516, opened 2021-10-26
test-network: Make the regexp of checking version more strict (backpo…

justin-themedium:
(Backport #515)

Make the regular expression of checking version more strict in `test-network/network.sh` for comparing only the main version.

If the version string (by `'peer version'`) includes other "Version: " strings, the comparing version between `LOCAL_VERSION` and `DOCKER_IMAGE_VERSION` would be failed because of the cropping by '`head -1`'.


mbwhite (approved):
Thanks!

PR #519, opened 2021-10-27
nit fixs for test network k8s

SamYuan1990:
1. busybox latest image tag
1. making pod image point to local reg
1. pull images and push images to local reg
1. make network down able to re run

Signed-off-by: Sam Yuan <yy19902439@126.com>


SamYuan1990:
@jkneubuh , I got some time today and tried with `localhost:5000` registry.
I had update the PR, please take a look at.

SamYuan1990:
I am curious that if someone build image as
```
docker build -t localhost:5000/someimage:latest
```
will the k8s pod auto upgrade?
or...
we still need 
```
docker build -t localhost:5000/someimage:latest
kind load docker-image -n kind localhost:5000/someimage
```

test-network-k8s/scripts/test_network.sh
```diff
@@ -247,7 +247,13 @@ function stop_services() {
 
 function scrub_org_volumes() {
   push_fn "Scrubbing Fabric volumes"
-
+  kubectl -n $NS delete jobs --all
+  #kubectl delete pvc fabric-org0 -n test-network
```
mbwhite, line 251: Good idea to remove the commented out code that's not needed anymore

test-network-k8s/scripts/kind.sh
```diff
@@ -115,6 +135,9 @@ function kind_init() {
   kind_create
   apply_nginx_ingress
   launch_docker_registry
+
+  pull_docker_images
+  # load_kind_image_plane
```
mbwhite, line 140: worth removing this; or change to a comment saying that the docker images are pushed to a local registry rather than loaded into KIND.

mbwhite:
@SamYuan1990 very minor couple of comments on the latest changes.
@jkneubuh all happy?

test-network-k8s/scripts/kind.sh
SamYuan1990: @mbwhite, how about below?
```
+ pull_docker_images
+ push_images_to_local
```

SamYuan1990:
@mbwhite / @jkneubuh  please help rerun failed ci cases.

mbwhite (approved):

mbwhite:
@SamYuan1990 @jkneubuh ... just running this now - spotted that if I wanted to use the local docker images I've rebuilt, then the `docker pull` fails.... I need to comment out these parts.

Do you think that running with a local docker image.. is sufficiently 'advanced' that this isn't needed?

SamYuan1990:
May I know the details？

I suppose if you want to use your local image to as refresh into the kind.
It may need to 
Docker push localhost:5000/your_image
Kind load localhost:5000/your_image

For my understanding, if the kind based auto trigger a refresh with image update?

@mbwhite

SamYuan1990:
> @SamYuan1990 @jkneubuh ... just running this now - spotted that if I wanted to use the local docker images I've rebuilt, then the `docker pull` fails.... I need to comment out these parts.
> 
> Do you think that running with a local docker image.. is sufficiently 'advanced' that this is ok.

btw, https://github.com/hyperledger/fabric-samples/pull/519#issuecomment-955282412

SamYuan1990:
> @SamYuan1990 @jkneubuh ... just running this now - spotted that if I wanted to use the local docker images I've rebuilt, then the `docker pull` fails.... I need to comment out these parts.
> 
> Do you think that running with a local docker image.. is sufficiently 'advanced' that this is ok.

or maybe as a balance?
we have a file mapping local image and local repo
for ex
```
fabric-ca:tag localhost:5000:fabric-ca:latest
```

```
docker pull fabric-ca:tag 
docker tag fabric-ca:tag  localhost:5000:fabric-ca:latest
docker push localhost:5000:fabric-ca:latest
```

but the default value for the file(images) will as 
${FABRIC_CONTAINER_REGISTRY}/fabric-ca:$FABRIC_CA_VERSION localhost:5000:fabric-ca:$FABRIC_CA_VERSION

@jkneubuh , @mbwhite  may I know what do you think of this?

mbwhite:
@SamYuan1990 to recap my scenario I'd built a version of the peer docker image locally I wanted to use in the KIND cluster. 

The `docker pull` will alway try and pull the image down whereas here I wanted it to use the local copy.... 

I think looking at this again today - what we've got is fine. very few people are going to want to run this with locally rebuilt docker images...just contributors - and they (hopefully?) know what they are doing... 

Let's leave everything as is... thanks for the contribution and discussion though @SamYuan1990

jkneubuh:
Hi @SamYuan1990 and @mbwhite 

I agree and this PR is OK in the current form.  I may want to make a couple of changes to streamline the usage for Fabric development.  The problem with using the KIND image plane is that the `imagePullPolicy: Always` will always check the repository URL to see if a different image should be loaded, overriding the local images pushed to the cluster.  On the other hand, setting `imagePullPolicy: IfNotPresent` allows the user to load into the KIND image plane, but requires the user to ensure everything has been loaded into the cluster, which is really inconvenient.  I understand that the "easy path" of loading everything from the public repositories is not actually "easy" for everyone, especially when there are problems accessing the network.

What I would propose as an addition, for a future PR, is to include an additional action for the `/network` script that will run the docker pull, tag, and push to the localhost:5000 registry.  For users with poor network connectivity, the command can be run prior to initializing the network to stage all the images locally.  The scenario would look something like: 


``` 
export TEST_NETWORK_REPO=localhost:5000/hyperledger

./network load-images 

./network up 
```

Matthew this I think could actually make it a little easier for the local Fabric development pipeline.  We could break up the load-images into two phases such that it can tag/push all of the outputs from a `make docker` without overwriting local edits with images from the Internet.

jkneubuh:
BTW pushing a set of images up to a container registry can be very helpful for testing local edits to Fabric binaries, but on a remote cluster / target.   For instance, I had to upload some images to de.icr.io to run some remote testing on an IKS cluster.  This command would have helped in that circumstance.

mbwhite:
@jkneubuh very happy with that plan..

jkneubuh:
Hi Sam and Matthew.

I apologize for the confusion, but I'd like to back out all of these changes.  I tried working with the new environment today to work with some custom Fabric images, and it caused more problems than solutions.  Working with the localhost:5000 registry is a good fit for chaincode, gateway applications, and local apps being developed that work _with_ fabric.  But it's not a good fit for the core fabric images.

Let's try to come up with another workaround that will get Sam going, as well: 

- Revert the image pull policy to "IfNotPresent" 
- Point all of the images to ${FABRIC_CONTAINER_REGISTRY}/fabric-peer:${FABRIC_VERSION} 
- Add a new function `load_docker_images` that will PULL the images and run `kind load docker-image ...`
- Run `load_docker_images` from `network kind ` if TEST_NETWORK_STAGE_IMAGES=true 


Here are some use cases: 

a)  Sam gets timeouts when reading from docker hub, causing deployments to fail in KIND:  use the kind image plane. 
```
export TEST_NETWORK_STAGE_IMAGES=true

./network kind     # << will run docker pull && kind load docker-image ... 
./network up 
``` 

b)  Matthew wants to build a *LOCAL* fabric peer binary for testing:   use the kind image plane. 
```
./network kind 

make docker 
kind load docker-image hyperledger/fabric-peer:2.4.0 

export TEST_NETWORK_FABRIC_VERSION=2.4.0 
./network up 
```

c)  Josh wants to run the nightly / latest builds locally for testing:  read images from ghcr.io: 

```
./network kind 

export TEST_NETWORK_FABRIC_CONTAINER_REGISTRY=ghcr.io
export TEST_NETWORK_VERSION=amd64-2.4.0-snapshot-a956e72a6 

./network up 
```

PR #532, opened 2021-11-10
Setup basic HA features for the Peers

mbwhite:
- Add a new label to each peer to mark which org it is in
- Create new service per org, that matches this label - so the service can pick from one of multiple pods
- Update the kubeproxy to give more choice of ha stratergies
- Update the application configmaps and samples to refer to this new service rather than specific peers

### DIFF
--- a/test-network-k8s/docs/HIGH_AVAILABILITY.md
+++ b/test-network-k8s/docs/HIGH_AVAILABILITY.md
@@ -1,0 +1,71 @@
+# High Availability
+
+The peers have been configured so they implemented a essential failover/high-availability configuration.
+
+Two important notes:
+
+1. The word 'gateway' in the k8s definitions is being used in a generic way. It is not tied to the concept of the 'Fabric Gateway' component. However using the 'Fabric-Gateway' with the udpated SDKs, make connecting to Fabric even easier. There is a single connection, that can easily be handled with core k8s abilities. Attempting the approach described below with the older SDKs is not recommended.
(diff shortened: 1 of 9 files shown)


mbwhite:
@jkneubuh  this should be workable today.

Any thoughts appreciated... I wasn't sure about the new label naming.. so suggestions welcome

jkneubuh:
I suggest changing the service name to ${org}-peer-gateway-svc, or some position to include 'gateway' in the name.  It's hard to differentiate between org1-peer1 and org1-peer-svc.  For the selector labels, maybe switch from `block` to `org` ?

On the ipvs attribute for the cluster config ... I think this is OK but I wonder if you encountered any issues when running on a Windows host?  Since this enables features that are implemented in the Docker container it should be OK.

mbwhite:
> I suggest changing the service name to ${org}-peer-gateway-svc, or some position to include 'gateway' in the name. It's hard to differentiate between org1-peer1 and org1-peer-svc. For the selector labels, maybe switch from `block` to `org` ?
>

Yes - good suggestions... #
 
> On the ipvs attribute for the cluster config ... I think this is OK but I wonder if you encountered any issues when running on a Windows host? Since this enables features that are implemented in the Docker container it should be OK.
Nope worked like a charm :-)

mbwhite:
@jkneubuh updates made - now with a README :-)

jkneubuh (approved):
This is amazing.

K8s does a great job of offloading the gateway connection failover, client load balancing, and discovery with an "out of the box" configuration.  This solution is simple, effective, and clear.

I approve!

Issue #511, opened 2021-10-25
test-network-k8s


itcalvin:
./network chaincode deploy：

Deploying chaincode "asset-transfer-basic":
✅  - Packaging chaincode folder chaincode/asset-transfer-basic ...
✅  - Transferring chaincode archive to org1 ...
✅  - Installing chaincode for org org1 ...
⚠️ - Launching chaincode container "ghcr.io/hyperledgendary/fabric-ccaas-asset-transfer-basic" ...

![image](https://user-images.githubusercontent.com/11419457/138677989-bde3b709-b159-4889-9c85-10af3da06f3f.png)

denyeart:
Is this still an issue? Next steps?

SamYuan1990:
I suppose we can merge this issue with https://github.com/hyperledger/fabric-samples/pull/528 @itcalvin

denyeart:
Closing!

PR #533, opened 2021-11-10
Address Issue #511 with docs and better error handling

jkneubuh:
The chaincode installation script had an error that was masking the lack of `shasum` on an end-user's machine.  This made it really hard to pick out the root cause of a downstream error.

This PR addresses the first part / root cause of Issue #511, which can be closed after this PR lands. 

This snippet will not crash with error.  $? is set to the return value of the `local` command.  
```
set -o errexit 

local foo=$(broken-command | tr -s ' ' | cut -d ' ' -f 1)
```

This snippet will error when `broken-command` terminates with a nonzero exit code: 
```
set -o errexit 

foo=$(broken-command | ...)
```

### DIFF
--- a/test-network-k8s/docs/KUBERNETES.md
+++ b/test-network-k8s/docs/KUBERNETES.md
@@ -165,6 +165,8 @@ export TEST_NETWORK_FABRIC_CA_VERSION=amd64-latest
 
 make docker    # in hyperledger/fabric 
 
+export TEST_NETWORK_FABRIC_VERSION=2.4.0
+
 ./network load-images 
 ./network up 
 ```
--- a/test-network-k8s/docs/KUBERNETES.md
+++ b/test-network-k8s/docs/KUBERNETES.md
@@ -131,8 +131,43 @@ to Pods deployed to the local cluster.
 
 For dev/test/CI based flows using an external registry, the traditional Kubernetes practice of 
 [Adding ImagePullSecrets to a service account](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#add-imagepullsecrets-to-a-service-account) 
-still applies. 
+still applies.
 
+In some environments, KIND may encounter issues loading the Fabric docker images from the public container 
+registries.  In addition, for Fabric development it can be advantageous to work with Docker images built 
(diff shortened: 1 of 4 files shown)


SamYuan1990 (approved):
LGTM, BTW, do we have any plan to add test-network-k8s in CI for this repo?

denyeart (commented):
It looks like some other changes were also picked up in this PR. Did you intend to make all these changes in this PR?

jkneubuh:
The one additional change in this PR is that the chaincode image was incorrectly pointing to the local registry, instead of the ghcr.io.   

Sam good question about the CI, there are a lot of test cases set up to run the fabric samples against the Docker Compose based test-network.  I will take a look to see if / how these could be tied in to the Kubernetes test network.

test-network-k8s/docs/KUBERNETES.md
```diff
@@ -131,8 +131,43 @@ to Pods deployed to the local cluster.
...
+
+1.  For network-constrained environments, pull all images to the local docker cache and load to KIND: 
+```shell
+export TEST_NETWORK_STAGE_DOCKER_IMAGES=true 
+
+./network kind 
+./network up  
+```
+
+2.  For alternate registries (e.g. local or Fabric CI/CD builds):
```
mbwhite, line 151: For images that I've built myself locally, do I need to so step2 AND step3 or just step3?

mbwhite (changes requested):

test-network-k8s/docs/KUBERNETES.md
jkneubuh: For local development, only step #3 is required.  **However** ... the `make docker` on the mainline will currently tag the images as 2.4.0, whereas the test network scripts are by default running 2.3.2.   So to run local edits, the revision needs to match what's being loaded into KIND.  I added an extra step in the doc as: 

```
export TEST_NETWORK_FABRIC_VERSION=2.4.0

./network load-images
./network up
```

mbwhite: that's useful thanks.. 

is the word 'local' in step2 confusing?

mbwhite (approved):

PR #535, opened 2021-11-10
Retire dormant maintainers

denyeart:
Several fabric-samples maintainers have moved on to other projects.
See maintainer policies at
https://hyperledger-fabric.readthedocs.io/en/latest/CONTRIBUTING.html#becoming-a-maintainer


mbwhite (approved):

lehors (approved):
Makes sense to me.

lehors (changes requested):
oh wait, shouldn't we have a list of former/retired maintainers further down rather than remove them entirely?

lehors (approved):
Well, we haven't kept the record of retired maintainers before so I guess it's ok to continue doing so.

PR #542, opened 2021-11-22
Extend test network to support prometheus/grafana performances framework. Fixes hyperledger/fabric-test#347

fraVlaca:

### DIFF
--- a/test-network/docker/docker-compose-test-net.yaml
+++ b/test-network/docker/docker-compose-test-net.yaml
@@ -43,7 +43,8 @@ services:
       - ORDERER_ADMIN_TLS_ROOTCAS=[/var/hyperledger/orderer/tls/ca.crt]
       - ORDERER_ADMIN_TLS_CLIENTROOTCAS=[/var/hyperledger/orderer/tls/ca.crt]
       - ORDERER_ADMIN_LISTENADDRESS=0.0.0.0:7053
-      - ORDERER_OPERATIONS_LISTENADDRESS=0.0.0.0:17050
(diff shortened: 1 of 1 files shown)


test-network/docker/docker-compose-test-net.yaml
```diff
@@ -43,7 +43,8 @@ services:
       - ORDERER_ADMIN_TLS_ROOTCAS=[/var/hyperledger/orderer/tls/ca.crt]
       - ORDERER_ADMIN_TLS_CLIENTROOTCAS=[/var/hyperledger/orderer/tls/ca.crt]
       - ORDERER_ADMIN_LISTENADDRESS=0.0.0.0:7053
-      - ORDERER_OPERATIONS_LISTENADDRESS=0.0.0.0:17050
+      - ORDERER_OPERATIONS_LISTENADDRESS=orderer.example.com:9443
```
mbwhite, line 46: You've updated the port, but also changed the address that operations listener is working on. 
Did it not work with 0.0.0.0, as that usually means the socket listener is bound to all networking interfaces?

fraVlaca: to communicate between containers you can use the names of the services (orderer, peer...) as ordinary host names and they’ll resolve to the internal IP address of the container, whatever it happens to be. I can change it back to 0.0.0.0 if you prefer, but I thought this way was more clear

nikhil550 (approved):

davidkel:
referencing #544

PR #561, opened 2021-12-14
Go samples Migration to fabric-gateway

sapthasurendran:


asset-transfer-basic/application-gateway-go/assetTransfer.go
```diff
@@ -0,0 +1,258 @@
...
+	exampleErrorHandling(contract)
+
+	log.Println("============ application-golang ends ============")
+}
+
+
+
+func initLedger(contract *client.Contract){
+	fmt.Printf("Submitting \"InitLedger\" transaction. \n", )
+	// Submit transaction, blocking until the transaction has been committed on the ledger
```
bestbeforetoday, line 94: The tutorial describes the blocking behaviour of the call so I would remove this, similar to the TypeScript sample

asset-transfer-basic/application-gateway-go/assetTransfer.go
```diff
@@ -0,0 +1,258 @@
...
+	_, err := contract.SubmitTransaction("InitLedger")
+	if err != nil {
+		panic(fmt.Errorf("failed to submit transaction: %w", err))
+	}
+	fmt.Printf("InitLedger Transaction successful. \n")
+
+
```
bestbeforetoday, line 101: Not sure the reason for the trailing whitespace at the end of the function

asset-transfer-basic/application-gateway-go/assetTransfer.go
```diff
@@ -0,0 +1,258 @@
...
+	fmt.Println("exampleErrorHandling:")
+	exampleErrorHandling(contract)
+
+	log.Println("============ application-golang ends ============")
+}
+
+
+
+func initLedger(contract *client.Contract){
```
bestbeforetoday, line 92: Some contextual information for the work done in this function could be added as function comments, similar to the TypeScript sample.

asset-transfer-basic/application-gateway-go/assetTransfer.go
```diff
@@ -0,0 +1,258 @@
...
+	// Create a Gateway connection for a specific client identity
+	gateway, err := client.Connect(
+		id,
+		client.WithSign(sign),
+		client.WithClientConnection(clientConnection),
+		// Default timeouts for different gRPC calls
+		client.WithEvaluateTimeout(5*time.Second),
+		client.WithEndorseTimeout(15*time.Second),
+		client.WithSubmitTimeout(5*time.Second),
+		client.WithCommitStatusTimeout(1*time.Minute),
```
bestbeforetoday, line 59: It is good practice to have these timeouts set. We should add the same settings to the TypeScript sample.

asset-transfer-basic/application-gateway-go/assetTransfer.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+
+func getAllAssets(contract *client.Contract){
+	fmt.Println("Evaluating \"GetAllAssets\" query.")
+	evaluateResult, err := contract.EvaluateTransaction("GetAllAssets")
+	if err != nil {
+		panic(fmt.Errorf("failed to evaluate transaction: %w", err))
+	}
+	fmt.Printf("GetAllAssets query result = %s\n", string(evaluateResult))
```
bestbeforetoday, line 111: Check the output from this line when the sample is run. I suspect the JSON string will be collapsed to a single line with no formatting and very hard to read. If so, it would be worth trying to format it, perhaps with [json.Indent()](https://pkg.go.dev/encoding/json#Indent)

asset-transfer-basic/application-gateway-go/assetTransfer.go
```diff
@@ -0,0 +1,258 @@
...
+	fmt.Printf("CreateAsset transaction successful.\n")
+}
+
+func readAssetByID(contract *client.Contract){
+	fmt.Printf("Evaluating \"ReadAsset\" query with given assetID %s .\n",assetId)
+	evaluateResult, err := contract.EvaluateTransaction("ReadAsset", assetId )
+	if err != nil {
+		panic(fmt.Errorf("failed to evaluate transaction: %w", err))
+	}
+	fmt.Printf("ReadAsset query result = %s\n", string(evaluateResult))
```
bestbeforetoday, line 130: Same JSON formatting comment as for `getAllAssets()`

asset-transfer-basic/application-gateway-go/go.mod
```diff
@@ -0,0 +1,11 @@
+module assetTransfer.go
+
+go 1.14
```
bestbeforetoday, line 3: We need Go 1.16 for fabric-gateway

bestbeforetoday (changes requested):
1. I think we need all the messages that go to the terminal to be consistent with the ones you chose for the TypeScript sample, since they are also included in the tutorial as expected terminal output.
2. For general readability, I would prefer to see a blank line either between code that actually invokes the API and the console output lines that surround them. It would add some visual separation and make it easier for readers to focus on exactly the code that invokes the API.
3. There is often several lines of whitespace between functions. I'm not sure how helpful this is as viewers will be able to see less of the code on a single screen. I'd be tempted to have the usual single blank line between functions.
4. The function to create gRPC connection, identity and signer should probably be nearer the top so the order of functions in the file match the order they are invoked from the `main()` function. It's a shame the error handling makes these function quite long, but the ordering should make it easier for readers to follow the flow of execution from top to bottom.

asset-transfer-basic/application-gateway-go/assetTransfer.go
```diff
@@ -0,0 +1,258 @@
...
+	cryptoPath   = "../../test-network/organizations/peerOrganizations/org1.example.com"
+	certPath     = cryptoPath + "/users/User1@org1.example.com/msp/signcerts/cert.pem"
+	keyPath      = cryptoPath + "/users/User1@org1.example.com/msp/keystore/"
+	tlsCertPath  = cryptoPath + "/peers/peer0.org1.example.com/tls/ca.crt"
+	peerEndpoint = "localhost:7051"
+	gatewayPeer  = "peer0.org1.example.com"
+	channelName  = "mychannel"
+	chaincodeName= "basic"
+)
+	var assetId  = "asset" + strconv.FormatInt(time.Now().Unix(),10);
```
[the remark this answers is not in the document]
sapthasurendran: UnixMilli() function  is not available with go 1.1.4

denyeart:
For a single feature we typically like to see a single commit in the main commit history.
Here are instructions for squashing your commits into a single commit:
https://hyperledger-fabric.readthedocs.io/en/latest/github/github.html#updating-a-pull-request
In the future it may be easier to use the `git commit --amend` approach that is described instead of creating new commits.

bestbeforetoday:
> For a single feature we typically like to see a single commit in the main commit history.

I definitely agree with having a single commit for the merged PR, but we can get regardless of whether multiple commits are pushed to the PR using **Squash and merge**. Pushing a new commit to address review comments does make it much easier to see what changed after the last review, whereas always squashing commits makes it impossible.

asset-transfer-basic/application-gateway-go/assetTransfer.go
```diff
@@ -0,0 +1,269 @@
...
+	certPath     = cryptoPath + "/users/User1@org1.example.com/msp/signcerts/cert.pem"
+	keyPath      = cryptoPath + "/users/User1@org1.example.com/msp/keystore/"
+	tlsCertPath  = cryptoPath + "/peers/peer0.org1.example.com/tls/ca.crt"
+	peerEndpoint = "localhost:7051"
+	gatewayPeer  = "peer0.org1.example.com"
+	channelName  = "mychannel"
+	chaincodeName= "basic"
+)
+    var now = time.Now()
+	var assetId = fmt.Sprintf("asset%d", now.UnixNano()/int64(1e6))
```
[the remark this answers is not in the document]
bestbeforetoday: This method of calculating epoch milliseconds will fail for dates outside of a reasonable range, whereas the code I gave you earlier will always work. But this is less code and should work fine if the system clock is set correctly so whichever you prefer. If you are going with this approach then there is no need for a separate `now` variable though.

ci/scripts/lint.sh
```diff
@@ -8,7 +8,7 @@ function print() {
 	echo -e "${GREEN}${1}${NC}"
 }
 
-go install golang.org/x/tools/cmd/goimports@latest
+go get golang.org/x/tools/cmd/goimports@latest
```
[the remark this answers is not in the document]
sapthasurendran: It was throwing me error with version 1.14 , `can only use path@version syntax with 'go get'`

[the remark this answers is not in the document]
bestbeforetoday: The build uses Go 1.16.7. There is no need to make it work with Go 1.14.

bestbeforetoday (approved):
👍

mbwhite (approved):
LGTM

PR #574, opened 2022-01-10
Private data samples migration

sapthasurendran:
Signed-off-by: sapthasurendran <saptha.surendran@ibm.com>

### DIFF
--- a/asset-transfer-private-data/application-gateway-typescript/src/app.ts
+++ b/asset-transfer-private-data/application-gateway-typescript/src/app.ts
@@ -1,0 +1,286 @@
+/*
+ * Copyright IBM Corp. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { connect, Contract } from '@hyperledger/fabric-gateway';
+import { TextDecoder } from 'util';
+import {
+    certPathOrg1, certPathOrg2, keyDirectoryPathOrg1, keyDirectoryPathOrg2, newGrpcConnection, newIdentity,
+    newSigner, peerEndpointOrg1, peerEndpointOrg2, peerNameOrg1, peerNameOrg2, tlsCertPathOrg1, tlsCertPathOrg2
+} from './connect';
+
+const channelName = 'mychannel';
+const chaincodeName = 'private';
+const mspIdOrg1 = 'Org1MSP';
+const mspIdOrg2 = 'Org2MSP';
+
+const utf8Decoder = new TextDecoder();
+
+// Collection Names
+const org1PrivateCollectionName = 'Org1MSPPrivateCollection';
+const org2PrivateCollectionName = 'Org2MSPPrivateCollection';
+
+const RED = '\x1b[31m\n';
+const RESET = '\x1b[0m';
+
+// Use a unique key so that we can run multiple times
+const now = Date.now();
+const assetID1 = `asset${now}`;
+const assetID2 = `asset${now + 1}`;
+
+async function main(): Promise<void> {
+    const clientOrg1 = await newGrpcConnection(
+        tlsCertPathOrg1,
+        peerEndpointOrg1,
+        peerNameOrg1
+    );
+
+    const gatewayOrg1 = connect({
+        client: clientOrg1,
+        identity: await newIdentity(certPathOrg1, mspIdOrg1),
+        signer: await newSigner(keyDirectoryPathOrg1),
+    });
+
+    const clientOrg2 = await newGrpcConnection(
+        tlsCertPathOrg2,
+        peerEndpointOrg2,
+        peerNameOrg2
+    );
+
+    const gatewayOrg2 = connect({
+        client: clientOrg2,
+        identity: await newIdentity(certPathOrg2, mspIdOrg2),
+        signer: await newSigner(keyDirectoryPathOrg2),
+    });
+
+    try {
+        // Get the smart contract as an Org1 client.
+        const contractOrg1 = gatewayOrg1
+            .getNetwork(channelName)
+            .getContract(chaincodeName);
+
+        // Get the smart contract as an Org2 client.
+        const contractOrg2 = gatewayOrg2
+            .getNetwork(channelName)
+            .getContract(chaincodeName);
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org1 Client ~~~~~~~~~~~~~~~~');
+
+        // Create new assets on the ledger.
+        await createAssets(contractOrg1);
+
+        // Read asset from the Org1's private data collection with ID in the given range.
+        await getAssetsByRange(contractOrg1);
+
+        try{
+            //Attempt to transfer asset without prior aprroval from Org2, transaction expected to fail.
+            console.log('\nAttempt TransferAsset without prior AgreeToTransfer');
+            await transferAsset(contractOrg1, assetID1);
+            doFail('TransferAsset transaction succeeded when it was expected to fail');
+        }
+        catch(e){
+            console.log(`*** Received expected error: ${e}`);
+        }
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org2 Client ~~~~~~~~~~~~~~~~');
+
+        // Read the asset by ID.
+        await readAssetByID(contractOrg2, assetID1);
+
+        // Make agreement to transfer the asset from Org1 to Org2.
+        await agreeToTransfer(contractOrg2, assetID1);
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org1 Client ~~~~~~~~~~~~~~~~');
+
+        // Read transfer agreement.
+        await readTransferAgreement(contractOrg1, assetID1);
+
+        // Transfer asset to Org2.
+        await transferAsset(contractOrg1, assetID1);
+
+        // Again ReadAsset : results will show that the buyer identity now owns the asset.
+        await readAssetByID(contractOrg1, assetID1);
+
+        // Confirm that transfer removed the private details from the Org1 collection.
+        const org1ReadSuccess = await readAssetPrivateDetails(contractOrg1, assetID1, org1PrivateCollectionName);
+        if (org1ReadSuccess) {
+            doFail(`Asset private data still exists in ${org1PrivateCollectionName}`);
+        }
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org2 Client ~~~~~~~~~~~~~~~~');
+
+        // Org2 can read asset private details: Org2 is owner, and private details exist in new owner's Collection
+        const org2ReadSuccess = await readAssetPrivateDetails(contractOrg2, assetID1, org2PrivateCollectionName);
+        if (!org2ReadSuccess) {
+            doFail(`Asset private data not found in ${org2PrivateCollectionName}`);
+        }
+
+        try {
+            console.log('\nAttempt DeleteAsset using non-owner organization');
+            await deleteAsset(contractOrg2, assetID2);
+            doFail('DeleteAsset transaction succeeded when it was expected to fail');
+        } catch (e) {
+            console.log(`*** Received expected error: ${e}`);
+        }
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org1 Client ~~~~~~~~~~~~~~~~');
+
+        // Delete AssetID2 as Org1.
+        await deleteAsset(contractOrg1, assetID2);
+    } finally {
+        gatewayOrg1.close();
+        clientOrg1.close();
+
+        gatewayOrg2.close();
+        clientOrg2.close();
+    }
+}
+
+main().catch((error) => {
+    console.error('******** FAILED to run the application:', error);
+    process.exitCode = 1;
+});
+
+/**
+ * Submit a transaction synchronously, blocking until it has been committed to the ledger.
+ */
+async function createAssets(contract: Contract): Promise<void> {
+    const assetType = 'ValuableAsset';
+
+    console.log(`\n--> Submit Transaction: CreateAsset, ID: ${assetID1}`);
+
+    const asset1Data = {
+        objectType: assetType,
+        assetID: assetID1,
+        color: 'green',
+        size: 20,
+        appraisedValue: 100,
+    };
+
+    await contract.submit('CreateAsset', {
+        transientData: { asset_properties: JSON.stringify(asset1Data) },
+    });
+
+    console.log('*** Transaction committed successfully');
+    console.log(`\n--> Submit Transaction: CreateAsset, ID: ${assetID2}`);
+
+    const asset2Data = {
+        objectType: assetType,
+        assetID: assetID2,
+        color: 'blue',
+        size: 35,
+        appraisedValue: 727,
+    };
+
+    await contract.submit('CreateAsset', {
+        transientData: { asset_properties: JSON.stringify(asset2Data) },
+    });
+
+    console.log('*** Transaction committed successfully');
+}
+
+async function getAssetsByRange(contract: Contract): Promise<void> {
+    // GetAssetByRange returns assets on the ledger with ID in the range of startKey (inclusive) and endKey (exclusive).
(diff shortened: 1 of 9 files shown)


asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+        // Get a network instance representing the channel where the smart contract is deployed.
+        const networkOrg2 = gatewayOrg2.getNetwork(channelName);
+
+        // Get the smart contract from the network.
+        const contractOrg2 = networkOrg2.getContract(chaincodeName);
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org1 Client ~~~~~~~~~~~~~~~~');
+
+        // Create new assets on the ledger.
+        await createAsset(contractOrg1);
```
bestbeforetoday, line 170: This creates two assets. That's fine but to be picky, maybe either:
- the function name should be `createAssets()` instead of `createAsset()`; or
- there should be two separate functions; or
- there should be one `createAsset()` function that additionally takes the asset to create as an argument

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+
+        //Read  asset from the org1's private data collection.
+        await readAssetPrivateDetails(contractOrg1,org1PrivateCollectionName,assetID1);
```
bestbeforetoday, line 173: This function takes the asset ID as a parameter, whereas all the others use the asset ID const variable directly. It would be good to be consistent, whichever way 'round you prefer.

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+        await readTransferAgreement(contractOrg1);
+
+        // Tranfer asset to org2.
+        await transferAsset(contractOrg1);
+
+        //Again ReadAsset : results will show that the buyer identity now owns the asset.
+        await readAssetByID(contractOrg1);
+
+        //Confirm that transfer removed the private details from the Org1 collection.
+        await checkPrivateDataExist(contractOrg1);
```
bestbeforetoday, line 198: This actually checks the private data **no longer** exists. It might be better to reflect that in the naming.

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+    const files = await fs.readdir(keyDirectoryPath);
+    const keyPath = path.resolve(keyDirectoryPath, files[0]);
+    const privateKeyPem = await fs.readFile(keyPath);
+    const privateKey = crypto.createPrivateKey(privateKeyPem);
+    return signers.newPrivateKeySigner(privateKey);
+}
+
+function doFail(msgString: string) {
+    console.error(`${RED}\t${msgString}${RESET}`);
+    process.exit(1);
```
bestbeforetoday, line 248: `proxess.exit()` is very aggressive. Console output may not get written before the process terminates so you may not see the errors. It might be better to thrown an error to terminate execution more cleanly

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+
+async function newSigner(keyDirectoryPath: string): Promise<Signer> {
+    const files = await fs.readdir(keyDirectoryPath);
+    const keyPath = path.resolve(keyDirectoryPath, files[0]);
+    const privateKeyPem = await fs.readFile(keyPath);
+    const privateKey = crypto.createPrivateKey(privateKeyPem);
+    return signers.newPrivateKeySigner(privateKey);
+}
+
+function doFail(msgString: string) {
```
bestbeforetoday, line 246: This should have a return type of [never](https://www.typescriptlang.org/docs/handbook/2/functions.html#never)

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+    };
+
+    await contract.submit('CreateAsset', {
+        transientData: { asset_properties: JSON.stringify(asset2Data) },
+    });
+
+    console.log('*** Transaction committed successfully');
+}
+
+async function readAssetPrivateDetails(contract: Contract,collectionName:string,assetID:string) {
```
bestbeforetoday, line 293: Should have a return type

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+
+    const resultString = utf8Decoder.decode(resultBytes);
+    if (!resultString || resultString.length === 0) {
+        doFail('Received empty query list for ReadAssetPrivateDetails');
+    }
+    const result = JSON.parse(resultString);
+    console.log('*** Result:', result);
+}
+
+async function transferAssetWithoutAprroval(contract: Contract) {
```
bestbeforetoday, line 313: Should have a return type

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+            transientData: { asset_owner: JSON.stringify(buyerDetails) },
+        });
+
+        console.log('******** FAILED: above operation expected to return an error');
+    } catch (error) {
+        console.log(`   Successfully caught the error: \n    ${error}`);
+    }
+}
+
+async function readAssetByID(contract: Contract) {
```
bestbeforetoday, line 332: Should have a return type

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+
+    const resultString = utf8Decoder.decode(resultBytes);
+    if (!resultString || resultString.length === 0) {
+        doFail('Received empty result for ReadAsset');
+    }
+    const result = JSON.parse(resultString);
+    console.log('*** Result:', result);
+}
+
+async function agreeToTransfer(contract: Contract) {
```
bestbeforetoday, line 344: Should have a return type

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+    );
+
+    await contract.submit('AgreeToTransfer', {
+        transientData: { asset_value: JSON.stringify(dataForAgreement) },
+    });
+
+    console.log('*** Transaction committed successfully');
+}
+
+async function readTransferAgreement(contract: Contract) {
```
bestbeforetoday, line 361: Should have a return type

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+
+    const resultString = utf8Decoder.decode(resultBytes);
+    if (!resultString || resultString.length === 0) {
+        doFail('Received empty result for ReadTransferAgreement');
+    }
+    const result = JSON.parse(resultString);
+    console.log('*** Result:', result);
+}
+
+async function transferAsset(contract: Contract) {
```
bestbeforetoday, line 377: Should have a return type

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+
+    const buyerDetails = { assetID: assetID1, buyerMSP: mspIdOrg2 };
+    await contract.submit('TransferAsset', {
+        transientData: { asset_owner: JSON.stringify(buyerDetails) },
+    });
+
+    console.log('*** Transaction committed successfully');
+}
+
+async function checkPrivateDataExist(contract: Contract) {
```
bestbeforetoday, line 388: Should have a return type

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+        org1PrivateCollectionName,
+        assetID1
+    );
+
+    const resultString = utf8Decoder.decode(resultBytes);
+    if (resultString && resultString.length > 0) {
+        doFail('Expected empty data from ReadAssetPrivateDetails');
+    }
+}
+async function deleteAsset(contract: Contract) {
```
bestbeforetoday, line 402: Should have a return type

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,412 @@
...
+        gatewayOrg1.close();
+        clientOrg1.close();
+
+        gatewayOrg2.close();
+        clientOrg2.close();
+    }
+}
+
+main().catch((error) =>
+    console.error('******** FAILED to run the application:', error)
```
bestbeforetoday, line 218: If you care about the process exit code, you can set it here as [process.exitCode](https://nodejs.org/api/process.html#processexitcode_1) and then let the process reach the end of execution gracefully

asset-transfer-private-data/application-gateway-typescript/src/app.ts
bestbeforetoday: Actually, I'm not really sure of the value of the second asset. The only thing that happens to it is being created here and then deleted near the end of this function. It doesn't seem to be obviously demonstrating any aspect of private data not already covered. I'd be tempted to remove it to simplify the sample. @denyeart might also have an opinion.

bestbeforetoday (changes requested):
I've left lots of comments but they are small things. Mostly 👍

sapthasurendran:
@denyeart Could you please confirm  whether to remove the assetID2 asset or not ?

asset-transfer-private-data/application-gateway-typescript/src/app.ts
bestbeforetoday: It's not the collection name I'm questioning; it's the asset ID parameter. You only ever pass `assetID1` as the value so you could avoid passing that in as an argument and just have the **readAssetPrivateDetails()** function hard-coded to use `assetID1`, just like all the other functions

denyeart:
> @denyeart Could you please confirm whether to remove the assetID2 asset or not ?

@sapthasurendran 

The reason two private assets were created in the original javascript application sample was to demonstrate that you can do queries of multiple private assets. The original sample called the chaincode function GetAssetByRange() to demonstrate this.

The original application sample also attempted two Delete calls, one as the non-owner to demonstrate an auth failure, and one as the owner to demonstrate auth success.

It's not too big of a deal, but I would suggest matching the original sample application functionality in terms of which chaincode calls to make, as the sample calls were put in place for a reason. If you have reason to doubt the original sample intent, that's fine, but I think the default answer would be to mirror the original application functionality.

In terms of which of the two assets to delete at the end, it doesn't really matter.

bestbeforetoday:
I agree with Dave's comments above. Unlike the events sample, where I argued for stripping back the private data elements that did nothing to demonstrate the eventing behaviour, this sample is specifically demonstrating private data so it would be good to be thorough and demonstrate create, read, update (transfer) and delete behaviour.

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,451 @@
...
+    'ca.crt'
+);
+
+// Gateway peer endpoint.
+const peerEndpointOrg1 = 'localhost:7051';
+const peerEndpointOrg2 = 'localhost:9051';
+
+// Gateway peer container name.
+const peerNameOrg1 = 'peer0.org1.example.com';
+const peerNameOrg2 = 'peer0.org2.example.com';
```
bestbeforetoday, line 111: It would be great to move all this connection-specific stuff (including the connection helper functions, like `newGrpcConnection()`) to a separate file, as we did with the events sample. Only the basic sample is really demonstrating how to connect a Gateway. For these other samples, this just adds a lot of extra content that makes it much harder to focus on the bits this sample is really trying to demonstrate.

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,451 @@
...
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org2 Client ~~~~~~~~~~~~~~~~');
+
+        //Non-owner Org2 should not be able to delete assetID2. Expect an error from DeleteAsset
+        await deleteAssetOrg2(contractOrg2);
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org1 Client ~~~~~~~~~~~~~~~~');
+
+        // Delete AssetID2 as Org1.
+        await deleteAssetOrg1(contractOrg1);
```
bestbeforetoday, line 202: We're duplicating the same deleteAsset code, with the only difference being a try-catch around the call we expect to fail. It might be clearer to just have a single `deleteAsset()` function and put the try-catch around the one we expect to fail up here in the main method. As well as avoiding duplication, it helps to visualise the high-level flow in the main() function with the called functions providing details on the mechanics of the calls.

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,451 @@
...
+
+        console.log('\n~~~~~~~~~~~~~~~~ As Org2 Client ~~~~~~~~~~~~~~~~');
+
+        //Org2 can ReadAssetPrivateDetails: Org2 is owner, and private details exist in new owner's Collection
+        await readAssetPrivateDetailsOrg2(contractOrg2);
```
bestbeforetoday, line 207: It might make the high-level flow clearer here if the function is just named `readAssetPrivateData()` and the collection name (`org2PrivateCollectionName`) is passed as an argument along with the contract.

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,451 @@
...
+        clientOrg1.close();
+
+        gatewayOrg2.close();
+        clientOrg2.close();
+    }
+}
+
+main().catch((error) =>{
+    console.error('******** FAILED to run the application:', error)
+    process.exit(1)
```
bestbeforetoday, line 220: I suggest setting [process.exitCode](https://nodejs.org/api/process.html#processexitcode_1) instead. `process.exit()` is quite aggressive and can cause the process to exit before console.log() messages have actually been written to the console.
```suggestion
    process.exitCode = 1;
```

asset-transfer-private-data/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,451 @@
...
+        console.log('\n~~~~~~~~~~~~~~~~ As Org1 Client ~~~~~~~~~~~~~~~~');
+
+        // Create new assets on the ledger.
+        await createAssets(contractOrg1);
+
+        //Read asset from the org1's private data collection with ID in the given range.
+        await getAssetsByRange(contractOrg1);
+
+        //Attempt to transfer asset without prior aprroval from org2, transaction expected to fail.
+        await transferAssetWithoutAprroval(contractOrg1);
```
bestbeforetoday, line 170: This duplicates the API calls made in `transferAsset()`. It might be clearer to call `transferAsset()` here and catch the expected error.

bestbeforetoday (changes requested):
I had a careful look through and I think there is some refactor that would still help reduce duplication and (importantly) help readability of the high-level flow in the main function. See my last commit on top of your working branch here:

https://github.com/bestbeforetoday/fabric-samples/tree/typescript-private-data

Feel free to just pull those changes into your PR.

I think this would really benefit from a README.md at the `asset-transfer-private-data` directory level, similar to the one in `asset-transfer-events`. Just a brief description of what the sample does and (importantly) the commands you need to run to use it. Also a link to the tutorial in case people want to dig deeper, although the tutorial instructions are now out of date as it describes the commands and flow of the legacy samples.

bestbeforetoday (approved):
👍

mbwhite (approved):

PR #581, opened 2022-01-14
Added missing prerequisite

vinnizworld:
`jq` is often not installed on most linux systems, so it must be listed under pre-requisite. If not mentioned, CC installation fails without any descriptive errors.


jkneubuh:
Vinay thanks kindly for catching this and patching up the doc.   While you are in this area of the routines, would you mind adding a `jq` check to `scripts/prereqs.sh` to the PR?

jkneubuh (changes requested):
This looks great.  Thanks Vinay.

Can you also please add a test / check into the `scripts/prereqs.sh` script?   That will be a nice supplement to the new dependency in the docs listing.

vinnizworld:
@jkneubuh I've created a new pull request

PR #582, opened 2022-01-14
Added 'jq' to prerequisite script and readme for k8s test network

vinnizworld:
New pull as per suggestion given on #581


jkneubuh (approved):
Looks great!   Thanks for finding, fixing, and smooshing the bug.

PR #583, opened 2022-01-15
peer address export

PonderWOnder:
environment variable should be exported to take effect i guess...


denyeart:
Since it is a single shell script, I don't think the export is needed.
See https://www.baeldung.com/linux/bash-variables-export
Please re-open if you disagree.

PonderWOnder:
In kubernetes the next command `peer channel list` fails for using localhost:7501. so admin-cli should remember which peer it should talk to...  
So if it shouldn't be global then 
`CORE_PEER_ADDRESS='${org}'-peer2:7051 \
    peer channel list`
will alleviate the error message.

PR #586, opened 2022-01-18
Fix: remove `shift` after `-verbose` 

kth-tw:
There are no argument after `-verbose`. So, it should not run `shift`.
If `-verbose` is not latest parameter, it will throw errro.
image


lehors (approved):
You're right. It's pretty amazing to me that we've had that bug in there for so long without anyone noticing it!...
Thanks!

nikhil550 (approved):

lehors:
I overrode the failing lint test which had nothing to do with this change.

PR #588, opened 2022-01-19
Linting oddity

mbwhite:
Running the find command ahead of the actual loop.

### DIFF
--- a/ci/scripts/lint.sh
+++ b/ci/scripts/lint.sh
@@ -9,6 +9,7 @@ function print() {
 }
 
 go install golang.org/x/tools/cmd/goimports@latest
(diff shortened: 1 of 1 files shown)


lehors:
Could you please explain? How does running the same find command beforehand avoids the problem??
I looked into this problem yesterday and couldn't figure it out so I'm very curious. :-)

mbwhite:
@lehors I've got no idea either; I added that extra find to try and determine what directories where being picked up... I wasn't expecting it to work!

mbwhite:
A rerun of one of the PRs that was affected today passed without any issue.
Also I reran this PR with the same and the main branch, and again all ok.

PR #589, opened 2022-01-21
Remove JCenter as a Gradle package repository

bestbeforetoday:
JCenter is deprecated, can no longer be published to, and is scheduled for removal. It is now causing build failures. Replace with Maven Central.

Also remove mavenLocal() as this is not recommended practice:
    
https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:case-for-maven-local

Note that Jitpack still needs to be included as a package repository for Java chaincode since it has dependencies on an old version of com.github.everit-org.json-schema:org.everit.json.schema that is only published there.

### DIFF
--- a/asset-transfer-basic/application-java/build.gradle
+++ b/asset-transfer-basic/application-java/build.gradle
@@ -18,9 +18,8 @@ ext {
 }
 
 repositories {
-    // Use jcenter for resolving dependencies.
     // You can declare any Maven/Ivy/file repository here.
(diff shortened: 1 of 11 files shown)


jkneubuh:
Hop along, JFrog,
Into the sunset.
Tomorrow's sunshine starts with maven, 
And green builds.


 Approve!

jkneubuh (approved):
👍

PR #593, opened 2022-01-24
ERC20 Java chaincode sample with unit test

renjithpta:
This is an example of the Java chaincode implementation of the ERC20 token. It follows the same logic as the existing chain code in go and nodejs.


renjithpta:
Not followed code commit rule

PR #595, opened 2022-01-24
ERC20 Java chaincode implementation example.

renjithpta:
Signed-off-by: renjithkn@gmail.com <renjithkn@gmail.com>

This is an example of ERC20 token standard Java chaincode implementation.  The functionality is the same as the existing go and nodejs ERC20 chaincode.


jkneubuh:
I really like this PR.  The code is clear, legible, has a good flow, and I enjoyed reading through it.  I'm not familiar with the golang token transfer reference, but I felt like I had learned something about Fabric and chaincode by reading through the commit.  

Here are a few considerations, more for discussion points than for "must have" features on this PR.  (These are more about providing some consistency across the samples as a whole, rather than this specific contract.) 

- Objections to the use of Maven.    This example uses maven to build a shaded Jar.  It would be improved (?) by aligning with general practices and presented a standardized front through the use of Gradle to compile, package, and test the chaincode.  That said, there's nothing inherently _wrong_ with the use of Maven in this example.  That said, building the jar file is just one step in the lifecycle for managing chaincode, and it would be nice if Fabric had some clear guidelines and examples of building Java based Docker images running with Chaincode-as-a-Service.   (This would make a nice follow-on, or additional PR once we have the story sorted out.) 

- Objections to the use of System.out.   It's hard to object to the practice of the routine emitting _anything_ to the chaincode standard output.   I really like seeing slf4j and logback (not log4j) pulled into the Maven dependencies.  Instead of System.out, can we use this as an example to highlight proper logging techniques with slf4j in the routine?

- Objections to the use of org.json :   Similar to the choice of "best logger" and "best build system" -- are we 100% on track with the recommendation of org.json for the JSON handling in the example?   (Likewise are we 100% on track with Gson in other examples?)  If we are going to standardize and recommend a JSON parser, why not use Jackson?

- Objections to the Java Chaincode SDK :   I read through this sample and was reminded that we have a disconnect between `ChaincodeException` and Fabric's adherence to golang Error strings in other parts of the system.  It's even more pronounced in this example, where the code is clean enough that I can understand what's happening, but the context is lost through the use of `ChaincodeException(String error)`.  For instance, how does a chaincode describe the difference between "not found" and "insufficient balance available" ?   In consuming applications, everything will need to dig in and parse the string error out of the exception.  It would be much better (IMHO) if the natural exception / response from a chaincode invocation were a struct, consisting of an application-specific error code coupled with an (optional) explanation. 

```
{ 
  int code;
  String message; 
}
``` 

ChaincodeException sounds to me like : "Network error reading from a DB", or "no space left on device" when in the middle of a chaincode invocation.   Can we improve this somehow?  (Again: not in this PR but it's what came to mind when reviewing the new sample.)


Last bit:  
- Objection to the hard-coded certificates in the test chaincode stub.   What's in these stub certificates?  Will they expire?


Overall, I really like the new sample.  Thank you for the great contribution.

jkneubuh:
ping: @mbwhite and @denyeart and @bestbeforetoday -   any objections to this PR?   

I really like the addition and would like to merge it.  Also can we use this to align on any new-fangled and/or best practices coming out of Chaincode-as-a-Service, Gateway client dev, and Java samples currently in the wings?

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+        }
+
+        // Retrieve the current balance of the recepient
+        CompositeKey toBalanceKey = stub.createCompositeKey(balancePrefix, _to);
+        String toCurrentBalance = stub.getStringState(toBalanceKey.toString().trim());
+
+        long _toCurrentBalance = 0;
+        // If recipient current balance doesn't yet exist, we'll create it with a
+        // current balance of 0
+        if (toCurrentBalance.isBlank() || toCurrentBalance.length() == 0) {
```
bestbeforetoday, line 260:
```suggestion
        if (toCurrentBalance.isBlank()) {
```

bestbeforetoday, line 260:
```suggestion
        if (toCurrentBalance.isBlank()) {
```

bestbeforetoday, line 260: Second check is unreachable as it duplicates the first.
```suggestion
        if (toCurrentBalance.isBlank()) {
```

bestbeforetoday, line 260:
```suggestion
        if (toCurrentBalance.isBlank()) {
```

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+    /**
+     * Allows `spender` to spend `value` amount of tokens from the owner.
+     *
+     * @param {Context} ctx the transaction context
+     * @param {String}  spender The spender
+     * @param {Integer} value The amount of tokens to be approved for transfer
+     * @returns {Boolean} Return whether the approval was successful or not
+     */
+    @Transaction()
+    public boolean approve(final Context ctx, final String spender, final String value) {
```
bestbeforetoday, line 289: This only ever returns true. Maybe `void` would be a better return type

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+     * Set optional infomation for a token.
+     *
+     * @param {Context} ctx the transaction context
+     * @param {String}  name The name of the token
+     * @param {String}  symbol The symbol of the token
+     * @param {String}  decimals The decimals of the token
+     * @param {String}  totalSupply The totalSupply of the token
+     */
+    @Transaction()
+    public boolean setOptions(final Context ctx, final String name, final String symbol, final String decimals) {
```
bestbeforetoday, line 345: Only ever returns true so perhaps `void` return type

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+
+    /**
+     * Mint creates new tokens and adds them to minter's account balance
+     *
+     * @param {Context} ctx the transaction context
+     * @param {Integer} amount amount of tokens to be minted
+     * @returns {Object} The balance
+     */
+    @Transaction()
+    public boolean mint(final Context ctx, final String amount) {
```
bestbeforetoday, line 363: Only ever returns true so perhaps `void` return type

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+
+    /**
+     * Burn redeem tokens from minter's account balance
+     *
+     * @param {Context} ctx the transaction context
+     * @param {Integer} amount amount of tokens to be burned
+     * @returns {Object} The balance
+     */
+    @Transaction()
+    public boolean burn(final Context ctx, final String amount) {
```
bestbeforetoday, line 435:
```suggestion
    public void burn(final Context ctx, final String amount) {
```

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+        }
+
+        CompositeKey balanceKey = stub.createCompositeKey(balancePrefix, minter);
+
+        String currentBalanceBytes = stub.getStringState(balanceKey.toString().trim());
+        // If minter current balance doesn't yet exist, we'll create it with a current
+        // balance of 0
+        long currentBalance = 0;
+
+        if (currentBalanceBytes.isBlank() || currentBalanceBytes.length() == 0) {
```
bestbeforetoday, line 388:
```suggestion
        if (currentBalanceBytes.isBlank()) {
```

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+
+        }
+        long updatedBalance = currentBalance + amountInt;
+
+        stub.putStringState(balanceKey.toString().trim(), String.valueOf(updatedBalance));
+
+        // Increase totalSupply
+        String totalSupplyBytes = stub.getStringState(totalSupplyKey.trim());
+        long totalSupply = 0;
+        if (totalSupplyBytes.isBlank() || totalSupplyBytes.length() == 0) {
```
bestbeforetoday, line 404:
```suggestion
        if (totalSupplyBytes.isBlank()) {
```

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+        }
+
+        String minter = ctx.getClientIdentity().getId();
+
+        long amountInt = Long.parseLong(amount);
+
+        CompositeKey balanceKey = stub.createCompositeKey(balancePrefix, minter);
+
+        String currentBalanceBytes = stub.getStringState(balanceKey.toString().trim());
+        if (currentBalanceBytes.isBlank() || currentBalanceBytes.length() == 0) {
```
bestbeforetoday, line 452:
```suggestion
        if (currentBalanceBytes.isBlank()) {
```

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+            throw new ChaincodeException("The balance does not exist");
+        }
+        long currentBalance = Long.valueOf(currentBalanceBytes.toString());
+        long updatedBalance = currentBalance - amountInt;
+
+        stub.putStringState(balanceKey.toString().trim(), String.valueOf(updatedBalance));
+
+        // Decrease totalSupply
+        String totalSupplyBytes = stub.getStringState(totalSupplyKey.toString().trim());
+        if (totalSupplyBytes.isBlank() || totalSupplyBytes.length() == 0) {
```
bestbeforetoday, line 462:
```suggestion
        if (totalSupplyBytes.isBlank()) {
```

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+    @Transaction()
+    public String getClientAccountID(final Context ctx) {
+        // Get ID of submitting client identity
+        String clientAccountID = ctx.getClientIdentity().getId();
+        return clientAccountID;
+    }
+
+    private byte[] serialize(Object object) {
+        String jsonStr = new JSONObject(object).toString();
+        return jsonStr.getBytes(UTF_8);
```
bestbeforetoday, line 516: This is only ever called with a `JSONObject` as the argument.
```suggestion
    private byte[] serialize(final JSONObject object) {
        return object.toString().getBytes(UTF_8);
```

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+     */
+
+    @Transaction()
+    public long getClientAccountBalance(final Context ctx) {
+        // Get ID of submitting client identity
+        ChaincodeStub stub = ctx.getStub();
+        String clientAccountID = ctx.getClientIdentity().getId();
+        CompositeKey balanceKey = stub.createCompositeKey(balancePrefix, clientAccountID);
+        String balanceBytes = stub.getStringState(balanceKey.toString().trim());
+        if (balanceBytes.isBlank() || balanceBytes.length() == 0) {
```
bestbeforetoday, line 493:
```suggestion
        if (balanceBytes.isBlank()) {
```

bestbeforetoday (commented):
Generally looks fine to me, although I haven't carefully checked all the logic.

Lots of little Java niggles that could be addressed (see comments). Also, all through the code there is lots of trimming of strings going on, like `key.toString().trim()`. Do generated keys really need to be trimmed to ensure they are valid? If not, I would avoid this as it would be easy to introduce a bug by forgetting to trim somewhere, or perhaps even by trimming to produce the same key that would have been generated from a different set of inputs. If they really do need to be trimmed then it might be worth having a helper function to generate and trim keys to make sure it's done consistently rather than duplicating this trimming code throughout the code.

I think the README may need updating to add Java-specific examples to the Go and JavaScript ones already there.

token-erc-20/chaincode-java/src/main/java/org/example/TokenERC20Contract.java
```diff
@@ -0,0 +1,519 @@
...
+     * @returns {Number} Returns the account balance
+     */
+    @Transaction()
+    public long balanceOf(final Context ctx, final String owner) {
+
+        ChaincodeStub stub = ctx.getStub();
+        CompositeKey balanceKey = stub.createCompositeKey(balancePrefix, owner);
+
+        String balance = stub.getStringState(balanceKey.toString().trim());
+        if (balance == null || balance.isEmpty() || balance.length() == 0) {
```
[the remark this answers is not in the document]
renjithpta: code modified as per the review suggests.

lehors:
@renjithpta May I asked why you closed this PR? @bestbeforetoday's review was quite positive.

PR #596, opened 2022-01-26
Support for using podman with the test-network

mbwhite:
## Podman support

A copy of the `install_fabric.sh` script is in the `test-network` directory. This has been enhanced to support a `podman` argument; if used it will use `podman` to pull down images and tag them rather than docker. The images are the same, but need to pulled differently.

The `network.sh` script has been enhanced so that it can use `podman` and `podman-compose` instead of docker. Ensure that `CONTAINER_CLI` is set as below when running anet `network.sh` script. 

```bash
CONTAINER_CLI=podman ./network.sh up
````

As there is no Docker-Daemon in this context, only Chaincode-as-a-service is supported.

### CODE REVIEW

**Comment by jkneubuh** [comment]
@mbwhite I had good "success" using this branch to bring up the test network using a combination of `nerdctl` (docker CLI equivalent) and `nerdctl compose` (docker-compose) as  bundled with http://rancherdesktop.io.  

Effectively: 
```
CONTAINER_CLI="nerdctl" \
CONTAINER_CLI_COMPOSE="nerdctl compose" \
  ./network.sh up 
```

In general this works really well with the PR above.  However, it looks like there may be some subtle differences in how `nerdctl compose` is mounting the volumes, meaning that we _may_ need to juggle multiple sub-folders, one for each compose dialect (docker, podman, nerdctl, etc.)   This is a bummer, it would be nice to maintain one suite of compose files that works equivalently on all platforms.

Would you have any objections to me introducing a new compose asset folder `nerdctl` to align with the direction of this PR?   Something like: 

```
+ test-network/
    | 
    + docker/
    |    docker-compose-ca.yaml
    |    docker-compose-couch.yaml
    |    docker-compose-test-net.yaml
    |
    + podman/
    |    docker-compose-ca.yaml
    |    docker-compose-couch.yaml
    |    docker-compose-test-net.yaml
    | 
    + nerdctl/
    |    docker-compose-ca.yaml
    |    docker-compose-couch.yaml
    |    docker-compose-test-net.yaml
```

and 

```
  if [ "${CONTAINER_CLI}" == "podman" ] || [ "${CONTAINER_CLI}" == "nerdctl" ]; then
    cp ./${CONTAINER_CLI}/core.yaml ./organizations/peerOrganizations/org2.example.com/peers/peer0.org2.example.com/
    cp ./${CONTAINER_CLI}/core.yaml ./organizations/peerOrganizations/org1.example.com/peers/peer0.org1.example.com/
  fi
```

There are a couple of other places where this creeps in, for instance in bringing down the network, addOrg3, etc.  It's a bit messy but I don't see another good way to support a variety of compose files for the test network. 

Is there a better way to (re) organize some of the compose descriptors to support another container runtime?

**Comment by mbwhite** [comment]
@jkneubuh no problem with adding the `nerdctl` folder. 

I've been able to remove the `cp core.yml` in the latest update.  I think we might be able to use core basic compose descriptor that is then extended for specifics.

**Comment by jkneubuh** [comment]
Here's the best compromise I could come up with for containerd / `nerdctl` in the context of this PR.  I DON'T like this approach, but functionally it should work.  podman-compose and nerdctl compose provide close approximations to docker-compose, but there are still a few ways where the differences bleed into the test network.  This will not come as a surprise but I'd like to see us advance the Rancher / KIND / k8s route for the test network + _CCaaS_, providing some compatibility in CI and teams that are still working with compose.

Bias aside, here are the edge conditions where I found `nerdctl` differed from the docker and podman compose.  In general the approach can / will work, despite it still needing a little TLC: 

- nerdctl will refuse to start a service when a volume mount references a file that is not present on the host OS.  For instance, the test net compose references a system.genesis.block in a volume mount, but this is never available locally as the network has been updated to use the osnadmin and channel participation APIs.  While the file/volume can be moved, it highlights that there is a slight behavioral difference between docker and nerdctl compose - docker is OK with the missing file, but nerdctl will refuse to launch the container.

- nerdctl does NOT support the `--filter` flag for `ps` and `rm` activities.  This feature is very convenient for annotating chaincode images constructed with the legacy / internal builder.  As a work around, the images can be manually removed (by name) in the `down` target, or --filter replaced with a little bit of sed/awk/cut/grep.   This is really _messy_ in the test network as the system creates all sorts of images in various contexts.

- It looks like `nerdctl compose` doesn't support the service ENV overrides that are in the couch compose files.  When bringing down the network this emits some scary log output about invalid services in the descriptor.  It's not clear that this will work correctly with couch.

- For some unknown mysterious reason, `nerdctl compose` will **refuse** to mount the MSP and TLS certificate volumes (or any volume, for that matter) into `/var/hyperledger` in the orderer container.  The only clue I can come up with is that the file is somehow created in the docker build process (it's an empty folder on the fabric-orderer image) in a way that prevents the volume mount points from coming up cleanly.  My workaround in this PR was to change the volume path from `/var/hyperledger` to `/opt/hyperledger`, and ignore any desire to track down differences in how containerd and moby are mounting volumes.  Again this is a really subtle difference in how volume mounts work between podman/docker/nerdctl compose, and it's a shame to manage three separate compose descriptor folders to handle the differences.


Overall, my experience on OSX was that it was a real struggle to get podman up and running cleanly, and had a much better experience with the installation to work with nerdctl/rancher.   That said, there are still a few open edge cases / differences between nerdctl and docker that make it a challenge to work with the current test-network.

**Comment by SamYuan1990** [comment]
hi @mbwhite, 

ref to document below, I am not sure if we can have a chance to install `podman` in CI steps and running with `alias docker=podman`?
https://podman.io/whatis.html
https://podman.io/getting-started/installation

Sam

File: asset-transfer-basic/chaincode-java/Dockerfile [hunk 1]
```diff
@@ -24,8 +24,8 @@ RUN addgroup --system javauser && useradd -g javauser javauser
 COPY --chown=javauser:javauser --from=GRADLE_BUILD /home/gradle/build/libs/chaincode.jar /chaincode.jar
 COPY --chown=javauser:javauser docker/docker-entrypoint.sh /docker-entrypoint.sh 
 
-ENV PORT $CC_SERVER_PORT
-EXPOSE $CC_SERVER_PORT
+ENV PORT 9999
```
jkneubuh, line 27: ?

I haven't tested this but this looks "sus."  

Why set the ARG default above, and then stomp on it here with a literal value?

jkneubuh (changes requested):
I think the compose/nerdctl/* and org3/nerdctl dirs+files need to go.  `nerdctl compose` had too many differences in the CLI behavior.   Using rancher's `docker-compose` dockerd(moby) runtime was a _much_ better outcome.

test-network/install-fabric.sh
```diff
@@ -0,0 +1,313 @@
...
+# Argbash is FREE SOFTWARE, see https://argbash.io for more info
+# Generated online by https://argbash.io/generate
+
+# Default values
+_positionals=()
+_arg_comp=('' )
+
+# if version not passed in, default to latest released version
+# if ca version not passed in, default to latest released version
+_arg_fabric_version="2.4.0"
```
jkneubuh, line 24: 2.4.1 ?   2.4 ?   Is there a way to align this somehow with the script returned in the bit.ly link?   Not just the revision, but in how/where it comes from?

I don't like that we need to "install" fabric before using it.  This is way out of context of this PR, but I will take the opportunity to rant here a little bit.   Users should NOT need to install binaries locally to run a container-based application.   Running `peer` commands locally to administer a network .... maybe.  But for casual consumption and sample usage - no.

asset-transfer-basic/chaincode-java/Dockerfile
mbwhite: cold light of day, with fresh coffee.. yes that is 'sus'  
I'll circle back round on why that was an issue.

test-network/install-fabric.sh
mbwhite: There's a PR on fabric (https://github.com/hyperledger/fabric/pull/3208) where this install script really lives.
WIll remove it from this PR..

(and yes I agree with your last point)

mbwhite:
@jkneubuh I think this is good to check again :-) thanks

jkneubuh (approved):
Welcome, Podman!

PR #597, opened 2022-01-26
Asset transfer events Go sample for Fabric Gateway

bestbeforetoday:


jkneubuh (approved):
This is a great sample:  clear, concise, straight to the point and focuses attention on the async submission and replay loop.   I definitely approve. 

One very, very minor suggestion (not worth holding up the PR), would be to re-organize the functions in the golang module / file such that the "gateway connection" bits are hidden down towards the bottom of the app.go, and the event replay / submission functions to bubble up right under `main()`.  That way a top-down linear scan has all of the relevant bits right in focus as soon as a user opens the routine.

jkneubuh:
A couple of follow-on discussion points, post merge of this PR, so as to not lose forever in the ephemeral discussion threads of slack... 

- The test-network is … is what it is. …    I’m wondering what it would take to start to incorporate some of these additional samples into the kube-test-network as an optional / supplemental / primary reference.   Some of this is the raw mechanics of building/packaging/running/etc. chaincode and gateway apps on k8s.  But some of this is also in the documentation trail.   If we can figure out how to incorporate multi-language-tab-selectors into the coding snippets, is this also an opportunity to introduce a tab selector for the target runtime?   (e.g… a tab for test-network, for test-net-k8s, for IBP/console, etc. …?)

- The second bit on the replay, was wondering if there was a natural way to incorporate the use of golang <-chan structures to handle some of the async event processing by example.   Maybe this is overkill, but when reading through the app.go I would have liked to see the async nature of the replay LOUD and CLEAR.   That said, I think a change would detract from the legibility of the narrative in the example, focusing more on channels and goroutines than the API presented in the event iterator.  IMO the app.go is in good shape as is.  Mostly pointing out that Future<T> , <-chan , and async coding is a larger topic than what can be reasonably addressed with a starter example.

bestbeforetoday:
> One very, very minor suggestion (not worth holding up the PR), would be to re-organize the functions in the golang module / file such that the "gateway connection" bits are hidden down towards the bottom of the app.go, and the event replay / submission functions to bubble up right under `main()`. That way a top-down linear scan has all of the relevant bits right in focus as soon as a user opens the routine.

This is a great suggestion, and something I considered doing in this PR, but I needed to get something delivered sooner rather than later. In the basic sample, those connection pieces are one of the focuses of the sample so it made sense to keep all the functions in top-to-bottom execution order. For this sample, those connection pieces are no longer the focus so it might even be a good idea to move them to a second support file to keep the main body of the sample as clear as possible.

bestbeforetoday:
> * The second bit on the replay, was wondering if there was a natural way to incorporate the use of golang <-chan structures to handle some of the async event processing by example.

It is already using Go channels! [ChaincodeEvents()](https://pkg.go.dev/github.com/hyperledger/fabric-gateway/pkg/client#Network.ChaincodeEvents) returns a channel, which is what is being ranged over in the event processing loops. And this is done entirely asynchronously in a goroutine for the real-time event listening. If you run the sample multiple times you will see console output interleaving in different ways.

PR #599, opened 2022-01-27
fix a bug in function [getBlockByNumber]

konoleoda:
This bug can lead to users can't change target peer to send query operations.


lehors (approved):
Nice! Thanks.

PR #601, opened 2022-01-27
Fix markdown logical structure in asset-transfer-events/README.md

bestbeforetoday:
Fenced code blocks within a list were at the wrong level if indentation and so logically terminated the list.


lehors (approved):
LGTM

PR #605, opened 2022-01-28
Run basic-asset-transfer on the Kubernetes Test Network in an Azure CI pipeline 

jkneubuh:
This PR adds a CI test case to run an E2E validation of the basic-asset-transfer chaincode and Gateway application on the Kubernetes Test Network.  This suite exercise only ONE test case, primarily as an exercise in validating the mechanics  of running the Kube Test Network on an Azure image.  After the mechanics of running on Azure have been worked out, additional tests can follow, both for additional Fabric routines (e.g. Commercial Paper) and new container orchestration runtimes (e.g. containerd / podman / nerdctl / etc.) 

The suite runs a complete E2E, including: 

- Creates a KIND Kubernetes cluster and Nginx ingress controller.

- Configures the Kubernetes Test Network (3 org: 2x peers/org + 3x orderers) using JFrog / STABLE tag Fabric images.

- Compiles and deploys asset-transfer-basic/chaincode-external Chaincode-as-a-Service.

- Creates a localhost:7051 -> service/org1-peer1:7051 port forward.

- Compiles and runs the application-gateway-typescript example on the host OS.

- Tears down the port-forward, Test Network, KIND cluster, and scrubs chaincode docker images.

### DIFF
--- a/asset-transfer-basic/application-gateway-typescript/src/app.ts
+++ b/asset-transfer-basic/application-gateway-typescript/src/app.ts
@@ -11,29 +11,35 @@ import { promises as fs } from 'fs';
 import * as path from 'path';
 import { TextDecoder } from 'util';
 
-const channelName = 'mychannel';
-const chaincodeName = 'basic';
-const mspId = 'Org1MSP';
+const channelName = envOrDefault('CHANNEL_NAME', 'mychannel');
+const chaincodeName = envOrDefault('CHAINCODE_NAME', 'basic');
+const mspId = envOrDefault('MSP_ID', 'Org1MSP');
 
 // Path to crypto materials.
-const cryptoPath = path.resolve(__dirname, '..', '..', '..', 'test-network', 'organizations', 'peerOrganizations', 'org1.example.com');
+const cryptoPath = envOrDefault('CRYPTO_PATH', path.resolve(__dirname, '..', '..', '..', 'test-network', 'organizations', 'peerOrganizations', 'org1.example.com'));
 
 // Path to user private key directory.
-const keyDirectoryPath = path.resolve(cryptoPath, 'users', 'User1@org1.example.com', 'msp', 'keystore');
+const keyDirectoryPath = envOrDefault('KEY_DIRECTORY_PATH', path.resolve(cryptoPath, 'users', 'User1@org1.example.com', 'msp', 'keystore'));
 
 // Path to user certificate.
-const certPath = path.resolve(cryptoPath, 'users', 'User1@org1.example.com', 'msp', 'signcerts', 'cert.pem');
+const certPath = envOrDefault('CERT_PATH', path.resolve(cryptoPath, 'users', 'User1@org1.example.com', 'msp', 'signcerts', 'cert.pem'));
 
 // Path to peer tls certificate.
-const tlsCertPath = path.resolve(cryptoPath, 'peers', 'peer0.org1.example.com', 'tls', 'ca.crt');
+const tlsCertPath = envOrDefault('TLS_CERT_PATH', path.resolve(cryptoPath, 'peers', 'peer0.org1.example.com', 'tls', 'ca.crt'));
 
 // Gateway peer endpoint.
-const peerEndpoint = 'localhost:7051';
+const peerEndpoint = envOrDefault('PEER_ENDPOINT', 'localhost:7051');
+
+// Gateway peer SSL host name override.
+const peerHostAlias = envOrDefault('PEER_HOST_ALIAS', 'peer0.org1.example.com');
 
 const utf8Decoder = new TextDecoder();
 const assetId = `asset${Date.now()}`;
 
 async function main(): Promise<void> {
+
+    await displayInputParameters();
+
     // The gRPC client connection should be shared by all Gateway connections to this endpoint.
     const client = await newGrpcConnection();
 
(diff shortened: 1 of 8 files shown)


jkneubuh:
cc @denyeart

jkneubuh:
This works really, really well.  

- The Azure pipeline setup is straightforward, and can be used to run some matrix tests when the opportunity arises.

- The Azure build image includes all of the dependencies (`kind`, `kubectl`, `docker`, etc.) necessary to bootstrap the network. 

- In current state, this PR will inflate the Kube test network, compile and build a chaincode image, deploy a smart contract, and compile+run a Typescript Gateway client application on the host OS. 

- The pre-merge pipeline fails (as it should) when the gateway client does not have access to the org1 admin certificates. 

- The application-gateway-typescript needs some work to correctly propagate a nonzero exit from `main()` when an exception is thrown.  @sapthasurendran and @bestbeforetoday can you please recommend a proper / node-idiomatic technique to make a controlled landing when / if things go wrong in the app? 

- The "correct" way to get to the gateway client should be to go through the nginx ingress with a proper host alias / etc.  That said, the client app is hardcoding the peer address as `localhost:7051`, so a cleaner patch will be to spin up a kubectl port-forward for the duration of the suite.

- Last bit of work remaining is to extract the Node and Admin certificates out into the `../test-network/organizations` folder structure.

bestbeforetoday:
Changing the line in the sample application that actually runs the `main()` function so that it sets the Node process exit code is a good approach:

```typescript
main().catch(error => {
    console.error('******** FAILED to run the application:', error);
    process.exitCode = 1;
});
```

bestbeforetoday:
Also, we could allow the hard-coded values within the sample to be overridden by environment variables, for when it's run in a non-default environment, like so:

```typescript
const peerEndpoint = process.env.ENDPOINT || 'localhost:7051';
```

asset-transfer-basic/application-gateway-typescript/src/app.ts
```diff
@@ -11,29 +11,44 @@ import { promises as fs } from 'fs';
...
+    console.log(`channelName:       ${channelName}`);
+    console.log(`chaincodeName:     ${chaincodeName}`);
+    console.log(`mspId:             ${mspId}`);
+    console.log(`cryptoPath:        ${cryptoPath}`);
+    console.log(`keyDirectoryPath:  ${keyDirectoryPath}`);
+    console.log(`certPath:          ${certPath}`);
+    console.log(`tlsCertPath:       ${tlsCertPath}`);
+    console.log(`peerEndpoint:      ${peerEndpoint}`);
+    console.log(`peerHostAlias:     ${peerHostAlias}`);
+
```
bestbeforetoday, line 51: It might be worth moving this to a separate function and calling it from main(). The intent was to keep main() a concise and readable set of high-level steps taken by the application, with the functions called the next level of implementation detail. This adds a lot of clutter to that concise, high-level overview of client API usage.

asset-transfer-basic/application-gateway-typescript/src/app.ts
```diff
@@ -205,3 +223,16 @@ async function updateNonExistentAsset(contract: Contract): Promise<void>{
...
+ */
+function envOrDefault(key: string, defaultValue: string): string {
+
+    const value = process.env[key];
+    if (! value) {
+        return defaultValue;
+    }
+
+    return value;
+}
```
bestbeforetoday, line 238: Logically this is fine. I'd probably expect a more concise form for Type/JavaScript code, as the behaviour of the `||` operator is well defined and it is frequently used:
```suggestion
function envOrDefault(key: string, defaultValue: string): string {
    return process.env[key] || defaultValue;
}
```

In general I like the separate named function to document the intent. The only downside with using `envOrDefault(key, defaultValue)` instead if directly using `process.env[key] || defaultValue` is that the default value is always calculated in order to invoke the function, whereas the inline approach avoids that calculation if the environment variable is present. For this example, that isn't really a concern though, so whichever approach you prefer.

bestbeforetoday (commented):
Just a couple of code comments to consider.

Does there need to be some README type documentation so people know how to run this, or is that coming in a later change?

ci/scripts/pullFabricImages.sh
```diff
@@ -1,15 +1,18 @@
...
 
+echo "Pulling images with $CONTAINER_CLI"
+
 for image in baseos peer orderer ca tools orderer ccenv javaenv nodeenv tools; do
-	docker pull -q "hyperledger-fabric.jfrog.io/fabric-${image}:${STABLE_TAG}"
-	docker tag "hyperledger-fabric.jfrog.io/fabric-${image}:${STABLE_TAG}" hyperledger/fabric-${image}
-	docker tag "hyperledger-fabric.jfrog.io/fabric-${image}:${STABLE_TAG}" "hyperledger/fabric-${image}:${FABRIC_VERSION}"
-	docker rmi -f "hyperledger-fabric.jfrog.io/fabric-${image}:${STABLE_TAG}"
+	$CONTAINER_CLI pull -q "hyperledger-fabric.jfrog.io/fabric-${image}:${STABLE_TAG}"
+	$CONTAINER_CLI tag "hyperledger-fabric.jfrog.io/fabric-${image}:${STABLE_TAG}" hyperledger/fabric-${image}
```
mbwhite, line 12: Particular reasons for pulling from JFrog and not DockerHub?

mbwhite (commented):
wondering why jfrog and not dockerhub?

jkneubuh:
Abandoning this PR
The new one (PR #637) is much better
No merge conflicts

PR #607, opened 2022-01-31
Minor improvement to test network docs

davidkel:
Also fix the version of prometheus and grafana being used
to ensure the queries remain compatible as well as avoid
issues whether they may not work if latest images already
exist on machine but are not at a level that will work

Signed-off-by: D <d_kelsey@uk.ibm.com>


lehors (dismissed):
LGTM but I spotted a few typos you might as well fix.

davidkel:
Many thanks for the review @lehors , I've hopefully corrected those typos now

mbwhite (approved):

PR #609, opened 2022-02-01
Update test-network to use org level ca-cert.pem

denyeart:
Although test network scripts functionally worked, they demonstrated
some strange patterns around use of the ca cert.

This change makes it clear that the ca cert is configured and used
at the org level, and not related to any individual peer or orderer node.
Also that when connecting to individual servers, clients can pass the
org level CA cert as the root of trust when establishing TLS connections.


mbwhite (dismissed):

jkneubuh:
@denyeart - good job with this PR.  You must have a bit of a headache after reviewing all the cert usage in context.  👍

mbwhite (approved):

PR #620, opened 2022-02-02
Run Go Gateway events sample in CI

bestbeforetoday:


jkneubuh (approved):
Looks good.  

I can see the new test in the CI suite.  
I can see the run in the Azure pipeline.  
I can see a 0 exit code from the gateway app / output log.

I see a merge following shortly.   Looks good. 👍

PR #626, opened 2022-02-03
Update samples to use published fabric-rest-sample

jt-nti:

### DIFF
--- a/asset-transfer-basic/rest-api-typescript/README.md
+++ b/asset-transfer-basic/rest-api-typescript/README.md
@@ -132,13 +132,9 @@ npm run start:dev
 
 ### Docker image
 
-Alternatively, run the following commands in the `fabric-rest-sample/asset-transfer-basic/rest-api-typescript` directory to start the sample in a Docker container
+It's also possible to use the [published docker image](https://github.com/hyperledger/fabric-samples/pkgs/container/fabric-rest-sample) to run the sample
 
-Build the Docker image
-
-```shell
-docker build -t fabric-rest-sample .
-```
+Clone the `fabric-samples` repository and change to the `fabric-samples/asset-transfer-basic/rest-api-typescript` directory before running the following commands
 
 Create a `.env` file to configure the server for the test network (make sure `TEST_NETWORK_HOME` is set to the fully qualified `test-network` directory and `AS_LOCAL_HOST` is set to `false` so that the server works inside the Docker Compose network)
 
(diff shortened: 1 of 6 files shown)


jkneubuh:
@jt-nti what actually does the heavy lifting of pushing the REST docker image up to ghcr.io ?  Last bit - remind me to add, or add an E2E test for this once we land the Azure CI suite at [ci/scripts/run-k8s-test-network-basic.sh](https://github.com/jkneubuh/fabric-samples/blob/feature/k8s-asset-transfer-basic-test/ci/scripts/run-k8s-test-network-basic.sh#L118) (This is still in a PR off in the wings.)

Something along the lines of : 
```
createNetwork
print "Running REST Sample Application" 
export export SAMPLE_APIKEY=97834158-3224-4CE7-95F9-A148C886653E
( ./network rest-easy \
  && curl -s --header "X-Api-Key: ${SAMPLE_APIKEY}" http://localhost/api/assets \
  && curl ... )
stopNetwork 
```

If / when we get the ghcr.io and GHA sorted out, we should add a proper tag to the image loaded by the kube descriptor.  Currently this is `image: ghcr.io/hyperledger/fabric-rest-sample`, which will test whatever was the latest published REST container. I think this needs to be tied in to test whatever was the target of *this* build.

jkneubuh:
@jt-nti this looks good.  Before hitting the OK button would like to call out that it's not "correct" to apply the latest label during the Azure pipeline CI step and then reference the `latest` tag from the k8s descriptors.   It will run a valid test but it does mean that if there's a regression it will check in a broken image at the `latest` line and linger until the line is patched up.

What if we referenced a build tag, (e.g. `git describe`) or something consistent with the build number passed into the Azure scope?   We could include the formal "latest" label and defer to the release practice, or possibly apply in the CI pipeline only after a successful CI build.   This would need to have an env parameter introduced into the kube descriptors, to make sure the image being tested was the same rev as what was generated by the azure build.

Or leave "as is" and we can sort this out with a move to GitHub actions. There's a lot of good work that can be done on the CI pipeline.   Are you OK with this line publishing a "broken" image up to the latest tag?

jt-nti:
@jkneubuh hopefully it's unlikely a broken image will get published- the push happens after running the REST sample tests, and only on a merge build, so all the fabric-sample tests will have passed in the PR build.

jkneubuh:
Got it.  Thanks for sorting me out James. Looks good I'll land the PR.

jkneubuh (approved):
Looks good.  

As a future enhancement I'd like us to sort out a proper strategy to publish each of the samples independently up to a public container registry.  Using the `latest` tag is convenient, but can be improved by pinning everything to specific builds of the Fabric components and samples.

PR #627, opened 2022-02-03
sync ports for operations endpoint with main

varadgit:
Ports for the operations endpoint seems different between the main branch and the 2.2 branch... need to sync so that consumer code can reference is consistently between releases.

Signed-off-by: Varad Ramamoorthy <varad@us.ibm.com>

### DIFF
--- a/test-network/docker/docker-compose-test-net.yaml
+++ b/test-network/docker/docker-compose-test-net.yaml
@@ -27,7 +27,7 @@ services:
       - ORDERER_GENERAL_GENESISFILE=/var/hyperledger/orderer/orderer.genesis.block
       - ORDERER_GENERAL_LOCALMSPID=OrdererMSP
       - ORDERER_GENERAL_LOCALMSPDIR=/var/hyperledger/orderer/msp
-      - ORDERER_OPERATIONS_LISTENADDRESS=0.0.0.0:17050
+      - ORDERER_OPERATIONS_LISTENADDRESS=0.0.0.0:9443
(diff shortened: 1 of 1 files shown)


jkneubuh (changes requested):
Hi varad.  I think this is an OK update but a couple of questions.

1.  Check the references in the `ports:` stanza.  These still reference { 17050, 17051, and 19051 } and I _believe_ need to be updated to 9443, etc. 

2.  Why is the 2.2 branch referencing the 0.0.0.0 NIC and the mainline hitting the docker network for peer0.org2.example.com?   I don't think there's any need to unwind this mystery, it's just mysterious that some of the interfaces are on 0.0.0.0 and others are on the virtual network.  

3.  This is not a question, but more a comment....  `<RANT>`I _despise_ the port mapping and networking fiddly bits in the compose test network.  It's ... absolute ... craziness.  Can we please start to use and build out the Kube test network and figure out good solutions for the ingress routes into Nginx, or just "leave in kube" and reference via `Service` DNS hostnames?   The Kube stuff is a new pattern but the networking stack - just - makes sense. `</RANT>`  (thanks.)

varadgit:
oops! thanks @jkneubuh for catching the port mapping! fixed that...

I fixed the network address binding for the operations endpoint to match what's in the main branch as well so that it is consistent.

jkneubuh:
better.  Will give this a 👍 👍 and docker-compose 👎 : )

jkneubuh (approved):

PR #628, opened 2022-02-03
Java gateway application sample for asset-transfer-events

bestbeforetoday:


asset-transfer-events/application-gateway-java/src/main/java/App.java
```diff
@@ -0,0 +1,162 @@
...
+				.startBlock(startBlock)
+				.build();
+
+		try (CloseableIterator<ChaincodeEvent> eventIter = request.getEvents()) {
+			while (eventIter.hasNext()) {
+				ChaincodeEvent event = eventIter.next();
+				String payload = prettyJson(event.getPayload());
+				System.out.println("\n<-- Chaincode event replayed: " + event.getEventName() + " - " + payload);
+
+				if (event.getEventName().equals("DeleteAsset")) {
```
mbwhite, line 156: could I suggest adding a comment about the importance of the `break`
So it's clear this will keep listening for events indefinately

asset-transfer-events/application-gateway-go/app.go
```diff
@@ -66,7 +66,7 @@ func main() {
 }
 
 func startChaincodeEventListening(ctx context.Context, network *client.Network) {
```
mbwhite, line 68: Should this `app.go` file be in this PR?

asset-transfer-events/application-gateway-typescript/src/app.ts
```diff
@@ -134,8 +134,9 @@ async function deleteAssetByID(contract: Contract): Promise<void>{
     console.log('\n*** DeleteAsset committed successfully');
```
mbwhite, line 134: likewise this typescript file?

asset-transfer-events/application-gateway-go/app.go
bestbeforetoday: Yes, absolutely. The sample is not useful if it doesn't contain a sample application.

asset-transfer-events/application-gateway-typescript/src/app.ts
bestbeforetoday: Yes, again This is the sample application.

asset-transfer-events/application-gateway-go/app.go
bestbeforetoday: You asked me to update the comments around stopping event listening

mbwhite (approved):

PR #630, opened 2022-02-04
DRAFT:Secured agreement samples using gateway

sapthasurendran:
Signed-off-by: sapthasurendran <saptha.surendran@ibm.com>


asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,427 @@
...
+        await changePublicDescriptionAfterTransfer(assetKey,mspIdOrg2,`Asset ${assetKey} owned by ${mspIdOrg2} is NOT for sale`,contractOrg2);
+
+        // Read the public details by  org1.
+        await readAsset(assetKey, mspIdOrg2, contractOrg1,mspIdOrg1);
+
+        // Read the public details by  org2.
+        await readAsset(assetKey, mspIdOrg2, contractOrg2,mspIdOrg2);
+
+    } finally {
+        gatewayOrg1.close();
```
mbwhite, line 168: can the `close()` itself fail?  Might be an idea to wrap each in a `try {}` such that even if one fails, the the others might still be able to clean up properly

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,427 @@
...
+    const resultString = utf8Decoder.decode(resultBytes);
+    const result = JSON.parse(resultString);
+
+    if (result) {
+        if (result.ownerOrg === ownerOrg) {
+            console.log(`*** Result from ${org} - asset ${result.assetID} owned by ${result.ownerOrg} DESC:${result.publicDescription}`);
+        } else {
+            console.log(`${RED}*** Failed owner check from ${org} - asset ${result.assetID} owned by ${result.ownerOrg} DESC:${result.publicDescription}${RESET}`);
+        }
+    }
```
mbwhite, line 212: ... if result is null?

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,427 @@
...
+        }
+        const resultBytes = await contract.evaluateTransaction('GetAssetPrivateProperties', assetKey);
+
+        const resultString = utf8Decoder.decode(resultBytes);
+        const result = JSON.parse(resultString);
+
+        console.log('*** Result:', result);
+    }
+    catch(e){
+        console.log(`${RED}*** Failed evaluateTransaction readPrivateAsset: ${e}${RESET}`);
```
mbwhite, line 229: Should the error be sent back to the application here?  As far as the app is concerned all the transactions will execute successfully?

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,427 @@
...
+}
+
+async function transferAsset(assetKey:string, org:string,buyerOrgID:string, price:number,contract:Contract):Promise<void>{
+    try{
+        const asset_properties = {
+            object_type: 'asset_properties',
+            asset_id: assetKey,
+            color: 'blue',
+            size: 35,
+            salt: Buffer.from(now.toString()).toString('hex')
```
mbwhite, line 365: accessing current time is usually problematic - as the time will change when different peers execute the transaction. Therefore this has great potential to fail... 

Unless this is only executed on one peer, or there are other mitigating factors... which if they are are worth noting in comments.

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
sapthasurendran: The transaction is expected to fail .This is to demonstrate access control as org2 is not the owner and does not have the private details.

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
bestbeforetoday: The TypeScript implementation of `close()` currently does nothing. It might do something in the future but we don't expect `close()` to fail

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
bestbeforetoday: As Matthew points out, you probably want a better check. Something like:
```typescript
if (result?.ownerOrg === ownerOrg) {
    console.log(`*** Result from ${org} - asset ${result.assetID} owned by ${result.ownerOrg} DESC:${result.publicDescription}`);
} else {
    console.log(`${RED}*** Failed owner check from ${org} - asset ${result.assetID} owned by ${result.ownerOrg} DESC:${result.publicDescription}${RESET}`);
}
```
You don't need the surrounding `if (result) {}`.

Although there is another code smell here, which is that you're dealing with `any` type values. I wonder if it would be worth defining an interface that describes the shape of the objects you expect back from `JSON.parse()`?

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
sapthasurendran: @mbwhite Can you help me understand the significance of passing the `salt` parameter here?

The current time is being accessed at the client side and stored as a constant value now which is  being passed to the transactions .

bestbeforetoday: This is generated in the client, not the chaincode. So there is only a single value per transaction proposal, regardless of how many peers endorse or when that happens.

Since it seems to be intended to prevent the message content being guessable from the hash of the data on the main ledger, perhaps it would be better to use something like [crypto.randomBytes()](https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto_randombytes_size_callback) or [crypto.randomUUID()](https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto_randomuuid_options) rather than current date, which is much less cryptographically secure as a salt.

sapthasurendran:
Made the changes as suggested. Please have a look.

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,434 @@
...
+
+const channelName = 'mychannel';
+const chaincodeName = 'secured';
+
+//Use a random key so that we can run multiple times
+const now = Date.now();
+const assetKey= `asset${now}`;
+
+//Generate random bytes using crypto
+const randomBytes = crpto.randomBytes(256).toString('hex')
```
bestbeforetoday, line 26: Using the same salt for every transaction that requires salt is not really best-practice. You could just make this a function and call it (as `randomBytes()`) for a new salt every time you need one.
```suggestion
function randomBytes(): string {
    return crpto.randomBytes(256).toString('hex');
}
```

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,434 @@
...
+
+async function readAsset(assetKey:string, ownerOrg:string, contract:Contract,org:string):Promise<void> {
+    console.log(`${GREEN}--> Evaluate Transactions: ReadAsset as ${org}, - ${assetKey} should be owned by ${ownerOrg}${RESET}`);
+
+    const resultBytes = await contract.evaluateTransaction('ReadAsset', assetKey);
+
+    const resultString = utf8Decoder.decode(resultBytes);
+    if (resultString.length !== 0) {
+
+        const result:{ objectType: string, assetID: string, ownerOrg: string, publicDescription: string } = JSON.parse(resultString);
```
bestbeforetoday, line 211: It might be more readable to declare an interface for the asset at the module scope rather than include the typing inline, something like:
```typescript
interface Asset {
    objectType: string,
    assetID: string,
    ownerOrg: string,
    publicDescription: string,
}
```

And alongside that, perhaps extract the parsing of an asset to a named function and use that for every method that's parsing an asset.
```typescript
function parseAsset(json: string): Asset {
    return JSON.parse(json);
}
```
You could do some validation within the parseAsset() function that the output contains the expected fields but for the sample I'd be tempted to skip that to keep the code concise.

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
```diff
@@ -0,0 +1,434 @@
...
+            asset_id: assetKey,
+            color: 'blue',
+            size: 35,
+            salt: randomBytes
+        };
+        const asset_price = {
+            asset_id: assetKey,
+            price:price,
+            trade_id: now.toString()
+        };
```
bestbeforetoday, line 378: Since asset properties and price get used multiple times throughout the file, it would be good practice to declare a module-scoped interface for **AssetProperties** and **AssetPrice**, and use them throughout the application so the compiler can help ensure your data is consistent.

bestbeforetoday (changes requested):
This really needs a `README.md` in the `asset-transfer-secured-agreement` directory giving a brief description of the sample and the commands to run it, similar to the basic and events asset-transfer samples. Just running the network and chaincode deployment commands from the [tutorial in the main Fabric docs](https://hyperledger-fabric.readthedocs.io/en/latest/secured_asset_transfer/secured_private_asset_transfer_tutorial.html), running the client application fails with this error:
```
******** FAILED to run the application: [Error: ENOENT: no such file or directory, open '/Users/mark/src/fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/User1@org1.example.com/msp/signcerts/cert.pem'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/mark/src/fabric-samples/test-network/organizations/peerOrganizations/org1.example.com/users/User1@org1.example.com/msp/signcerts/cert.pem'
}
```

This is because the tutorial doesn't use CAs to create the network identities. Just having the network set up command including the CAs in a README would solve this.

bestbeforetoday:
This sample is not showing anything new in the Fabric Gateway client API and it also has a fair bit more complicated application flow than some of the more basic samples. Maybe this is a point to consider demonstrating some good practices in real-world client application design. If so, I would be tempted to:

1. Define classes for complex data types, like Asset, AssetProperties and AssetPrice, with an instance method to serialize the object and a static factory method to construct an instance from serialized data. Or the data types could simply be interfaces and then have some "serializers" that know how to convert these interfaces to/from serialized form.
2. Define a class that represents the smart contract (wrapping an instance of the client API's Contract object), with methods on it that mirror the smart contract transaction functions. Use the complex data types as parameters or return values as appropriate. These methods would deal with de/serializing data objects and invoking the client API appropriately.

Now your main application flow would read as a set of business operations using business-level data objects. What do you think?

jkneubuh:
Hi @sapthasurendran and @bestbeforetoday - 

Are there still open dev items or work remaining for this PR?   This has been lingering for a while - shall we close it, or push it over the end-line into the samples repo?

bestbeforetoday:
It would be good to address the comments from the last review and then get this in in its current form. If we want to refactor later that can happen in another PR.

asset-transfer-secured-agreement/application-gateway-typescript/src/app.ts
sapthasurendran: This actually failed the verify asset properties function as it is comparing the hash stored and the data passed.

asset-transfer-secured-agreement/application-gateway-typescript/src/contractWrapper.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright IBM Corp. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { Contract, ProposalOptions } from '@hyperledger/fabric-gateway';
+import { TextDecoder } from 'util';
+
+export class ContractWrapper {
```
bestbeforetoday, line 9: Right now, this class is adding so little to the Fabric Gateway client API, I suspect it would be better not to have the extra code to look through. When talking about providing a more business-level abstraction of the smart contract at the client end, I was thinking something like this:

https://github.com/hyperledger/fabric-samples/blob/05791d30bc68fee8f9408e34bc1d5cee68d842f9/off_chain_data/application-typescript/src/contract.ts

So having a class with methods on it that correspond to the smart contract transaction functions (or combinations of transaction functions) used by the client application, but which hide the mechanics of invoking those transactions, and un/marshalling both the arguments and responses.

With that in place, the main body of the application _might_ work as a flow of calls to that contract abstraction and not require all the the specific methods (like createAsset, readAsset etc.) in the app.ts code itself.

asset-transfer-secured-agreement/README.md
```diff
@@ -0,0 +1,64 @@
...
+   ```
+   # To run the Typescript sample application
+   cd application-gateway-typescript
+   npm install
+   npm start
+
+   # To run the Javascript sample application
+   cd application-gateway-javascript
+   node app.js
+   ```
```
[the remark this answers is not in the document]
sapthasurendran: Sorry , I meant application-javascript which does exist.

asset-transfer-secured-agreement/application-gateway-typescript/src/contractWrapper.ts
```diff
@@ -0,0 +1,264 @@
...
+import { Asset, AssetJSON, AssetPrice, AssetPriceJSON, AssetPrivateData, AssetProperties, AssetPropertiesJSON, GREEN, parse, RED, RESET } from './utils';
+
+export class ContractWrapper {
+
+    readonly #contract: Contract;
+    readonly #org: string;
+    readonly #utf8Decoder  = new TextDecoder();
+    readonly #randomBytes: string;
+
+    public constructor(contract: Contract, org: string, randomBytes: string) {
```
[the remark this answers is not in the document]
bestbeforetoday: `org` looks just to be used for log messages and tailor them based on whether it matches owner org arguments. For a sample that's probably fine. Generally I would think of this type of class being just an abstraction layer to the mechanics of transaction invocation, and that level of logging detail living outside this class in the application business logic, but for this purpose whatever you prefer is fine by me.

asset-transfer-secured-agreement/application-gateway-typescript/src/contractWrapper.ts
```diff
@@ -0,0 +1,264 @@
...
+
+            await this.#contract.submit('CreateAsset', {
+                arguments: [asset.AssetId, asset.PublicDescription],
+                transientData: { asset_properties: JSON.stringify(assetPropertiesJSON)},
+            });
+
+            console.log(`*** Result: committed, asset ${asset.AssetId} is owned by Org1`);
+        }catch (e) {
+            console.log(`${RED}*** Failed: createAsset - ${e}${RESET}`);
+        }
```
bestbeforetoday, line 40: I don't really like this pattern of catching, logging and then ignoring for all transaction errors. It makes it hard to see if the application is actually working correctly because it never actually fails. I would be tempted to let these methods throw if the transaction invocation fails, then try/catch around the call in the app main() function for only the invocation that are expected to fail because they are demonstrating a failure condition.

asset-transfer-secured-agreement/application-gateway-typescript/src/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    AssetId: string;
+    Color: string;
+    Size: number;
+}
+
+export interface AssetPrice {
+    AssetId: string;
+    Price: number;
+    TradeId: string;
+}
```
bestbeforetoday, line 54: I really like the separation of JSON interfaces to define the shape of JSON data sent to/from transactions, and then these interfaces for the application-level view of data sent to/from the ContractWrapper class. Two things that jump out at me though:

1. The application-level interfaces use capitalised property names. The normal convention for Java/TypeScript applications is lowercase property names, so I expected to see that here.
2. These interfaces are closely related to the ContractWrapper class so they might fit better in the contractWrapper module.

asset-transfer-secured-agreement/application-gateway-typescript/src/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    color: string;
+    size: number;
+    salt: string;
+}
+
+export interface AssetPriceJSON {
+    assetID: string;
+    price: number;
+    tradeID: string;
+}
```
bestbeforetoday, line 31: These JSON interfaces are only used internally by the ContractWrapper class. No other code uses or needs to know about them. They should be non-exported members of the contractWrapper module.

asset-transfer-secured-agreement/application-gateway-typescript/src/contractWrapper.ts
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    public async createAsset(asset: Asset, privateData: AssetPrivateData): Promise<void> {
+        console.log(`${GREEN}--> Submit Transaction: CreateAsset, ${asset.AssetId} as ${asset.OwnerOrg} - endorsed by Org1${RESET}`);
+        try {
+            const assetPropertiesJSON: AssetPropertiesJSON = Object.assign({}, {objectType: 'asset_properties',
+                assetID: asset.AssetId,
+                color: privateData.Color,
+                size: privateData.Size,
+                salt: this.#randomBytes });
```
bestbeforetoday, line 30: Object.assign() is only really useful here if you have an object that is a subset of (or at least has only property names that match) the object you want to produce, and that you want to augment with additional properties (or modify existing properties). Or as a shallow copy of an object. Here it's not doing anything useful.
```suggestion
            const assetPropertiesJSON: AssetPropertiesJSON = {
                objectType: 'asset_properties',
                assetID: asset.AssetId,
                color: privateData.Color,
                size: privateData.Size,
                salt: this.#randomBytes,
            });
```

bestbeforetoday (commented):
The code in the app module looks much clearer to me now. Definite improvement.

asset-transfer-secured-agreement/application-gateway-typescript/src/contractWrapper.ts
bestbeforetoday: What you have now looks much better. I can't seem to mark this resolved so just a note to ignore it now.

asset-transfer-secured-agreement/application-gateway-typescript/src/contractWrapper.ts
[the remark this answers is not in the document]
sapthasurendran: > `org` looks just to be used for log messages and tailor them based on whether it matches owner org arguments. For a sample that's probably fine. Generally I would think of this type of class being just an abstraction layer to the mechanics of transaction invocation, and that level of logging detail living outside this class in the application business logic, but for this purpose whatever you prefer is fine by me.

Ya I agree. I was thinking better not to clutter main function with many logs.

bestbeforetoday (approved):
Looks good to me! 👍

mbwhite (approved):

PR #634, opened 2022-02-08
Make network.sh more tolerant

mbwhite:
The network.sh is dependent on being executed from the test-network directory.

This PR makes it a little more tolerant, and will add a bin directory based on the location of the script - not just the PWD


lehors (approved):
That's nice. It worked for me.

PR #649, opened 2022-02-17
fix docker compose version 2 -> 2.1

kaushikkumarbora:
Earlier compose version was giving a property error when running the [fabric-samples/test-network/network.sh](https://github.com/hyperledger/fabric-samples/blob/main/test-network/network.sh) script.
compose_error

Docker Compose Documentation says that the property [networks:name](https://docs.docker.com/compose/compose-file/compose-file-v2/#name-1) was added in [v2.1](https://docs.docker.com/compose/compose-file/compose-versioning/#version-21) file format.
Signed-off-by: kaushikkumarbora <kaushikkumarbora@gmail.com>

### DIFF
--- a/test-network/addOrg3/docker/docker-compose-ca-org3.yaml
+++ b/test-network/addOrg3/docker/docker-compose-ca-org3.yaml
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-version: '2'
+version:'2.1'
 
 networks:
   test:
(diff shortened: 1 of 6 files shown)


lehors (dismissed):
LGTM. Thanks.

kaushikkumarbora:
Missed some spaces. Should be fine now. Sorry for my mistake.

lehors:
Although the change seems all right I haven't been able to reproduce the problem. Which version of docker-compose gave you this error?

kaushikkumarbora:
docker-compose version `1.25.0` gave this error.

lehors:
ok, I have version 1.29.2 which is apparently more lenient but I don't see any harm anyway.
Thanks.

lehors (approved):
LGTM

Issue #504, opened 2021-10-20
test-network-nano-bash still requires docker


davidkel:
test-network-nano-bash is a great way to run fabric on native platforms (mac/linux), but it still requires docker in order to run chaincode.

it would be great if that docker dependency could be removed completely by including a simple external chaincode builder and launcher that supports building chaincode as well as chaincode running as a server

PR #650, opened 2022-02-17
Add external chaincode service for Nano Test Network

lehors:
This addresses Issue #504. This extends the test-network-nano-bash
sample to offer an option to run a fabric network without using any
containers at all. The chaincode is running as an external service
directly on the host machine.

### DIFF
--- a/test-network-nano-bash/README.md
+++ b/test-network-nano-bash/README.md
@@ -2,7 +2,7 @@
 
 Test network Nano bash provides a set of minimal bash scripts to run a Fabric network on your local machine.
 The network is functionally equivalent to the docker-based Test Network, you can therefore run all the tutorials and samples that target the Test Network with minimal changes.
-The Fabric release binaries are utilized rather than using docker containers to avoid all unnecessary layers. Only the chaincode and chaincode builder runs in a docker container behind the scenes.
+The Fabric release binaries are utilized rather than using docker containers to avoid all unnecessary layers. And you can choose between running the chaincode and chaincode builder in a docker container behind the scenes or running the chaincode as a service without any containers at all.
 Using the Fabric binaries also makes it simple for Fabric developers to iteratively and quickly modify Fabric code and test a Fabric network as a user.
 
 As the name `nano` implies, the scripts provide the smallest minimal setup possible for a Fabric network while still offering a multi-node TLS-enabled network:
@@ -17,6 +17,20 @@ As the name `nano` implies, the scripts provide the smallest minimal setup possi
 
 - Follow the Fabric documentation for the [Prereqs](https://hyperledger-fabric.readthedocs.io/en/latest/prereqs.html)
 - Follow the Fabric documentation for [downloading the Fabric samples and binaries](https://hyperledger-fabric.readthedocs.io/en/latest/install.html). You can skip the docker image downloads by using `curl -sSL https://bit.ly/2ysbOFE | bash -s -- -d`
+
+## To run the chaincode as a service
+- You need to have the `ccaas_builder` binaries. If you do not have them in `fabric-samples/bin` you can build them from the Fabric source with the command `make ccaasbuilder`, you will then find the builder in `fabric/release/darwin-amd64/bin` or equivalent for your system. Just move the whole hierarchy starting there to `fabric-samples/bin` with something like: `mv release/darwin-amd64/bin/ccaas_builder ../fabric-samples/bin`
+- You need to edit the `fabric-samples/config/core.yaml` file to point to that builder. The path specified in the default config file is only valid within the peer container which you won't be using. Modify the `externalBuilders` field in the `core.yaml` file to add the local external builder so that the configuration looks something like the following:
+```
+externalBuilders:
+       - name: ccaas_builder
+         path: /opt/hyperledger/ccaas_builder
+         propagateEnvironment:
+           - CHAINCODE_AS_A_SERVICE_BUILDER_CONFIG
+       - name: external-sample-builder
+	     path: ../bin/ccaas_builder
+```
+The path must be absolute or relative to where the peer will run so that it can find the builder when installing the chaincode.
 
 # Instructions for starting network
 
@@ -42,7 +56,12 @@ The remaining peer admin scripts join their respective peers to `mychannel`.
 
 # Instructions for deploying and running the basic asset transfer sample chaincode
 
-To deploy and invoke the chaincode, utilize the peer1 admin terminal that you have created in the prior steps.
+To deploy and invoke the chaincode, utilize the peer1 admin terminal that you have created in the prior steps. You have two possibilities:
+
+1. Using a chaincode container
+2. Running the chaincode as a service
+
+## 1. Using a chaincode container
 
 Package and install the chaincode on peer1:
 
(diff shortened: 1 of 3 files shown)


mbwhite:
@lehors thanks for doing this... I've been meaning to do this for *ages* 
I'll add some comments to the review

asset-transfer-basic/chaincode-external-nano/README.md
```diff
@@ -0,0 +1,78 @@
+# Asset-Transfer-Basic as an external service - Nano
+
+This sample complements the [Test Network Nano sample](../../test-network-nano-bash) demonstrating how you can run the chaincode as an external service without using any container at all.
+
+## Setting up the external builder and launcher
+
+Fabric's external builders and launchers is an advanced feature that typically requires custom packaging of the peer image so that it contains all the tools your builder and launcher require. This sample uses very simple (and crude) shell scripts that can be run directly by the default Fabric peer.
```
mbwhite, line 7: The latest `core.yaml` and the latest 2.4 binaries have a preconfigured external builder for 'chaincode-as-a-service' these are go applications - so don't need to use 'the crude shell scripts'

Though ironically the scripts used are mentioned in the docs :-)

asset-transfer-basic/chaincode-external-nano/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Open the `config/core.yaml` file at the top of the `fabric-samples` directory. If you do not have this file, follow the instructions to [Install the Samples, Binaries and Docker Images](https://hyperledger-fabric.readthedocs.io/en/latest/install.html) to download the Fabric binaries and configuration files alongside the Fabric samples.
+
+Modify the `externalBuilders` field in the `core.yaml` file to add the external sample builder which comes with this sample so that the configuration looks something like the following:
+```
+externalBuilders:
+    - name: external-sample-builder
```
mbwhite, line 14: _cf_ comment re the current core.yaml being pre-configured

asset-transfer-basic/chaincode-external-nano/README.md
```diff
@@ -0,0 +1,78 @@
...
+```
+export CC_PACKAGE_ID=basic_1.0:c70ccad6efd2d7ccb4f68816b62eb038a01dacae3ebcc90b23edab27c2db127d
+```
+
+## Starting the chaincode service
+
+Back in your chaincode service terminal, you now need to build the chaincode with the following command:
+```
+go build
+```
```
mbwhite, line 62: The `asset-transfer-basic` Java and Node chaincodes are already setup to be able to run in the required 'mode' 
I've got a change to make to go to allow it to run in a simpler manner. 

But that change is stacked up behind other things...

asset-transfer-basic/chaincode-external-nano/connection.json
```diff
@@ -0,0 +1,5 @@
+{
```
mbwhite, line 1: The prepackaged external builders have a richer semantics here, and can template the address etc. 

as an FYI - a lot of the external builder scripts are wrong with their handling of TLS.

asset-transfer-basic/chaincode-external-nano/README.md
lehors: Yes, I agree with that. I think it'd be nice to be able to use the "standard" chaincode samples instead of this custom one. I tried and I couldn't seem to provide it with the right environment to make it happy. Maybe we could have a chat about it.

asset-transfer-basic/chaincode-external-nano/README.md
lehors: Yup, I'm well aware of the ccaas builder. I actually wanted to use it but that's not a binary that we make available independently and that people will have on their machine so I decided against using it.
If we do want to use it we will have to instruct people on how to build it first.

asset-transfer-basic/chaincode-external-nano/connection.json
lehors: Yes, as a matter of fact this PR doesn't really provide for TLS support.

mbwhite: yes - and that's perfectly fine for this example... TLS would just make it too complex

asset-transfer-basic/chaincode-external-nano/README.md
mbwhite: The ccaas builder is in the same tgz file as the other binaries... so if the install script is used to pull those - you can the ccaas for *free*

lehors: I just downloaded the binaries for 2.4.2 and it's not there.

lehors: Well, even on jfrog it's only included in the linux distro. 👎

asset-transfer-basic/chaincode-external-nano/README.md
jkneubuh: This works but is not ideal...   hello, `fabric-cli`, where are you?? 

https://github.com/hyperledger/fabric-samples/blob/main/test-network-k8s/scripts/chaincode.sh#L153

```
function set_chaincode_id() {
  local cc_package=build/chaincode/${CHAINCODE_NAME}.tgz
  cc_sha256=$(shasum -a 256 ${cc_package} | tr -s ' ' | cut -d ' ' -f 1)

  label=$( jq -r '.label' chaincode/${CHAINCODE_NAME}/metadata.json)

  CHAINCODE_ID=${label}:${cc_sha256}
}
```

lehors:
So, I took another crack at this, following a minimalist approach. Let me know what you guys think.

asset-transfer-basic/chaincode-external-nano/README.md
lehors: So, I switched to using ccaasbuilder adding instructions to build it for the poor souls that won't find it from installing the binaries. We can eventually remove those instructions.

asset-transfer-basic/chaincode-external-nano/README.md
lehors: You still need to modify the config file though because the path specified by default is only valid within the container. I don't think instructing people to install the ccaasbuilder in /opt/hyperledger/ccaas_builder on their machine is really an option.

asset-transfer-basic/chaincode-external-nano/README.md
lehors: So, I must say I'm rather confused by the fact that it seems that different versions of the chaincode use different environment variables. The Java version wants CORE_CHAINCODE_ID_NAME, while the Typescript version wants CHAINCODE_ID. What gives? Is it just because we are in transition and not all versions are in sync?

asset-transfer-basic/chaincode-external-nano/README.md
mbwhite: good point!

asset-transfer-basic/chaincode-external-nano/README.md
mbwhite: Wouldn't say we in 'transition' more like everything was written by different people originally. 

We need to get this sorted.

mbwhite (approved):
LGTM

asset-transfer-basic/chaincode-external-nano/README.md
lehors: If we could agree on one way to go I'll be happy to help with transitioning everything in that direction.

PR #678, opened 2022-03-16
Override default port of CA connection on CCP teample

charalarg:
**Signed-off-by**: Charalarg [charalarg@gmail.com](mailto:charalarg@gmail.com)

**Type of change**:
- Bug fix

**Description**:

In test-network-k8s the certificate authorities are deployed with TLS enabled under the port 443. But the connection profile template implements a connection to the default (7054) port of the certificate authority.

### DIFF
--- a/test-network-k8s/scripts/ccp-template.json
+++ b/test-network-k8s/scripts/ccp-template.json
@@ -36,7 +36,7 @@
     },
     "certificateAuthorities": {
         "org${ORG}-ca": {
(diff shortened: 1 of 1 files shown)


jkneubuh (approved):
Hi @charalarg - this looks like a good update.  I don't believe any services were consuming this attribute, so it's quite possible the connection was slipping through the cracks.   

7054 is an odd choice of a default port for an HTTPS connection.  Thanks for tracking this down.

PR #684, opened 2022-03-23
Provide clear guidance for debugging Java chaincode as a service

jkneubuh:
**READY FOR MERGE** 

This PR addresses the need for a simple, clear example that illustrates the use of Chaincode-as-a-Service to build and run Java chaincode on a local development environment.

In addition, the README includes a section that describes how users may employ a Docker loopback interface to connect from a remote Peer binary to a local port on the host OS.  In this manner, chaincode may be run locally while attached to a debugger, with the peer running on a remote pod in Kubernetes.

In the first iteration of this PR, we introduced a new `ccpackage/*` folder to hold descriptive files that would be used to generate the chaincode-as-a-service package, AND help with the deployment to k8s: 

- ccpackage/metadata.json     (defines the package label)
- ccpackage/connection.json  (defines the CC service URL for gRPC access by the peer) 
- ccpackage/ccaas.json           (describes the CC name, used to construct the CC pods in k8s, and CC IMAGE) 

After community review, the PR has been reflected to *remove* the /ccpackage/ folder, opting to dynamically construct the cc package based on conventions and additional CLI arguments.  This has the benefit of applying directly to most (if not all) of the current sample code base, without introducing the crufty "ccaas" deployment descriptors for yet-another-deployment-technique-for-fabric.

A multi-phase Azure pipeline has been included to verify the overall approach for both golang and Java CCaaS asset transfer chaincode samples.

Some additional / loose ends that may warrant future PRs and improvements: 

- Additional testing on the `host.docker.internal` DNS alias is required (e.g. Win, Linux, revs of Docker, etc.) 

-  The doc sets up a flow with three installations of the same chaincode, but does not delete the chaincode between steps.  (How does one _delete_ a chaincode, btw?) 

-  file-based arguments must all be absolute paths.


asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+export PATH=${PWD}/../../test-network-k8s:$PATH 
+
+network kind 
+
+network up 
+network channel create
+```
+
+```shell
+network chaincode deploy  ${PWD}
```
mbwhite, line 29: A single deploy command is powerful.. and makes dev easier without a doubt. 

Thinking longer term about when this comes to deployments on multi-org network, can we make a single deploy work still?

asset-transfer-basic/chaincode-java/.gitignore
```diff
@@ -0,0 +1,2 @@
+.idea/
+.gradle/
```
mbwhite, line 2: WARNING:  some of the comments may be more thinking out loud, so discussion points rather than fully formed ideas.

asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+network chaincode invoke  asset-transfer-basic '{"Args":["InitLedger"]}'
+network chaincode query   asset-transfer-basic '{"Args":["ReadAsset","asset1"]}' | jq
```
mbwhite, line 32: hint.. call metatata :-)

asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+```shell
+kubectl -n test-network logs -f deployment/org1peer1-cc-asset-transfer-basic
+```
+
+## Debugging 
+
+### Build
+
+```shell
+./gradlew shadowJar
```
mbwhite, line 83: For debug, you can either launch from the IDE or attach to a running process from the IDE.

Matter of preference to an extent. 

Creating the shadow jar might be irrelevant - and really only needed for production deployment.  Plus if you are Eclipse say, then you can just hit run! 

Or maybe just later, we do `./gradlew run` so that compiles and starts? then attach?

asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+```
+or 
+```shell
+docker build -t fabric-samples/asset-transfer-basic/chaincode-java . 
+```
+
+
+### Package
+
+By instructing the peer to connect to chaincode at the Docker host alias `host.docker.internal`, pods running in
```
mbwhite, line 93: didn't use to work on Linux...

asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+on the local system:
+
+```shell
+export CHAINCODE_SERVER_ADDRESS=0.0.0.0:9999
+export CORE_CHAINCODE_ID_NAME=$(network chaincode id $PWD/build/asset-transfer-debug.tgz)
+
+java -jar build/libs/chaincode.jar  
+```
+
+Or using the editor/debugger/IDE of your choice, create a launch target for `ContractMain.main()`, specifying the
```
mbwhite, line 120: Is the main really just a convenience for the IDE to find something to start?

asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+### Approve, Invoke, and Query
+
+After the contract main has launched, install, approve, commit, and invoke the chaincode:
+
+```shell
+# Complete the chaincode lifecycle 
+export CORE_CHAINCODE_ID_NAME=$(network chaincode id $PWD/build/asset-transfer-debug.tgz)
+
+network chaincode install $PWD/build/asset-transfer-debug.tgz 
+network chaincode approve asset-transfer-debug $CORE_CHAINCODE_ID_NAME
```
mbwhite, line 144: Need to clarify how the extra bits, such as private data, version, etc play into this API (for the more complex cases).

jt-nti (changes requested):
This is definitely the simplest experience I've had standing up chaincode as a service, which is fantastic!

I've added some comments below but in general I think it would be worth avoiding the `ccpackage` folders since these mix up chaincode examples with a specific deployment technique, which doesn't seem quite right. I would also be against including "non-standard" files in the ccaas packages (I know, I know, what standard!), plus there are additional files, such as couch indexes that can be included.

This review has reminded me that I had planned to add an updated script to take care of packaging chaincode somewhere more official. It's been floating around for a while but without the peer command supporting ccaas, or a new Fabric CLI, I think it's an important missing piece, and could help here. I've opened a [draft PR on the fabric repo](https://github.com/hyperledger/fabric/pull/3299) to start discussions.

asset-transfer-basic/chaincode-java/README.md
timo-kang, line 32: I think we need `CreateAsset` chaincode first, don't we ?

asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+# Complete the chaincode lifecycle 
+network chaincode install $PWD/build/asset-transfer.tgz 
+network chaincode approve asset-transfer-basic $CORE_CHAINCODE_ID_NAME
+network chaincode commit  asset-transfer-basic
+```
+
+```shell
+# execute the smart contract by name 
+network chaincode invoke  asset-transfer-basic '{"Args":["InitLedger"]}'
+network chaincode query   asset-transfer-basic '{"Args":["ReadAsset","asset1"]}'
```
timo-kang, line 71: don't we need any test `create` chaincode here?

asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+This project is also compatible with the legacy chaincode builder pipeline and the compose based test-network.
+For additional details, see the [End-to-end with the test-network](../../test-network/CHAINCODE_AS_A_SERVICE_TUTORIAL.md#end-to-end-with-the-the-test-network)
+documentation.
+
+## [Quickstart](../../test-network-k8s#quickstart)
+
+```shell
+export PATH=${PWD}/../../test-network-k8s:$PATH 
+
+network kind
```
timo-kang, line 22: for common reason, why don't we use network cmd with `./network` instead of `network`?

asset-transfer-basic/chaincode-java/README.md
```diff
@@ -0,0 +1,164 @@
...
+docker build -t hyperledger/fabric-samples/asset-transfer-basic/chaincode-java .
+
+# Load the docker image directly to the KIND control plane.  
+# (Alternately, build/tag/push the image to a remote container registry, e.g. localhost:5000) 
+kind load docker-image hyperledger/fabric-samples/asset-transfer-basic/chaincode-java
+```
+
+```shell
+# Assemble the chaincode package archive 
+network chaincode package $PWD/ccpackage/ $PWD/build/asset-transfer.tgz
```
timo-kang, line 54: In this project, it seems that the extension of package is used as `tar.gz`

asset-transfer-basic/chaincode-java/SCRATCH.md
```diff
@@ -0,0 +1,49 @@
...
+[params] => --param=value or NETWORK_$GROUP_$COMMAND_$PARAM=value from env 
+```
+
+```shell
+network chaincode package       <folder-path> <bundle-path> 
+network chaincode id            <bundle-path> 
+network chaincode install       <bundle-path> 
+network chaincode approve       <name> <id> 
+network chainocde commit        <name> 
+```
```
timo-kang, line 22: It's fascinating and seems useful that the `network`can do all steps.
but, I cannot asure that this pattern doesn't obey SRP.

appreciated if you give any comments

asset-transfer-basic/chaincode-java/gradlew
```diff
@@ -97,92 +132,103 @@ Please set the JAVA_HOME variable in your environment to match the
...
+      *)
+        ulimit -n "$MAX_FD" ||
+            warn "Could not set maximum file descriptor limit to $MAX_FD"
+    esac
 fi
 
-# For Darwin, add options to specify how the application appears in the dock
-if $darwin; then
-    GRADLE_OPTS="$GRADLE_OPTS \"-Xdock:name=$APP_NAME\" \"-Xdock:icon=$APP_HOME/media/gradle.icns\""
-fi
```
timo-kang, line 126: not implemented for Darwin?

asset-transfer-basic/chaincode-java/src/main/java/org/hyperledger/fabric/samples/assettransfer/ContractMain.java
```diff
@@ -0,0 +1,23 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.hyperledger.fabric.samples.assettransfer;
+
+import org.hyperledger.fabric.contract.ContractRouter;
+
+public final class ContractMain {
```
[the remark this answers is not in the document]
timo-kang: I don't think so.
I think that this changes give more advantages rather than backward

timo-kang (commented):
Thank you for this good example!
I think that most of the changes are great, but I have some doubts that whether using single `network` interface is better or not, by loosing options for selecting certain peers to run features.

asset-transfer-basic/chaincode-java/README.md
jkneubuh: For "deploy" to work we will need to introduce some higher-order descriptors that articulate the structure of the overall network.  This would be a really nice addition, IMO.  At a meta level, the person configuring a Fabric network really should only focus on high-level constructs such as Organizations, Consortiums, Users (Identities), Channels, Chaincode, etc... and let some sort of software-automation package take responsibility for the minutia of configuration tasks.

Test-network[-k8s] is a good area for prototyping a high-level configuration target, but we should plan for alignment with some of the other "meta" Fabric configuration packages such as fabric-test-operator, minikube, *-k8s-operator, etc.   Ideally the '/network deploy' target can utilize whatever configuration specification is used by a common descriptor language.

asset-transfer-basic/chaincode-java/.gitignore
jkneubuh: Don't ignore good `.ideas` 
👍

asset-transfer-basic/chaincode-java/README.md
jkneubuh: `gradle run` would be a nice addition here.  IMO it should be managed in the context of a build / gradle plugin.

The trick to the run target is in picking up and determining the CC ID, which need to be in the scope of main()'s env.  This is straightforward, especially if the gradle task is constructing the chaincode package archive.    Note that this does not _need_ to be in a gradle plugin - it should be possible to shell out to generate the sha256 hash, assemble the package, etc. from a gradle build script.

a `run` target would be nice.  In some ways it is actually easier in Eclipse / IntelliJ, which have good integration with gradle, for a local debug.  (I.e.. right-click-run/debug on a gradle task that computes the package ID would save the end-user from digging this out manually from the CLI and loading into the main() scope for the run target.)

asset-transfer-basic/chaincode-java/README.md
jkneubuh: oof.

https://stackoverflow.com/questions/48546124/what-is-linux-equivalent-of-host-docker-internal 

https://github.com/moby/libnetwork/pull/2348

test-network-k8s/scripts/chaincode.sh
```diff
@@ -5,45 +5,102 @@
...
+  local cc_archive=$2
+  local archive_name=$(basename $cc_archive)
+  push_fn "Packaging chaincode ${archive_name}"
 
-  # Install the chaincode
-  echo 'set -x
-  export CORE_PEER_ADDRESS='${org}'-'${peer}':7051
-  peer lifecycle chaincode install build/chaincode/'${CHAINCODE_NAME}'.tgz
-  ' | exec kubectl -n $NS exec deploy/${org}-admin-cli -c main -i -- /bin/bash
+  tar -C ${cc_folder} -zcf ${cc_folder}/code.tar.gz connection.json ccaas.json
```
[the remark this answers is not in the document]
jkneubuh: @jt-nti you are doing a good job of convincing me that the ccaas.json does NOT belong in the package.

test-network-k8s/scripts/chaincode.sh
```diff
@@ -70,120 +127,177 @@ function launch_chaincode_service() {
...
+function transfer_chaincode_archive_for() {
   local org=$1
+  local cc_archive=$2
+  local dirname=$(dirname $cc_archive)
+  local filename=$(basename $cc_archive)
+
+  push_fn "Transferring chaincode archive to ${org}"
+
+  # Like kubectl cp, but targeted to a deployment rather than an individual pod.
+  tar cf - -C ${dirname} ${filename} | kubectl -n $NS exec -i deploy/${org}-admin-cli -c main -- tar xvf -
```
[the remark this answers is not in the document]
jkneubuh: Yes, and no...   I am working on a new PR that will start using the `peer` binaries locally on the host OS, rather than copying ANYTHING to/from k8s with obscure and arcane tar commands.  I have NO idea what that `cp` line above is doing.  (It hurts that I wrote it... and still have no idea what it's doing...)

asset-transfer-basic/chaincode-java/README.md
jkneubuh: `InitLedger` will create a few assets from within a method in the chaincode.

CreateAsset is also OK to use.  It is just a little more "typing" of JSON, which is sometimes hard for my fingers.

asset-transfer-basic/chaincode-java/README.md
jkneubuh: Adding `network` to the PATH makes it a little easier for users to work with other chaincode projects in the fabric-samples project.   I.e.. a user should be able to: 

```
cd fabric-samples

network chaincode deploy asset-transfer-basic/chaincode-java 
``` 

or 

```
cd fabric-samples/test-network-k8s 

network chaincode deploy ../asset-transfer-basic/chaincode-go
```

`./network` is OK but it means users always need to be in the test-network-k8s directory.

asset-transfer-basic/chaincode-java/SCRATCH.md
jkneubuh: I think we can use `./network` as an approximation to the functional patterns for a `fabric` CLI.

There was some work on a [fabric-cli](https://github.com/hyperledger/fabric-cli) a few years ago, but it stalled.

Each of the different routines in the `network` script are really part of a _command group_, i.e.. `chaincode` or `channel` or ... 

I think a really good way to carry this effort forward is to construct a golang `fabric-config` SDK, with individual modules or packages that understand the various functional command groups.  The routines should ONLY work with the gRPC and https endpoints, and introduce a better organization for _Identity_ and _Context_ of an API call.

If we have a solid golang module that can administer channels, chaincode, enrollments, etc., this can be incorporated into a `chaincode` CLI binary, running as a sub-command of a higher-level `fabric` CLI driver.

More importantly, the golang module should not ONLY serve the CLI consumers, but should also serve as the basis for software automation, such as Kubernetes Operators for Fabric.

With the SRP principal, think of `chaincode`, `channel`, `context`, etc... as the utility.  Ideally the `/network` (or `fabric` / `fabctl`) routine is a simple switch to dispatch to the correct sub-command.

asset-transfer-basic/chaincode-java/gradlew
jkneubuh: I updated a gradle bootstrap to use gradle 7.  There were some changes in the scripts but I am sure it works OK on OSX / Mac with the latest updates.

jkneubuh:
Hi @jt-nti - I made some updates today to remove the ccpackage/*.json files from the setup.  In this version, the parameters from the descriptor files are all determined from conventions and routines in the script.  In retrospect I like that the sample folder is a bit cleaner, and leaves all of the "CCaaS" bits bundled into the network setup scripts.  However, I do _not_ like how the changes have impacted some of the readability of the example or clarity of the approach.

The three bits of information necessary to infer to launch the chaincode are: 

- chaincode label 
- chaincode Docker image 
- chaincode name (for use launching the service endpoint associated with an approval / commit) 

I want to leave this approach here for a bit for review, but I would really prefer if we could compromise on an approach that leaves a single "ccaas.json or yaml" descriptor around as an input file in each of the sample folders.  Or at least - to code up the alternate approach and discuss pros + cons of the different techniques.  It was quite complicated to infer some of these attributes, and I must admit I'm running out of ideas to generate the Docker image name in a way that passes through the CI build, as Azure pulls the directory structure for the routine into a different folder root.

Overall I think the approach is better than what was in the PR yesterday, but still has some way to go before it's "ready."  Would appreciate a second pair of eyes to see how the approach sketched out this AM ended up in code.  Thanks.

jt-nti (changes requested):
The instructions shouldn't be in `asset-transfer-basic/chaincode-java/README.md` but looks good otherwise- much less cruft :)

jkneubuh:
The thread is completely lost in this PR....  closing and replacing with PR #724

PR #689, opened 2022-03-28
Update the REST Example

mbwhite:
1. For development purposes CORS can hinder testing examples. Therefore enable CORS if needed, default off
2. The default chaincode use Capitalized JSON Names, whereas for new asset the REST validation didn't.. Adjusted
so it matches the chaincode

### DIFF
--- a/asset-transfer-basic/rest-api-typescript/src/__tests__/api.test.ts
+++ b/asset-transfer-basic/rest-api-typescript/src/__tests__/api.test.ts
@@ -208,11 +208,11 @@ describe('Asset Transfer Besic REST API', () => {
       const response = await request(app)
         .post('/api/assets')
         .send({
-          identifier: 'asset3',
-          color: 'red',
-          size: 5,
-          owner: 'Brad',
-          appraisedValue: 400,
+          wrongidfield: 'asset3',
+          Color: 'red',
+          Size: 5,
+          Owner: 'Brad',
+          AppraisedValue: 400,
         })
         .set('X-Api-Key', 'ORG1MOCKAPIKEY');
       expect(response.statusCode).toEqual(400);
@@ -227,7 +227,7 @@ describe('Asset Transfer Besic REST API', () => {
           {
             location: 'body',
             msg: 'must be a string',
-            param: 'id',
+            param: 'ID',
           },
         ],
         message: 'Invalid request body',
@@ -239,11 +239,11 @@ describe('Asset Transfer Besic REST API', () => {
       const response = await request(app)
         .post('/api/assets')
         .send({
-          id: 'asset3',
-          color: 'red',
-          size: 5,
-          owner: 'Brad',
-          appraisedValue: 400,
+          ID: 'asset3',
+          Color: 'red',
+          Size: 5,
+          Owner: 'Brad',
+          AppraisedValue: 400,
         })
         .set('X-Api-Key', 'ORG1MOCKAPIKEY');
       expect(response.statusCode).toEqual(202);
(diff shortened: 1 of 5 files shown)


asset-transfer-basic/rest-api-typescript/package.json
```diff
@@ -8,7 +8,9 @@
     "npm": ">=5"
   },
   "dependencies": {
+    "async-mqtt": "^2.6.2",
```
jt-nti, line 11: I don't think this is required?

jt-nti (changes requested):
Thanks for the PR- it definitely makes sense for the case to match the contract. I think it also needs fixing for PUT and PATCH though.

asset-transfer-basic/rest-api-typescript/src/assets.router.ts
```diff
@@ -35,7 +35,7 @@ export const assetsRouter = express.Router();
 
 assetsRouter.get('/', async (req: Request, res: Response) => {
   logger.debug('Get all assets request received');
-
+  logger.debug(req);
```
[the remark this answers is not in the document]
mbwhite: it does get logged, but only in the event of an error.. it doesn't get logged if everything works.
added to see what request was that was coming from CURL that was working... 

I'll remove it

asset-transfer-basic/rest-api-typescript/src/server.ts
```diff
@@ -13,12 +13,23 @@ import { healthRouter } from './health.router';
...
+import { devCorsEnable } from './config';
+import cors from 'cors';
 
 const { BAD_REQUEST, INTERNAL_SERVER_ERROR, NOT_FOUND } = StatusCodes;
 
 export const createServer = async (): Promise<Application> => {
   const app = express();
 
+  // Only for development, wild card origin not recommended for production
+  if (devCorsEnable){
```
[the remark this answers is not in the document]
mbwhite: yep as discussed will remove it..

asset-transfer-basic/rest-api-typescript/src/__tests__/api.test.ts
```diff
@@ -208,11 +208,11 @@ describe('Asset Transfer Besic REST API', () => {
       const response = await request(app)
         .post('/api/assets')
         .send({
-          identifier: 'asset3',
-          color: 'red',
-          size: 5,
-          owner: 'Brad',
-          appraisedValue: 400,
+          wrongidfield: 'asset3',
```
jt-nti, line 211: Aha, I like that, much more obvious!

asset-transfer-basic/rest-api-typescript/.gitignore
```diff
@@ -13,3 +13,5 @@ jspm_packages/
 
 # Compiled TypeScript files
 dist
+
+package-lock.json
```
jt-nti, line 17: package-lock.json should not be in .gitignore

mbwhite:
@hyperledger/fabric-samples-maintainers   could you take a look and merge please?  Thanks

jkneubuh:
How can I take this PR out for a test drive?

Simple web app adjacent to the REST sample, exposed at a different host name?

mbwhite:
@jkneubuh by means, though you'll need to provide your own insurance..

The relevant parts ended up being making the JSON structure accepted on 'write' actions the same as on 'read' actions. 

(where we need a canonical model)

jkneubuh:
Why remove the `devCORS` override?  

I would like to access the REST sample endpoint, running on localhost, from a web app / `XMLHttpRequest` loaded from a different domain.  It's OK if this is restrictive by default - is there another way to enable this via config of the service endpoint / env?

jt-nti:
Although CORS support of some kind might be nice to have, the REST sample was not intended to get into all the various options available for running a REST server. After a bit of discussion the CORS change was removed from this PR because it wasn't related to the other changes.

I'm not fundamentally opposed to adding additional configuration options for scenarios that support the intent to show a production-like sample, however I am slightly nervous about adding too many variations when the sample is arguably already too complex.

jkneubuh (approved):
This PR Looks OK.

I Preferred The Normal, CamelCase Capitalization Scheme for JSON.

However... Consistency With The Underlying Contract Structures Is Important, Too.

Looks Good.  Ship It.

Issue #685, opened 2022-03-23
Kube test network : Run fabric binaries on the host OS, not in an "admin pod" 


jkneubuh:
The Kube test network currently uses an "admin CLI" pod to run administrative commands.  When `./network` scripts run on the host OS, they need to run a complicated series of `kubectl exec ...` into the admin CLI pod, set the environment scope, and run within the cluster.

This is impractical for a number of reasons.  Here are some challenges: 

- it's weird.

- it locks up "everything" in Kubernetes.   Transferring assets in/out of k8s is complicated, error prone, and arcane.

- It's an anti-pattern.  We have client-side binaries such as `peer`, `osnadmin`, and `fabric-ca-client` -- use them!

- Certificate structures (registration, enrollment, MSP construction, etc.) are _obscured_ by hiding within K8s. 

- fabric-utils Docker image has a limited support path.  It is 'on the way out'  

Update the test network such that it embraces the oddities of core Fabric, rather than work around them for convenience.   Specifically: 

1. Make use of the Nginx ingress controller.   Set up an Ingress such that each Service may be accessed from the host OS via a stable URL.   Use a DNS wildcard resolver such as *.vcap.me (-> 127.0.0.1) to minimize the need to construct virtual host entries on the client workstation.  (For instance:  `grpcs://org1-peer1.org1.vcap.me:443` will resolve to the Nginx ingress, routing traffic to the `org1-peer1` deployment.)

2. Migrate the `exec CLI` commands out to run on the host OS. 

3. Assemble Org Admin and channel MSP structures using the fabric-ca-client to the CA via Ingress.   I.e.. store the "wallet" on the host OS. 

4. Download the fabric client binaries as part of the script bootstrap.  This is also a good opportunity to "stage" the Fabric Docker images locally via the normal Fabric "samples + binaries" transfer process.

jkneubuh:
Fixed!

The new approach is great.  It's _MUCH EASIER_ to work with the native fabric binaries, connecting to grpcs and https endpoints via Ingress.

PR #692, opened 2022-03-29
test-network-k8s : Connect to Fabric services via Nginx Ingress - READY FOR MERGE

jkneubuh:
This PR updates the Kube test network to access Fabric services via an Ingress controller.  This allows for easy access to the CAs, peers, orderers, and load-balanced Gateway Peer via the nginx bound to localhost:443.

In previous iterations, any administration of the network involved a remote exec to execute commands on an "admin CLI pod" running in the cluster.  In addition, any time files needed to get to/from the volume mounts in Kubernetes, an overly complicated sequence of arcane `exec tar --.... ` commands was necessary.

With this PR, the network is instantiated, exposing core services at the *.vcap.me (wildcard match to 127.0.0.1) DNS domain.  Fabric binaries are run directly on the host, and certificate enrollments are managed locally as if they were maintained in a local wallet storage.


jkneubuh:
Addresses Issue #685

This PR is generally in good shape for a peer review.

There is still some errant flakiness when accessing the CA to determine the ECert signing certificate.  On occasion the Nginx ingress (or possibly the org CA) is returning an empty body from the /cainfo URL, leading to an empty ca-cert.pem.  Adding a slight delay between issuing setup commands seems to work around the issue.   If the network or test suite is run as a single copy/paste command, or in quick succession, it will occasionally flake.  I'll track this down before removing the DRAFT PR status.

mbwhite:
Test with Ubuntu image running WSL2 commencing...

- Fabric binary is 2.4.2.. (2.4.3 is the latest)

mbwhite (commented):
@jkneubuh `vcap.me` needs mapping to `127.0.0.1`  is this documented in the readme?

jkneubuh:
No *.vcap.me is a "real thing."  This was set up by VMWare to help ... access vm services running on a loopback with a DNS wildcard.  The alias does NOT need any local configuration - it should be handled OK with public DNS.  Are you having issues resolving this with BT?

If this approach "sticks" it would be nice to set up an alternative, more friendly name, managed by Hyperledger Foundation or an enterprise sponsor of the project.  For instance, publish a PUBLIC DNS A / wildcard record : 

*.loopback.fabric.hyperledger.org --> 127.0.0.1 

vcap.me gets the job done OK, but the name is not ideal...

mbwhite:
wow - did not know that...  you learn something new every day... yes `vcap.me` does work for me. 

And here's me thinking 'must ask Josh why he's chosen such an odd name,'

PR #696, opened 2022-03-31
CORS Enablement

mbwhite:
After feedback from the community, this PR adds CORS support in, but as an option
to enable.

Along with warnings that the default wildcard origin might not be applicable
in a production use case

### DIFF
--- a/asset-transfer-basic/rest-api-typescript/src/server.ts
+++ b/asset-transfer-basic/rest-api-typescript/src/server.ts
@@ -13,13 +13,12 @@ import { healthRouter } from './health.router';
 import { jobsRouter } from './jobs.router';
 import { logger } from './logger';
 import { transactionsRouter } from './transactions.router';
+import cors from 'cors';
 
 const { BAD_REQUEST, INTERNAL_SERVER_ERROR, NOT_FOUND } = StatusCodes;
 
 export const createServer = async (): Promise<Application> => {
   const app = express();
-
-  // Remember for production usage, to check any TLS or CORS requirements
 
   app.use(
     pinoMiddleware({
@@ -51,7 +50,7 @@ export const createServer = async (): Promise<Application> => {
   app.use(passport.initialize());
 
   if (process.env.NODE_ENV === 'development') {
-    // TBC
+    app.use(cors());
   }
 
   if (process.env.NODE_ENV === 'test') {
--- a/asset-transfer-basic/rest-api-typescript/package-lock.json
+++ b/asset-transfer-basic/rest-api-typescript/package-lock.json
@@ -10,6 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "bullmq": "^1.47.2",
+        "cors": "^2.8.5",
         "dotenv": "^10.0.0",
         "env-var": "^7.0.1",
         "express": "^4.17.1",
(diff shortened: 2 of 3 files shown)


asset-transfer-basic/rest-api-typescript/src/server.ts
```diff
@@ -13,13 +13,19 @@ import { healthRouter } from './health.router';
...
+import { devCorsEnabled } from './config';
 
 const { BAD_REQUEST, INTERNAL_SERVER_ERROR, NOT_FOUND } = StatusCodes;
 
 export const createServer = async (): Promise<Application> => {
   const app = express();
 
   // Remember for production usage, to check any TLS or CORS requirements
+  // this wildcard option *MIGHT* not be applicable to your environment
+  if (devCorsEnabled) {
```
jt-nti, line 26: Following on from config comment, I think it would be good to move this to the existing dev mode block

```
if (process.env.NODE_ENV === 'development') {
  // TBC
}
```

https://github.com/hyperledger/fabric-samples/blob/main/asset-transfer-basic/rest-api-typescript/src/server.ts#L54

asset-transfer-basic/rest-api-typescript/src/config.ts
```diff
@@ -283,3 +283,12 @@ export const org2ApiKey = env
   .required()
   .example('456')
   .asString();
+
+/** Enable CORS, set to be the wildcard origin
+ * THis might not be applicable for all production use cases
+ */
+export const devCorsEnabled = env
```
jt-nti, line 290: Rather than add a specific CORS option (with associated tests 👀) without any origin configuration, I think I would prefer to make CORS dev mode only to reinforce that it's really not intended for production

asset-transfer-basic/rest-api-typescript/src/server.ts
```diff
@@ -13,13 +13,19 @@ import { healthRouter } from './health.router';
...
+    app.use(cors());
```
jt-nti, line 27: The wildcard option still doesn't seem right- is `localhost` a possibility? Also, I think some extra config is required since the sample makes use of an `X-API-Key` header- is that enough to require preflight to be enabled?
>Certain CORS requests are considered 'complex' and require an initial OPTIONS request (called the "pre-flight request"). An example of a 'complex' CORS request is one that uses an HTTP verb other than GET/HEAD/POST (such as DELETE) or that uses custom headers.

jt-nti (changes requested):
A couple of comments, and the `package-lock.json` should stay on the same version.

asset-transfer-basic/rest-api-typescript/src/server.ts
mbwhite: can't say for certain what the theory states, but I was testing with all the HTTP verbs on offer and it was working :-)

asset-transfer-basic/rest-api-typescript/src/server.ts
mbwhite: So...

```
  if (process.env.NODE_ENV === 'development') {
    app.use(cors());
  }
```

asset-transfer-basic/rest-api-typescript/src/server.ts
jt-nti: Well if it works on your machine... :)

jt-nti (approved):
Ship it!

jt-nti (approved):

jkneubuh (approved):
💯 👍

PR #715, opened 2022-04-14
fix wrong indentationin in the connection profile  generated for Org 3

basilky:
The connection profile file(yaml) generated for Org3 is having wrong indentation and the Fabric Go SDK fails to parse it. This PR fixes the issue.


jkneubuh (approved):
The tlsCACerts pem looks OK as an array.  Thanks for the update.

PR #717, opened 2022-04-19
Can implement a simple CURD

ShanZJ-gvs:
Leave the serialization of parameters to the upper level

Signed-off-by: D <gvssimux@qq.com>

Co-authored-by: D <gvssimux@qq.com>


denyeart:
Thank you for the submission. We're trying to keep the official samples to a manageable set of blockchain core concepts and aligned with the tutorials in the documentation, therefore we are not looking to expand the set for various permutations of CRUD patterns beyond the asset-transfer samples. The main point of this PR is valid, but I would suggest to mention the option of passing JSON instead of inputs in the existing asset-transfer-basic sample and associated doc tutorial with a small code snippet, rather than creating yet another chaincode asset in fabric-samples that needs to be maintained going forward. I'll close for now.

PR #725, opened 2022-04-25
Allow custom peer images for test-network-k8s

jt-nti:
The Hyperledger provided Fabric peer images are only a sample so it would be good to make the k8s test network peer image configurable, for example to include a custom builder


jkneubuh (approved):
💯

PR #727, opened 2022-04-29
add js cc for the private data

abdou-chakhkhar:
I was testing the private data concept, I noticed that the chaincode is only written with go and java then I decided to take this initiative and implement it with js, the cc was tested with the application I hope that would help js developers understand private data and use nodejs sdk easily.


lehors:
@abdou-chakhkhar Any interest in pursuing this? It'd be nice to get this in a state where we can merge it.
Thanks.

abdou-chakhkhar:
Hello,

Sorry for the late reply, i did not see your emails i will see what i can
do.

Abdou.

Issue #550, opened 2021-12-01
Kube test network : Pods can not resolve external DNS host names 


jkneubuh:
Pods running in the Kubernetes Test Network are no longer able to resolve external DNS entries.  E.g.: 

Setup: 
- Set up a new Linux host 
- git clone https://github.com/hyperledger/fabric-samples.git
- cd fabric-samples/test-network-k8s 
- `./network kind` 

Resolve a local DNS host / pod (OK): 
```
root@hexapody1:~# kubectl apply -f https://k8s.io/examples/admin/dns/dnsutils.yaml
pod/dnsutils created
root@hexapody1:~# kubectl exec -i -t dnsutils -- nslookup kubernetes.default
Server:		10.96.0.10
Address:	10.96.0.10#53

Name:	kubernetes.default.svc.cluster.local
Address: 10.96.0.1
```

Resolve an external DNS host name (FAILS): 
```
root@hexapody1:~# kubectl exec -i -t dnsutils -- nslookup www.ibm.com 
Server:		10.96.0.10
Address:	10.96.0.10#53

** server can't find www.ibm.com.fyre.ibm.com: SERVFAIL

command terminated with exit code 1
```

jkneubuh:
This is caused by [KIND Issue #2326](https://github.com/kubernetes-sigs/kind/issues/2326). 

Recently we enabled the [`IPVS` proxy routing](https://kubernetes.io/docs/concepts/services-networking/service/#proxy-mode-ipvs) feature for the Kubernetes Test Network, in part to expose additional options for load balancing a set of gateway peers behind a shared service / DNS alias.  Unfortunately this tickled a bug in the way KIND proxies DNS lookups through the Docker network, causing external lookups to fail.

Fix this issue by : 

1. revert IPVS -> IPTABLES routing when bootstrapping the KIND cluster.
2. Update the Fabric Deployment Guide with additional instructions for IPVS Service Routing.
3. Document the [`route_localnet=1`](https://github.com/kubernetes-sigs/kind/issues/2326#issuecomment-881969982) workaround, for KIND environments validating IPVS: 

```
# Workaround https://github.com/kubernetes-sigs/kind/issues/2326
for node in $(kind get nodes); do
  docker exec "$node" sysctl net.ipv4.conf.all.route_localnet=1
done
```

I've confirmed the above workaround re-enables DNS. 

@denyeart @mbwhite cc:

jkneubuh:
Fixed in PR #729

Issue #506, opened 2021-10-21
test-network-k8s: `./network kind` error creating cluster should be reported in the terminal


seanpk:
I have `docker` and `kubectl` installed as snaps, and since I'm running a VPN I needed to add a manual network using `brctl` per [docker/for-linux #1105](https://github.com/docker/for-linux/issues/1105#issuecomment-948200872).

My `./network kind` command is only getting a little further however:
```
$ ./network kind
Initializing KIND cluster "kind":
✅ - Pulling docker images for Fabric 2.3.2 ...
⚠️  - Creating cluster "kind" ...
$
```

This is pretty, but not that helpful. At the very least the user should get a message to check out the file `network-debug.log`, but publishing the error message to the console would be even better. I've attached that file ([network-debug.log](https://github.com/hyperledger/fabric-samples/files/7385684/network-debug.log)) for reference (and maybe for help with whatever my config issue is), but here are the relevant lines from it too (which look similar to my earlier VPN issue):
```
Status: Downloaded newer image for hyperledger/fabric-tools:2.3.2
docker.io/hyperledger/fabric-tools:2.3.2
Deleting cluster "kind" ...
Creating cluster "kind" ...
 • Ensuring node image (kindest/node:v1.21.1) 🖼  ...
ERROR: failed to create cluster: failed to ensure docker network: command "docker network create -d=bridge -o com.docker.network.bridge.enable_ip_masquerade=true -o com.docker.network.driver.mtu=1500 --ipv6 --subnet fc00:f853:ccd:e793::/64 kind" failed with error: exit status 1

Command Output: Error response from daemon: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network
```

denyeart:
@jkneubuh

jkneubuh:
Fixed in PR #729

PR #729, opened 2022-05-02
Run the Kube Test Network on Rancher / k3s

### LINKED ISSUE #728: Kube test network : run on Rancher Desktop / k3s
KIND is good for local development, but it has some rough edges and can be improved.  As an alternate runtime for local development, use the [Rancher Desktop](https://rancherdesktop.io) and k3s as an alternate target for deployment on Kubernetes.

This is extremely useful in some environments, for example on Mac OSX without a Docker Desktop.

In addition, the k3s has visibility to the host's Docker image storage, so there is no longer a need to use the insecure Docker registry at localhost:5000 or load development images into the KIND control plane during local edit/test/build cycles.

Update the `./network` scripts such that it can support an alternate Kubernetes runtime.   Some issues that will need to be addressed in the scripts include: 

- Set STORAGE_CLASS=`local-path` when running on Rancher (was `standard` on KIND) 

- Update the Nginx ingress definitions to 1.1.2 and set `--enable-ssl-passthrough` on the controller CLI args: 
```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.1.2/deploy/static/provider/cloud/deploy.yaml
```
(Add these to manifests/namespace-install, or read from the base URL and kustomize the passthrough flag) 

- Update the README with a `RANCHER.md` guide 

- Add conditional branching for `TEST_NETWORK_CLUSTER_RUNTIME=rancher` to bypass any KIND-specific routines in the network scripts.

jkneubuh:
This PR enables users of the Kube Test Network to set up a Fabric network based on k3s / [Rancher Desktop](https://rancherdesktop.io)

- Fixes #728 
- Upgrades Nginx ingress to 1.1.2 
- Aligns both KIND and Rancher runtimes to use the local-path provisioner


jt-nti (changes requested):
Doesn't work on my machine :(

- scripts/prereqs.sh is checking for kind even when TEST_NETWORK_CLUSTER_RUNTIME is set to k3s
- there is no `envsubst` command on OSX so get errors, e.g. `scripts/fabric_config.sh: line 35: envsubst: command not found` (think this probably just needs to be in prereqs.sh as well)

test-network-k8s/docs/KUBERNETES.md
```diff
@@ -69,6 +69,36 @@ or:
...
+   4. Increase CPU allocation to 8 CPU 
+
+2. Reset Kubernetes 
+
+3. Initialize the Nginx ingress and cert-manager: 
+
+```shell
+export TEST_NETWORK_CLUSTER_RUNTIME=k3s
+
+./network cluster-init
```
jt-nti, line 91: Just had another thought- is there an equivalent to `unkind` for this?

jkneubuh: Rancher desktop maintains a 1:1 "rancher" : "kubernetes is running" strategy.   So there's no real way to delete the rancher kube without going through the Rancher GUI and hitting the "reset Kubernetes" button.

Not to mention ... I couldn't find a verb that served as the inverse of "to rancher."  E.g.  `./network unrancher` doesn't have the same zing as `unkind`  -- I did try to find a good word for this, but abandoned the effort. 

I did make some changes to the teardown / `network down` script, to better scrub the kubernetes cluster of test network artifacts.

jt-nti: Just wondering what undoes what `cluster-init` did really. I'm assuming that `network down` won't undo this?

```
✅ - Launching k3s ingress controller ...
✅ - Launching cert-manager ...
✅ - Waiting for cert-manager ...
✅ - Waiting for ingress controller ...
```

Maybe rename `network unkind` to `network clean`? Actually a generic `network prereqs` or some such which does the right thing depending on `TEST_NETWORK_CLUSTER_RUNTIME` might be better than having to pick between running `network kind` or `network cluster-init`?

Not worth holding this PR up for though.

jkneubuh:
There is some crossover in this PR with Fabric [Issue #3372](https://github.com/hyperledger/fabric/issues/3372).

The latest commit on this PR works around a slow DNS resolution issue (5s / timeout) by increasing the timeout for connections to the orderer with the peer `--connTimeout` flag.

We can, should, and will fix the DNS resolver issues on OSX.  Just - not in this ticket / PR.

jkneubuh:
Thanks @denyeart for the test and validation / troubleshooting.  The latest commit introduces:

- 10s delay for the peer delivery client connection timeout in the core configs. 
- Highlights the prerequisite for 8 CPU / 8GRAM in the docker / rancher runtimes.

mbwhite:
Hello @jkneubuh.... 

- the different start comments for the cluster - can these be combined? Something along these lines... (with one of them being a default)
```
export TEST_NETWORK_CLUSTER_RUNTIME=k3s
./network cluster-init 

export TEST_NETWORK_CLUSTER_RUNTIME=kind
./network cluster-init 
```

- probably error handling messages are the weakest point of an otherwise a winning script.

jkneubuh:
OK.  This PR is (over) cooked and done.

- Fixes #506 (pretty prints an error message AND an emoji on error) 
- Fixes #728 (runs with both containerd/nerdctl and mobyd/docker) 
- Fixes #550 (or verified that this is indeed addressed)
- Includes a good workaround for [Fabric #3372](https://github.com/hyperledger/fabric/issues/3372) (slow DNS on OSX)
- Ironically - test flakes in the CI pipeline were due to collisions (unrelated) to a Docker Compose bug.

_Goodbye, Docker!_

PR #731, opened 2022-05-06
Updated ERC tokens samples

fraVlaca:
- added check for math oveflow where necessary
- added name, symbol and default token options to all go erc sample that were missing these options
- renamed `SetOptions()` to `Initialize()`
- made sure that you need to initialize contract and its options (call `Initialize()`) first before calling any contract functions
- made sure `Inizialized()` can be called only once


jkneubuh (approved):
Thanks @fraVlaca these are great updates.

Looks like you caught a couple of bugs in the Java CC example, as well.

This is not a suggestion for this PR, which is great, but ...  Reading through the routines below I do wonder if it would be an interesting demonstration and/or functional test to set up a multi-organization blockchain, with each org installing a different revision / build of the token contract.

I.e.. 
- org1 runs the golang ERC contract, 
- org2 runs the Java ERC contract,
- org3 runs the node ERC contract 
- gateway app and/or E2E test runs through scenarios illustrated in the READMEs. 

I think that would be neat.   Illustrating / Enforcing that different orgs run different implementations of functionally equivalent chaincode.

fraVlaca:
Could definitely have a try, do you have a more detailed idea in what the experiment and goals of it would look like? @jkneubuh 
Would the aim be to do the same actions on different implementations of functionally equivalent chaincode and test that they all have the same result despite the implementation?

token-erc-20/chaincode-go/chaincode/token_contract.go
```diff
@@ -421,6 +534,90 @@ func (s *SmartContract) TransferFrom(ctx contractapi.TransactionContextInterface
...
+		return "", fmt.Errorf("failed to get Symbol: %v", err)
+	}
+
+	return string(bytes), nil
+}
+
+// Set information for a token and intialize contract.
+// param {String} name The name of the token
+// param {String} symbol The symbol of the token
+// param {String} decimals The name of the token
```
denyeart, line 584: Copy/Paste typo

token-erc-20/chaincode-javascript/test/tokenERC20.test.js
```diff
@@ -198,13 +200,18 @@ describe('Chaincode', () => {
         });
     });
 
-    describe('#SetOption', () => {
+    describe('#Initialize', () => {
         it('should work', async () => {
-            const response = await token.SetOption(ctx, 'some name', 'some symbol', '2');
+            //we consider that is been already initialize in the before each statement
```
denyeart, line 205: typo

token-erc-1155/chaincode-go/chaincode/contract.go
```diff
@@ -594,6 +757,84 @@ func (s *SmartContract) BroadcastTokenExistance(ctx contractapi.TransactionConte
...
+// Set information for a token and intialize contract.
+// param {String} name The name of the token
+// param {String} symbol The symbol of the token
+func (s *SmartContract) Initialize(ctx contractapi.TransactionContextInterface, name string, symbol string) (bool, error) {
+
+	// Check minter authorization - this sample assumes Org1 is the central banker with privilege to intitialize contract
+	clientMSPID, err := ctx.GetClientIdentity().GetMSPID()
+	if err != nil {
+		return false, fmt.Errorf("failed to get MSPID: %v", err)
+	} else if clientMSPID != "Org1MSP" {
```
denyeart, line 813: In idiomatic Go code we try to minimize indents.
In this case we don't need an "else if", if there was an error we would have already returned. So we can just do all the checks without nesting. This makes it easier to scan the code by just moving your eyes down vertically.

Same applies in several files in this PR.

PR #735, opened 2022-05-11
test-network-k8s: Add CC debugging procedures for Linux

satota2:
In chaincode debugging guidance, using the Docker host alias `host.docker.internal` is assumed.
But, the alias is not yet supported for Linux.
So, this patch adds its alternate procedure for Linux.

Signed-off-by: Tatsuya Sato <tatsuya.sato.so@hitachi.com>


satota2:
@jkneubuh
> BTW there are some recent updates to the Kube test network to run on Rancher Desktop (k3s). Rancher's Kubernetes has good visibility to the host NIC and does not need to use the host.docker.internal / host alias.

Thank you for your information. I will also try the Rancher (k3s) environment.

PR #743, opened 2022-05-18
Fix typo in asset-transfer-basic

tommytim0515:


lehors:
Thanks for your contribution. I know this is a bit silly for such a trivial change but all commits must have sign-off for us to be able to merge them. Could you please add one? Thanks.

tommytim0515:
Thanks so much for your prompt reply! I have added the sign-offs.

PR #745, opened 2022-05-18
fix to replace 443 with NGINX_HTTPS_PORT

redblade:
this fix replace hardcoded port 443 with NGINX_HTTPS_PORT which is useful in case ingress is installed on non conventional ports

### DIFF
--- a/test-network-k8s/scripts/chaincode.sh
+++ b/test-network-k8s/scripts/chaincode.sh
@@ -161,7 +161,7 @@ function invoke_chaincode() {
     -n              $cc_name \
     -C              $CHANNEL_NAME \
     -c              $@ \
-    --orderer       org0-orderer1.${DOMAIN}:443 \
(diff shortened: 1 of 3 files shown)


jkneubuh (changes requested):
Thanks @redblade ! 

This is a great PR.  Thanks for submitting it - we had this come up last week for a user that was trying to run a KIND + nginx ingress on a system that ALSO had Rancher running on the host and binding a _different_ ingress at :443 / :80.

Collisions on :80 and :443 are frequent out in the wild - thanks for patching this up.

One small request:  would you please re-submit the commit using the `git commit -s `, or amending the commit such that it passes the failing [DCO check](https://github.com/hyperledger/fabric-samples/pull/745/checks?check_run_id=6490561162) ?

redblade:
Hello @jkneubuh I hope it is now solved

jkneubuh (approved):
👍

PR #749, opened 2022-05-20
Java off-chain data store sample using Fabric Gateway

bestbeforetoday:
Also minor implementation changes to TypeScript sample for better consistency between implementations.

Contributes to hyperledger/fabric-gateway#190


off_chain_data/application-java/app/src/main/java/AssetTransferBasic.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.List;
+
+import com.google.gson.Gson;
+import org.hyperledger.fabric.client.CommitException;
+import org.hyperledger.fabric.client.CommitStatusException;
+import org.hyperledger.fabric.client.Contract;
+import org.hyperledger.fabric.client.EndorseException;
+import org.hyperledger.fabric.client.SubmitException;
+
+public final class AssetTransferBasic {
```
mbwhite, line 17: Should this be called something other than' AssetTransferBasic'?  

To avoid confusion.

bestbeforetoday: It's the client application view of the asset-transfer-basic smart contract so the name seems reasonable to me, and aligns with the naming in the corresponding TypeScript sample:

https://github.com/hyperledger/fabric-samples/blob/4681fe78653ab8a77584b54410538282ad7d0f08/off_chain_data/application-typescript/src/contract.ts#L20

I could change them both if you have a better name for them.

mbwhite (approved):

off_chain_data/application-java/app/src/main/java/AssetTransferBasic.java
mbwhite: no; not got a better name ;-)

so long as it was a conscious decision 👍

bestbeforetoday: Conscious but not necessarily a good one 🤣

Issue #765, opened 2022-06-14
ERC1155: Subtraction Overflow


rajat-dlt:
If a user follows the steps mentioned in the ERC1155, he faces issues with TransferFrom and BatchTransferFrom operations. 

For example - 
If a user runs: 
`peer chaincode invoke "${TARGET_TLS_OPTIONS[@]}" -C mychannel -n erc1155 -c "{\"function\":\"TransferFrom\",\"Args\":[\"$P1\",\"$P2\",\"3\",\"6\"]}" --waitForEvent`
The following error is obtained. 
`Error: endorsement failure during invoke. response: status:500 message:"Math: subtraction overflow occurred  300 - 6"`

This needs updating the [sub()](https://github.com/hyperledger/fabric-samples/blob/main/token-erc-1155/chaincode-go/chaincode/contract.go#L1150) function of the contract.

rajat-dlt:
Working on this issue. 
@denyeart can you please assign this to me.

PR #768, opened 2022-06-16
ERC1155: subtraction overflow issue resolved

rajat-dlt:
This resolves issue #765.
Signed-off-by: Rajat Sharma <connecttorajat@outlook.com>

### DIFF
--- a/token-erc-1155/chaincode-go/chaincode/contract.go
+++ b/token-erc-1155/chaincode-go/chaincode/contract.go
@@ -1151,9 +1151,9 @@ func sub(b uint64, q uint64) (uint64, error) {
 
 	// Check overflow
 	var diff uint64
-	diff = q - b
+	diff = b - q
 
-	if diff > q {
+	if diff > b {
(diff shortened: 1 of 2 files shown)


rajat-dlt:
@denyeart, could you please look into this PR and let me know if any further changes are needed on this.

denyeart:
Thanks @rajat-dlt . Could you also fix the same problem at https://github.com/hyperledger/fabric-samples/blob/main/token-erc-20/chaincode-go/chaincode/token_contract.go#L724 in this PR?

These issues that you are finding all got introduced in https://github.com/hyperledger/fabric-samples/pull/731.  I think that PR tried to do too much in a single commit and therefore these problems were not caught. This is a good reminder that we should do smaller PRs with a single change and re-test each of the chaincodes (Ideally these chaincodes should have CI tests similar to the asset-transfer CI tests).

@fraVlaca I believe you are re-visiting the samples as well... please double check each of them and re-test to make sure they are working per the readmes.

rajat-dlt:
@denyeart, the fix is added for ERC20.

denyeart:
Thanks Rajat!

Issue #770, opened 2022-06-19
Asset transfer private data test case asset_transfer_test.go filing


knarayan:

knarayan:
=== RUN   TestAgreeToTransferSuccessful
2022/06/19 13:18:22 ReadAsset: collection assetCollection, ID id1
2022/06/19 13:18:22 AgreeToTransfer Put: collection Org1TestmspPrivateCollection, ID id1
2022/06/19 13:18:22 AgreeToTransfer Put: collection assetCollection, ID id1, Key transferAgreementid1
    asset_transfer_test.go:215: 
        	Error Trace:	asset_transfer_test.go:215
        	Error:      	Not equal: 
        	            	expected: []byte{0x6d, 0x79, 0x4f, 0x72, 0x67, 0x31, 0x55, 0x73, 0x65, 0x72, 0x69, 0x64}
        	            	actual  : []byte{0x9b, 0x23, 0xab, 0x83, 0x55, 0x2c, 0x7a, 0xb8, 0x9d}
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	-([]uint8) (len=12) {
        	            	- 00000000  6d 79 4f 72 67 31 55 73  65 72 69 64              |myOrg1Userid|
        	            	+([]uint8) (len=9) {
        	            	+ 00000000  9b 23 ab 83 55 2c 7a b8  9d                       |.#..U,z..|
        	            	 }
        	Test:       	TestAgreeToTransferSuccessful
--- FAIL: TestAgreeToTransferSuccessful (0.00s)
=== RUN   TestTransferAssetSuccessful
2022/06/19 13:18:22 TransferAsset: verify asset exists ID id1
2022/06/19 13:18:22 ReadAsset: collection assetCollection, ID id1
    asset_transfer_test.go:264: 
        	Error Trace:	asset_transfer_test.go:264
        	Error:      	Received unexpected error:
        	            	failed transfer verification: error: submitting client identity does not own asset
        	Test:       	TestTransferAssetSuccessful
--- FAIL: TestTransferAssetSuccessful (0.00s)
=== RUN   TestTransferAssetWithoutAnAgreement
2022/06/19 13:18:22 TransferAsset: verify asset exists ID id1
2022/06/19 13:18:22 ReadAsset: collection assetCollection, ID id1
    asset_transfer_test.go:328: 
        	Error Trace:	asset_transfer_test.go:328
[...]

PR #771, opened 2022-06-19
GetID to return clientId in base64 form

knarayan:
Signed-off-by: Krishnasuri Narayanam <knaraya3@in.ibm.com>


mbwhite (commented):
Hello - could I check why the cast here... the clientIdentity interface returns a string already does it not?

https://github.com/hyperledger/fabric-chaincode-go/blob/7bc6fa0dd58be10ef9fa2a0452355b855c7bcad3/pkg/cid/interfaces.go#L23

lehors:
@knarayan Could you please followup and respond to @mbwhite ? Thanks.

knarayan:
@mbwhite Since GetID is returning base64 encoded string ( https://github.com/hyperledger/fabric-chaincode-go/blob/7bc6fa0dd58be10ef9fa2a0452355b855c7bcad3/pkg/cid/cid.go#L106 ), the test case also expects the same to be returned by the mocking function `GetIDReturns`. Hence the change. Without this the test cases were failing and here is the log for your reference #770 .

lehors (dismissed):
Thank you for your PR and sorry for the long delay merging it.
I tested it and you are correct.

PR #772, opened 2022-06-20
added last fixes for asset-transfer-secured-agreement

fraVlaca:
Included check for the client org id to match the peer org id for functions that requires it.

### DIFF
--- a/asset-transfer-secured-agreement/application-javascript/app.js
+++ b/asset-transfer-secured-agreement/application-javascript/app.js
@@ -301,7 +301,7 @@ async function main() {
 			try {
 				// Agree to a sell by Org1
 				const asset_price = {
-					asset_id: assetKey,
+					asset_id: assetKey.toString(),
 					price: 110,
 					trade_id: randomNumber.toString()
 				};
@@ -402,7 +402,7 @@ async function main() {
 				// This will fail due to the sell price and the bid price
 				// are not the same
 				const asset_price = {
-					asset_id: assetKey,
+					asset_id: assetKey.toString(),
 					price: 110,
 					trade_id: randomNumber.toString()
 				};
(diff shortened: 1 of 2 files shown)


jkneubuh (changes requested):
The code looks OK but could you please make one update?

Instead of passing boolean values to functions, please consider introducing a new method  that makes the behavior clear from the function name.  This is a good practice that will help consumers of the routine(s) immediately understand the context of the boolean flag, without needing to drop into the function being called or documentation to infer the context.

In this case the `getClientOrgID(ctx)` function could stay "as is," and introduce a new function that additionally runs the verification as: 

```
func getVerifiedClientOrgID(ctx contractapi.TransactionContextInterface, verifyOrg bool) (string, error) {
 	clientOrgID, err := ctx.GetClientIdentity().GetMSPID()
 	if err != nil {
 		return "", fmt.Errorf("failed getting client's orgID: %v", err)
 	}

	err = verifyClientOrgMatchesPeerOrg(clientOrgID)
	if err != nil {
		return "", err
	}

 	return clientOrgID, nil
 }
```

As a general rule, when refactoring or writing new code, always prefer adding a new function or method over introducing a boolean flag into the parameter list of an existing function.  While this isn't always possible, it's a good technique to help keep bugs from creeping in when the flags are passed incorrectly.

fraVlaca:
Ok thanks,  yep I think I usually would have done something like that. 
However, this was a feature that was already previously in that I removed but then understood that I needed to reintroduce and rewrote it as it was previously. Do you think I should take the occasion to refactor the code or leave it like this?

denyeart:
@fraVlaca I agree with Josh... while you have your fingers in the code, it is always a good idea to make any improvements that you find. Go for it!

jkneubuh:
The earlier comment is actually a good example of why NOT to introduce boolean parameters into functions.   I just read Enyeart's comments above and noticed that I had copy/pasted the function signature, including the boolean parameter, into an example of how not to include boolean parameters into function signatures.  Nice.

Usually when I am bitten by bugs introduced with boolean parameters, it's when copy/pasting function calls around.  Apparently that carries over into GitHub comments, as well.

- Go for it.
- Stamp out any and all boolean parameters that you ever find in code.  
- Write unit tests.
- Consider yourself fortunate that Haskell is not supported.

fraVlaca:
Changed it, I did test it by myself and it works fine; let me know if I need to write any unit tests since there are not any for this sample and I would like to get on other things and maybe come back later if not necessary.

PR #773, opened 2022-06-21
token samples: added fixes to documentation and readmes 

### LINKED ISSUE #762: token-erc-721 and token-utxo samples are not working per Readme
https://github.com/hyperledger/fabric-samples/tree/main/token-erc-721 sample is not working per the readme.

The Readme has the user invoke Initialize with args:
"Args":["some name", "some symbol", "2"]

However the chaincode function only takes two parameters (no decimal parameter).

I then tried with "Args":["some name", "some symbol"]. This time the chaincode function is invoked successfully, but it returns an error "Error: contract options are already set, client is not authorized to change them", even though it hasn't been initialized yet. This works for the Go chaincode but not the Javascript chaincode, so I suspect the check at https://github.com/hyperledger/fabric-samples/blob/main/token-erc-721/chaincode-javascript/lib/tokenERC721.js#L315-L317 is not correct.

It looks like the same issue will apply to the token-utxo readme (readme has three parameters but chaincode only has two parameters). I assume this was just a copy/paste issue in the Readme.

fraVlaca:
tokens docs fix update + small fix to erc721 and erc 20 java chaincode gradle.settings
Closes #762 

### DIFF
--- a/token-erc-1155/README.md
+++ b/token-erc-1155/README.md
@@ -148,7 +148,7 @@ export CORE_PEER_ADDRESS=localhost:7051
 
 We can then invoke the smart contract to initilize it
 ```bash
-peer chaincode invoke "${TARGET_TLS_OPTIONS[@]}" -C mychannel -n erc1155 -c '{"function":"Initialize","Args":["some name", "some symbol", "2"]}'
+peer chaincode invoke "${TARGET_TLS_OPTIONS[@]}" -C mychannel -n erc1155 -c '{"function":"Initialize","Args":["some name", "some symbol"]}'
 ```
 
 ### Get account ID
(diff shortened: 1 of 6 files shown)


jkneubuh (approved):
This is a nice addition to the documentation / READMEs.  A couple of questions: 

1.  Why is the Java contract returning a different spec / format for the Ids?  Shouldn't the contracts have the same input / output specs, independent of the runtime language?

2.  READMEs are good, but integration test scripts are better.

Consider combining these two arcs, and write a test suite in some high-level scripting language that exercises the overall flow.  Drive the test script with a test matrix, varying the chaincode implementation language.

fraVlaca:
Very good observations thanks, I think it is important to have some integrations test to make it easier to very that the samples works and make sure that new bugs are not introduced with new changes.
I do not know why the format is different between java and javascript ids. In this case, the data were still the same but java was using a comma separator instead of the slash. Also, go is different as it does use bytes instead of a string.
However, if you decode it decodes to a comma-separated list. SO it actually looks like the javascript one is the one that differs by using slashes

PR #775, opened 2022-06-22
Readme updated for token samples

rajat-dlt:
Signed-off-by: Rajat Sharma <connecttorajat@outlook.com>


rajat-dlt:
@denyeart @fraVlaca,

The following changes are made in this - 

- ERC721- If a user copies and pastes the command then an error will be obtained.
- ERC1155- The environment variable had to be updated based on the sample.

denyeart (approved):

PR #776, opened 2022-06-24
Fix test-network-k8s docs

satoshi-ito-ht:
This patch fixes errors in the document.
Fix errors as follows:

- Change shell script arguments
- Correct display results

Signed-off-by: Satoshi Ito <satoshi.ito.tf@hitachi.com>


satoshi-ito-ht:
@jkneubuh 

Thanks for the review and the suggestion!
I will try fabric-operator as soon as possible.

PR #784, opened 2022-07-04
Fix namespace in k8s configuration files

destinysky:
Issues:
1. When the name of NETWORK_NAME or NS in file "network" (i.e., the namespace in k8s) is changed from "test-network", the resources in the k8s cluster cannot be created successfully when the command "./network up" is executed.
Error message: error: the namespace from the provided object "test-network" does not match the namespace "xxxxxx". You must pass '--namespace=test-network' to perform this operation.
2. org2-cc.yaml is not completed.

Fix:
1. Replace "test-network" in yaml files in folder "kube" with "${NS}".
2. Complete "org2-cc-template.yaml" by referring to "org1-cc-template.yaml".

### DIFF
--- a/test-network-k8s/config/org0/fabric-ca-server-config.yaml
+++ b/test-network-k8s/config/org0/fabric-ca-server-config.yaml
@@ -321,7 +321,7 @@ csr:
     - localhost
     - 127.0.0.1
     - org0-ca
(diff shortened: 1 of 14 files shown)


jkneubuh (approved):
Hi @destinysky 

Thanks for the PR.  This looks like a good update.

PR #797, opened 2022-07-18
fix to prometheus-grafana docker compose

fraVlaca:
Due to an update in docker-compose, the networking properties and format had changed and broke the docker-compose file for the Prometheus and Grafana network, this small pr fixes it.


davidkel:
It doesn't look like compose has changed the format of external networks (see https://docs.docker.com/compose/networking/) and even if they did I would expect it to be version controlled so I don't think this change is valid

PR #803, opened 2022-07-29
Update podman support information

lehors:
This change updates the test-network README with up-to-date information
about using podman on Mac and improves support for podman in deployCCAAS.sh


nikhil550 (commented):
I would be explicit on saying that you should set the environment variables before running the command, instead of while running.

denyeart (approved):

PR #806, opened 2022-08-03
Add getCreator() to parsed transaction in off_chain_data sample

bestbeforetoday:
Example of how the client identity that submitted a transaction can be obtained.


mbwhite (approved):
LGTM

Issue #644, opened 2022-02-15
Kube test network : create sample example to deploy in multiple namespaces (each org running in seperate namespace)


knagware9:
This is really great example to test K8s deployment, it must have a sample deployment doc to deploy each org, Ordering service running in a separate namespace so this sample could be useful for people who like to use these samples for production deployment.

jkneubuh:
This is a great feature.  

I've actually used the K8s test network on a few occasions to help troubleshoot and diagnose errors that come up when connecting Fabric nodes across k8s namespaces.  (E.g. see [this thread on stack overflow](https://stackoverflow.com/questions/69133882/whats-the-correct-connection-profile-for-accessing-hyperledger-fabric-v2-x-depl/69179352#69179352))  Probably the most confusing part of this was related to setting up the TLS host names in the CSR/SAN aliases when running the node enrollments with the Fabric TLS CA.   That said: all the mechanics are "in place" to support running a test network in multiple namespaces.  The cert-manager.io `Issuer` also makes the setup of TLS SAN host aliases _really easy_.

I also like how the use of multiple namespaces (PVC, etc.) *force* the user to plan for the distribution of MSP contexts between organizations.  Running a network across namespaces will really help to illuminate the "cross organizational" nature of Fabric.

The one drawback of separating the orgs across namespaces is that it does distract a little bit from the convenience of seeing everything together in a single bucket.   Is there a way we could set this up such that the scripts are driven by a parameter for each org's namespace, with the default being a single namespace?

The changes for this feature should be straightforward, but touch a lot of places in the scripts: 

- Introduce new env parameters for the org namespaces (e.g. `${TEST_NETWORK_ORG1_NAMESPACE:-${NS}}`

- Extend the node [TLS cert requests](https://github.com/hyperledger/fabric-samples/blob/main/test-network-k8s/kube/org1/org1-peer1.yaml#L21) with Kube DNS hostnames (e.g. `<service>.<namespace>.svc.cluster.local`)

- Update all of the service URLs to use the Kube DNS hostname. 


On a larger arc, one place the Kube test network still falls short is that we don't have a good example of adding a new organization to the network to the blockchain after it's been set up with the two peer orgs.

basilky:
@knagware9 @jkneubuh Drafted a PR with required changes for this.
https://github.com/hyperledger/fabric-samples/pull/817

PR #817, opened 2022-08-22
separate namespace for each organization

basilky:
Currently all the organizations are in a single namespace. Changes are done in multiple files to create three namespaces and to separate organizations across namespaces.
- ORG0_NS, ORG1_NS, ORG3_NS are defined in network script
- $NS in the scripts are replaced with corresponding namespace

All till chaincode invoke/query is working. I'm yet to test the api server. kind was used for n/w setup, will test on Rancher also.

fixes https://github.com/hyperledger/fabric-samples/issues/644.

Signed-off-by: Basil K Y <techiebasil@gmail.com>

### DIFF
--- a/test-network-k8s/kube/fabric-rest-sample.yaml
+++ b/test-network-k8s/kube/fabric-rest-sample.yaml
@@ -11,7 +11,7 @@ metadata:
 data:
   HLF_CONNECTION_PROFILE_ORG1: |
     {
-        "name": "${NS}-org1",
+        "name": "Org1",
         "version": "1.0.0",
         "client": {
             "organization": "Org1",
@@ -36,14 +36,14 @@ data:
         },
         "peers": {
             "org1-peers": {
-                "url": "grpcs://org1-peer-gateway-svc:7051",
+                "url": "grpcs://org1-peer1.${ORG1_NS}.svc.cluster.local:7051",
                 "tlsCACerts": {
                     "pem": "-----BEGIN CERTIFICATE-----\\nMIICvzCCAmWgAwIBAgIULJGws7jbEY6ruSgDuvi9L7VphvIwCgYIKoZIzj0EAwIw\\naDELMAkGA1UEBhMCVVMxFzAVBgNVBAgTDk5vcnRoIENhcm9saW5hMRQwEgYDVQQK\\nEwtIeXBlcmxlZGdlcjEPMA0GA1UECxMGRmFicmljMRkwFwYDVQQDExBmYWJyaWMt\\nY2Etc2VydmVyMB4XDTIxMDkyMDE2MDkwMFoXDTIyMDkyMDE2MTQwMFowYDELMAkG\\nA1UEBhMCVVMxFzAVBgNVBAgTDk5vcnRoIENhcm9saW5hMRQwEgYDVQQKEwtIeXBl\\ncmxlZGdlcjENMAsGA1UECxMEcGVlcjETMBEGA1UEAxMKb3JnMS1wZWVyMTBZMBMG\\nByqGSM49AgEGCCqGSM49AwEHA0IABL9e3GZBf1MeoObGxwSHkcgDEjMo+/13Qc4u\\nfSG2MKrveHBIEA4MRkHNqd+sTjoz0/1B15y2n+RiPo8uJvlyC/CjgfQwgfEwDgYD\\nVR0PAQH/BAQDAgOoMB0GA1UdJQQWMBQGCCsGAQUFBwMBBggrBgEFBQcDAjAMBgNV\\nHRMBAf8EAjAAMB0GA1UdDgQWBBSeytspiXlEzMAsnF9/wxqc9fydETAfBgNVHSME\\nGDAWgBQwru1VH0OwH3dxfPdD8w74ZIlLRzAVBgNVHREEDjAMggpvcmcxLXBlZXIx\\nMFsGCCoDBAUGBwgBBE97ImF0dHJzIjp7ImhmLkFmZmlsaWF0aW9uIjoiIiwiaGYu\\nRW5yb2xsbWVudElEIjoib3JnMS1wZWVyMSIsImhmLlR5cGUiOiJwZWVyIn19MAoG\\nCCqGSM49BAMCA0gAMEUCIQDJEjPxceCfXU5B/emrHE4JbEzrZKxLVViBWCNMsHiR\\nFgIgY+8jsvr3rlBPkpRhl8CtT2DgaP7iWvovtMYsPKhLAqk=\\n-----END CERTIFICATE-----\\n"
                   },
                 "grpcOptions": {
                     "grpc-wait-for-ready-timeout": 100000,
-                    "ssl-target-name-override": "org1-peer-gateway-svc",
-                    "hostnameOverride": "org1-peer-gateway-svc"
+                    "ssl-target-name-override": "org1-peer1.${ORG1_NS}.svc.cluster.local",
+                    "hostnameOverride": "org1-peer1.${ORG1_NS}.svc.cluster.local"
                 }
             }
         },
@@ -87,7 +87,7 @@ data:
     -----END PRIVATE KEY-----
   HLF_CONNECTION_PROFILE_ORG2: |
     {
-        "name": "${NS}-org2",
+        "name": "Org2",
         "version": "1.0.0",
         "client": {
             "organization": "Org2",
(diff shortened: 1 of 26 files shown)


jkneubuh:
Wow.  This is great.  Thank you, @basilky ! 

A couple of questions / comments: 

- This is great. Thank you for the contribution. 

- Are the ci/ test script failures related to the PR?  Sometimes the tests are "flaky" at AZP.  ( running `fabric-samples/ci/scripts/run-k8s-test-network-basic.sh` locally is a good way to suss out errors from CI ) 

- The Kube Test Network is showing some ... "age."   It is good for a reference for deploying to k8s, or for illustrating different techniques and testing integrations (e.g. cert-manager.io or linkerd.io for a mesh).    Have you seen the new fabric-operator and associated [sample-network](https://github.com/hyperledger-labs/fabric-operator/tree/main/sample-network) ?    It is similar to the kube test network, but ... better in some ways.


Note I am not suggesting this PR shouldn't be merged.  Only that it's worth taking a look at fabric-operator, which is "production grade," following similar patterns as exposed by the kube test network.

jkneubuh (changes requested):
I think this is a really great PR.  Thanks again for moving it forward.   A couple of asks / requests :

1.  Patch up the AZP test runs.  (Not clear if this is an AZP test flake, or possibly related to the update to span multiple namespaces.)   The k8s tests under "/ci" should be enough to drive the overall flow. 

2.  Why was the change of working directory for the channel config necessary?  (Moving ../../build/channel-msp --> ./channel-msp)    I recall arguing with this area of the code / setup, and wasn't very happy embedding the working directory for the channel config block.

3.  Can we leave the existing behavior of deploying to a single namespace as the "default" behavior (e.g. -> `test-network`) and then use this PR + a guide to instruct on deployment of the different orgs to multiple namespaces?


This looks great.

basilky:
> I think this is a really great PR. Thanks again for moving it forward. A couple of asks / requests :
> 
>     1. Patch up the AZP test runs.  (Not clear if this is an AZP test flake, or possibly related to the update to span multiple namespaces.)   The k8s tests under "/ci" should be enough to drive the overall flow.
> 
>     2. Why was the change of working directory for the channel config necessary?  (Moving ../../build/channel-msp --> ./channel-msp)    I recall arguing with this area of the code / setup, and wasn't very happy embedding the working directory for the channel config block.
> 
>     3. Can we leave the existing behavior of deploying to a single namespace as the "default" behavior (e.g. -> `test-network`) and then use this PR + a guide to instruct on deployment of the different orgs to multiple namespaces?
> 
> 
> This looks great.

@jkneubuh Thanks for the review and suggestions.

1. I ran the CI script multiple times locally and the results are inconsistent. Sometimes it fails due to kubectl timeout, sometimes due to ssl handshake error, sometimes docker deamon failing to kill kind cluster etc. Will try to make it work.

2. I had added the namespace variable into the ordering and peers endpoints in configtx.yaml(and renamed configtx.yaml to configtx-template.yaml). `envsubst` is used to generate actual configtx.yaml from this template. The generated configtx.yaml is saved to the build directory. So I changed "../../build/" to "./" as configtx.yaml is already inside build directory. 

3. I will change all three namespace variables to $NS so that all organizations will be in a single namespace by default.

jkneubuh:
Thanks @basilky .   Holler if you are stuck on the test suites.  I agree they are ... "brittle."  (More reason / incentive to use the fabric-operator, IMO! ) 

Thanks for moving the configtx to a template / etc.  I understand the update now.  I think there may still be some lingering overlap between what runs "locally" vs. running in a clean sandbox, as is the case in the CI flow when everything starts up from scratch.

Also, I think there may be a couple of lingering issues where multiple "network up" calls will re-use the certificates and enrollments generated from a previous run of the network.   For instance, if you delete a kind cluster with the KIND cli, this will not scrub the old certs, and the next network to spin up will be doomed from the start.

I _think_ there may be a case in here where there is still some overlap between the configtx.yaml downloaded when the peer binaries are installed by the network script, and the template substitutions above.  Holler if you want another pair of eyes to look at the test scripts - I agree these are very ... fortunate to be passing at all.   Lots of room for improvement in here.

Thanks again for the PR.

test-network-k8s/kube/fabric-rest-sample.yaml
```diff
@@ -112,13 +112,13 @@ data:
...
                 "tlsCACerts": {
                     "pem": "-----BEGIN CERTIFICATE-----\\nMIICKDCCAc6gAwIBAgIUJJ4wGOSCfw8XOOIx29o67wBpFB4wCgYIKoZIzj0EAwIw\\naDELMAkGA1UEBhMCVVMxFzAVBgNVBAgTDk5vcnRoIENhcm9saW5hMRQwEgYDVQQK\\nEwtIeXBlcmxlZGdlcjEPMA0GA1UECxMGRmFicmljMRkwFwYDVQQDExBmYWJyaWMt\\nY2Etc2VydmVyMB4XDTIxMDkyMDExNDEwMFoXDTM2MDkxNjExNDEwMFowaDELMAkG\\nA1UEBhMCVVMxFzAVBgNVBAgTDk5vcnRoIENhcm9saW5hMRQwEgYDVQQKEwtIeXBl\\ncmxlZGdlcjEPMA0GA1UECxMGRmFicmljMRkwFwYDVQQDExBmYWJyaWMtY2Etc2Vy\\ndmVyMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEyzGJLZX6pe59QAIBacjfzU4I\\nHezBYLyEu4ySpFx4xwxNLE4BWqLhB1VaOuenSQATM8pmSAy7i1830oM9elKWK6NW\\nMFQwDgYDVR0PAQH/BAQDAgEGMBIGA1UdEwEB/wQIMAYBAf8CAQEwHQYDVR0OBBYE\\nFEoAAhmjq/3M8CFPc7N8SL53erL5MA8GA1UdEQQIMAaHBH8AAAEwCgYIKoZIzj0E\\nAwIDSAAwRQIhAJQ5PJOT4Gg8oiBU2KthMPkZqOLeu3Li4S3yBpLFgbsgAiB960P2\\nXPMu3HLoNXrktYOL9JzWlGyYRSPAnkap5Bsj0w==\\n-----END CERTIFICATE-----\\n"
                 },
                 "grpcOptions": {
-                    "ssl-target-name-override": "org2-peer-gateway-svc",
-                    "hostnameOverride": "org2-peer-gateway-svc"
+                    "ssl-target-name-override": "org2-peer1.${ORG2_NS}.svc.cluster.local",
```
jkneubuh, line 120: I think this is a "bug" but is OK in this context.  It's not worth holding up this PR for a fix.

When connecting with the new Gateway Client, the application can connect to ANY of the peers in the org, and the gateway peer will act as a proxy or tunnel to the actual peer instance that is invoking the transactions on the ledger.   In this case, we set up a Kube `Service`  - `org1-peer-gateway-svc` to distribute client connections across the peers in the org.    The URLs here will work fine, but can probably be improved through access via the Gateway / virtual URLs.  Good examples of this feature in action: 

- [Fabric Gateway](https://github.com/hyperledgendary/full-stack-asset-transfer-guide/blob/main/docs/ApplicationDev/01-FabricGateway.md)

- [Gateway Load Balancing](https://github.com/hyperledgendary/full-stack-asset-transfer-guide/blob/main/docs/CloudReady/40-bananas.md#gateway-load-balancing)

jkneubuh:
This PR looks great.  @basilky -- I'm more than happy to land this when you are satisfied and switch the PR out of a DRAFT state.  

👍

test-network-k8s/kube/fabric-rest-sample.yaml
basilky: @jkneubuh I have reverted this change and the apis are working

basilky:
@jkneubuh I think this PR is ready to merge. Let me know if any changes are needed. btw, the fabric-operator is new to me. I will definitely have a look at it.

jkneubuh:
Thanks @basilky ! 

Start with an introduction to the operator at https://github.com/hyperledgendary/full-stack-asset-transfer-guide 

The guide is still a little light on words / figures / explanatory text.  Paired with the operator sample-network I think there is enough here to convince you to stop improving the Kube Test Network.  : ) 

The "Cloud Native Fabric" section has the most graceful introduction, IMO.   It's "all here" - minus the cert-manager.io integration -- fabric-operator is a much, much, much better path for deploying production-ready Fabric networks.


Thank you for the great PR.  Cheers.

jkneubuh (approved):

jkneubuh (approved):

PR #818, opened 2022-08-24
Cloudathon

aneran01:


AkhzarFarhan (commented):
LGTM

denyeart:
Thank you for the submission. The intent of the fabric-samples repository is to support the official Fabric documentation tutorials with minimal duplication of application patterns. Since this chaincode is similar in overall pattern to the pre-existing asset-transfer-basic java chaincode, I don't think a second sample is needed. If you'd like to improve the existing application patterns, I'd suggest to enhance the asset transfer samples and related tutorial documentation rather than introducing a new sample.

PR #826, opened 2022-09-08
Sample application to add a golang version to commercial-paper

767829413:
Signed-off-by: Fang Yuan <wojiushifangyuanlove@gmail.com>


commercial-paper/organization/digibank/application-go/business/addtowallet.go
```diff
@@ -0,0 +1,95 @@
+package business
+
+import (
+	"io/ioutil"
```
bestbeforetoday, line 4: I'd suggest avoiding **ioutil** as it has been deprecated since Go 1.16. All the functionality in **ioutil** is available in either the **io** or **os** packages

bestbeforetoday (commented):
I really like the contribution of a Go implementation of this sample. However...

I wonder if this would be a good opportunity to implement commercial-paper using the new [Fabric Gateway client API](https://github.com/hyperledger/fabric-gateway) instead of legacy Go SDK. Samples like **asset-transfer-basic** demonstrate the new API usage, and the [API documentation](https://pkg.go.dev/github.com/hyperledger/fabric-gateway/pkg/client) contains lots of examples too.

commercial-paper/organization/digibank/application-go/business/addtowallet.go
767829413: Thank you very much for your reply, this is indeed a problem that deserves attention!

767829413:
> 



> I really like the contribution of a Go implementation of this sample. However...
> 
> I wonder if this would be a good opportunity to implement commercial-paper using the new [Fabric Gateway client API](https://github.com/hyperledger/fabric-gateway) instead of legacy Go SDK. Samples like **asset-transfer-basic** demonstrate the new API usage, and the [API documentation](https://pkg.go.dev/github.com/hyperledger/fabric-gateway/pkg/client) contains lots of examples too.

In China, due to policy intervention, using the GO SDK is the safer way to go.

mbwhite:
@767829413 could expand a little on why the Fabric Gateway isn't safer?

Of course you must choose the correct solution for your needs, but the GO SDK isn't maintained by the set of maintainers working on the main Fabric repo/Chaincode/Gateway SDKs.

PR #836, opened 2022-09-23
REST api added for asset transfer in Golang

basilky:
This PR adds a REST server in Golang with endpoints for chaincode invoke and query. I I'm not an expert in Golang. Let me know if any improvements required.


jt-nti:
Hi @basilky, thanks for the contribution! It's really good to get a sample using the new Fabric Gateway!

It looks like your REST api is actually a generic server which could run transactions for any sample, rather than being specifically related to the asset transfer sample- is that right? For contrast, there is a [more chaincode specific Go REST sample in the recent full stack workshop repository](https://github.com/hyperledgendary/full-stack-asset-transfer-guide/tree/main/applications/rest-api).

I've actually wanted a more generic REST api for a while, and neither your server or the full stack sample are equivalent to the `rest-api-typescript` sample, which has more focus on errors and retries, so I wonder if moving this code out of `asset-transfer-basic` would be better, e.g. `rest-client-go` or something at the root of the repository? Hopefully one of the samples maintainers can comment?

As for the Go code, I'm not an expert but it looks like you're reusing the GRPC connection/Gateway which is great. I would personally like to see more configuration done through env vars rather than being hard coded to the `test-network` directories but I think it would be fine to merge what you have so far and iterate on it.

Thanks again.

basilky:
@jt-nti Thanks for the feedback. The REST server is written in a generic way. The channel name, chaincode name, function and arguments are received from the request. I also think it's better to move the code out of asset-transfer.

lehors (approved):
I'm happy to merge this as is for now. Followup PRs to address @jt-nti 's comments would be welcome.
Thank you for your contribution!

PR #843, opened 2022-10-21
Avoid usage of deprecated ioutil Go package

bestbeforetoday:
1. replace deprecated ioutil functions (ioutil is deprecated since Go 1.16)
2. fix variable names that collide with imported package name
3. fix typos

Also update Go version specified by Go modules to ensure a Go version is used in which ioutil is deprecated and replacement functions are available in os and io packages.

Signed-off-by: Tommy TIAN <xtianae@connect.ust.hk>
Co-authored-by: Mark S. Lewis <mark_lewis@uk.ibm.com>


asset-transfer-basic/application-go/go.mod
```diff
@@ -1,5 +1,49 @@
 module asset-transfer-basic
 
-go 1.14
+go 1.19
```
jkneubuh, line 3: 1.18 or 1.19?

jkneubuh (commented):
@bestbeforetoday this looks like a good update.   The only question I had about this PR was the use of golang 1.19 for the application-go.  Should this be pinned to 1.18?

bestbeforetoday:
> @bestbeforetoday this looks like a good update. The only question I had about this PR was the use of golang 1.19 for the application-go. Should this be pinned to 1.18?

I just picked 1.19 since it's the current Go version. If you prefer to ensure the application samples can be run using Go 1.18 then I can change the Go version and re-tidy the dependencies.

bestbeforetoday:
@jkneubuh I've downgraded the go.mod files from go 1.19 to go 1.18, as suggeted

jkneubuh (approved):
👍

PR #845, opened 2022-10-25
minor chaincode cleanup

manonthemat:


lehors (changes requested):
Your changes look good to me but you need to add a sign-off to your commit so that we can merge them.
Thank you.

PR #862, opened 2022-11-11
Move lint and test-network-basic tests from Azure to GHA - READY FOR MERGE

jkneubuh:
This PR ports the _glacial_ Linter and Test Network Basic tests from Azure to GHA: 

- Test Network Basic - Go
- Test Network Basic - Java
- Test Network Basic - Javascript
- Test Network Basic - Typescript

In addition, the scripts have been updated to run each test case in a different chaincode scope / ledger, re-using a single Fabric network across all tests in the suite.


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - release-1.4
+      - release-2.2
+  pull_request:
+    branches:
+      - main
+      - release-1.4
+      - release-2.2
+
+env:
+  GO_VER: 1.18.3
```
mbwhite, line 21: These aren't the levels of environment for all releases however...

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,68 @@
...
+  NODE_VER: 16.x
+  JAVA_VER: 11.x
+
+jobs:
+  go:
+    runs-on: ubuntu-latest
```
mbwhite, line 27: Wonder if this should be fixed at the lowest level of os that we support.. namely 20.04... 
Avoid the issue we saw earlier where the wrong level c runtime was imported

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VER }}
+      - run: ci/scripts/lint-javascript.sh
+
+  java:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-java@v3
```
mbwhite, line 58: worth specifing which Java run time...

.github/workflows/test-network-basic.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+    steps:
+      - run: echo "🎉 The job was automatically triggered by a ${{ github.event_name }} event."
+
+      - name: checkout
+        uses: actions/checkout@v3
+
+      - name: Install fabric CLI
+        run: |
+          curl -sSL https://raw.githubusercontent.com/hyperledger/fabric/main/scripts/install-fabric.sh | bash -s -- binary
```
mbwhite, line 37: think the version flag will be needed here for different branches

.github/workflows/test-network-basic.yaml
```diff
@@ -0,0 +1,51 @@
...
+          echo ${PWD}/bin >> $GITHUB_PATH
+
+      - name: Install retry CLI
+        run: curl -sSL https://raw.githubusercontent.com/kadwanev/retry/master/retry -o ./bin/retry && chmod +x ./bin/retry
+
+      - name: Pull Fabric Docker Images
+        run: ./ci/scripts/pullFabricImages.sh
```
mbwhite, line 44: any reason for a different script - `install-fabric.sh` can also do docker images?

asset-transfer-basic/application-gateway-java/src/main/java/App.java
```diff
@@ -34,22 +34,22 @@
...
-	private static final String mspID = "Org1MSP";
-	private static final String channelName = "mychannel";
-	private static final String chaincodeName = "basic";
+	private static final String MSP_ID = System.getenv().getOrDefault("MSP_ID", "Org1MSP");
+	private static final String CHANNEL_NAME = System.getenv().getOrDefault("CHANNEL_NAME", "mychannel");
+	private static final String CHAINCODE_NAME = System.getenv().getOrDefault("CHAINCODE_NAME", "basic");
 
 	// Path to crypto materials.
-	private static final Path cryptoPath = Paths.get("..", "..", "test-network", "organizations", "peerOrganizations", "org1.example.com");
+	private static final Path CRYPTO_PATH = Paths.get("../../test-network/organizations/peerOrganizations/org1.example.com");
```
mbwhite, line 42: My years of working on systems running on windows and UNIX makes me twitch when I see I hard coded path separator :-)

Though perhaps it's not really a problem anymore?

ci/scripts/lint-java.sh
```diff
@@ -0,0 +1,24 @@
...
+}
+
+dirs=("$(find . -name "*-java" -type d -not -path '*/.*')")
+for dir in $dirs; do
+  print "Linting $dir"
+  pushd $dir
+
+  if [[ -f "pom.xml" ]]; then
+    print "Running Maven Build"
+    mvn clean package
```
mbwhite, line 18: worth updating to the use the maven wrapper approach as well

.github/workflows/lint.yaml
jkneubuh: I think this is OK?

```
  java:
..
      - uses: actions/setup-java@v3
        with:
          distribution: temurin
          java-version: ${{ env.JAVA_VER }}
```

.github/workflows/test-network-basic.yaml
jkneubuh: The docker image pulls are not right - was going to hit this in another PR but it can go in here as well.

asset-transfer-basic/application-gateway-java/src/main/java/App.java
jkneubuh: No problem, provided path sep is '/' not '\'.   _idiomatic Java_ is to use UPCASE for constants and '/' for path separators.

ci/scripts/lint-java.sh
jkneubuh: Better idea:   send `mvn` to the bin entirely...

jkneubuh:
@mbwhite thanks for the feedback.  Latest push will : 

- pin ubuntu-20.04 
- remove the release branch targets (release lines will still run azp.)   In general I think it is preferable to keep things simple and just run exactly the workflows that are declared on a branch, rather than overriding the env from the GHA context.  Nice and simple.
- Installs fabric docker images from install-fabric (This is better anyways, as the older ones are still pulling from jfrog)

denyeart (approved):

mbwhite (approved):

PR #882, opened 2022-12-07
Update org1-peer2.yaml

blockguardian:
Fixing TLS issues while using org1-peer-gateway-svc.${DOMAIN}. This would be needed so there are no TLS issues when sending requests.

Signed-off-by: Deepak Singh <91736795+blockguardian@users.noreply.github.com>


blockguardian:
@denyeart  can this be taken up  for review ?

jkneubuh (approved):

PR #884, opened 2022-12-13
added a check for prereq into deploycc script

Param-S:
Added a check for prerequisites into deployCC script.  
Issue: https://github.com/hyperledger/fabric/issues/3848

Signed-off-by: Parameswaran Selvam <parselva@in.ibm.com>


lehors (approved):
I must admit not to be sure how one can get that far without jq but I assume you ran into that situation somehow and I don't see any problems adding this check.
Thank you.

PR #902, opened 2023-01-09
Update the name of the certificate.

mbwhite:
Late 2022 it was discovered that when used via the CA, the certificate had a fixed named covention. cert.pem


jkneubuh (approved):
looks good.  Have verified locally as a fix, as well.

PR #908, opened 2023-01-18
Swap to use the Helper Function to create policies

mbwhite:
The old way used the direct protobuf functions, there is a helper function in the chaincode does exactly that.


jkneubuh (approved):
👍

PR #922, opened 2023-01-31
AssetTransfer Private - Incorrect Endorsement Policy

mbwhite:
For the asset transfer example (https://hyperledger-fabric.readthedocs.io/en/release-2.5/private_data_tutorial.html#pd-use-case)  there are three private data collections, one per org and one shared between the orgs.

The shared collection didn't have an endorsement policy so inherited the chaincodes; this was specifically set away from the default to be OR(Org1MSP,Org2MSP).   The documentation says this is to ensure that either organization can create an asset. Whilst this is strictly correct, it is misleading - it doesn't need to be a chaincode wide policy, but specifically to the shared collection. 

Therefore it is better to leave the chaincode policy at its default and move the specific policy down to the collection. 
With the context of this example as coded, this is exactly functionally the same - but gives a more accurate impression.

### DIFF
--- a/asset-transfer-private-data/chaincode-go/collections_config.json
+++ b/asset-transfer-private-data/chaincode-go/collections_config.json
@@ -6,7 +6,10 @@
    "maxPeerCount": 1,
    "blockToLive":1000000,
    "memberOnlyRead": true,
-   "memberOnlyWrite": true
+   "memberOnlyWrite": true,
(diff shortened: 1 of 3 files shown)


denyeart:
This is a good change - fix is coming to Fabric:
https://github.com/hyperledger/fabric/pull/3986

denyeart (approved):

PR #924, opened 2023-02-09
Move fabric-kube-test-network from Hyperledgendary -> samples

jkneubuh:
Configure a test-network on a local Kubernetes cluster, highlighting the enrollment, distribution, and exchange of channel MSP certificates in a production-style, multi-org environment.  The use of CA, ingress, and Fabric CLIs is illustrated through simplistic bash scripting and clear definitions of the org / MSP context.

### DIFF
--- a/test-network-k8s-operator/justfile
+++ b/test-network-k8s-operator/justfile
@@ -1,0 +1,189 @@
+#
+# Copyright contributors to the Hyperledgendary Full Stack Asset Transfer project
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at:
+#
+# 	  http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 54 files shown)


mbwhite (changes requested):
1) Having an operator based example is very useful 👍 
2) The focus on the different organizations doing working separately is also essential. 👍 

Though; I'm not going to suggest changes this example really should be connected with the actual fabric-operator rather than fabric-samples.

I've no problems in the production deployment sections of the main Fabric docs ( that I'm currently updated) reference this sample, likewise the other community deployment technologies.

test-network-k8s-operator/justfile
```diff
@@ -0,0 +1,191 @@
...
+    kubectl delete ns org2 --ignore-not-found=true
+
+# Check that all network services are running
+check-network:
+    scripts/check-network.sh
+
+
+###############################################################################
+# Test Network
+###############################################################################
```
mbwhite, line 121: I think it would be worth a small line here to check that the org exists... just check that the `organization/<org>` directory exists... first.

mbwhite (changes requested):

PR #926, opened 2023-02-10
Fix an npm dependency error that was breaking the FSAT app compilation

jkneubuh:


mbwhite (approved):

lehors (commented):
I meant to resolve the merge conflict but I realized that the change that this PR meant to make has already been made and is not needed. At the same time this is inserting a fabric-protos dependency that I don't think we want. It appears that this PR should simply be closed at this point. Am I missing something?

PR #1010, opened 2023-03-17
Add in examples and notes on using PurgePrivateData

mbwhite:
- Update the configtx.yaml to enable the 2.5 capabillity
- Added purge into the chaincode
- Added purge into the application client code.


denyeart:
@mbwhite It seems the original update has been lost, could you try to add the updated collection config files again?

mbwhite:
I've pulled down the original PR that made the changes and tried re-applying them (https://github.com/hyperledger/fabric-samples/pull/1012/files)  there is only the file that needs updating..

You mentioned collection config files?  Did you mean configtx.yaml ? 
@denyeart

PR #1011, opened 2023-03-21
Fixed fabric-samples/asset-transfer-basic/application-gateway-typescr…

k3t4ngit:
Fixed fabric-samples/asset-transfer-basic/application-gateway-typescript grpc client type check

### DIFF
--- a/asset-transfer-basic/application-gateway-typescript/src/app.ts
+++ b/asset-transfer-basic/application-gateway-typescript/src/app.ts
@@ -5,7 +5,7 @@
  */
 
 import * as grpc from '@grpc/grpc-js';
-import { connect, Contract, Identity, Signer, signers } from '@hyperledger/fabric-gateway';
+import { connect, Contract, GrpcClient, Identity, Signer, signers } from '@hyperledger/fabric-gateway';
 import * as crypto from 'crypto';
 import { promises as fs } from 'fs';
 import * as path from 'path';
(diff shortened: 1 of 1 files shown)


denyeart:
@bestbeforetoday Can you take a look?

k3t4ngit:
The connect method from @hyperledger/fabric-gateway expects GrpcClient obj as parameter.
But the '@grpc/grpc-js' client is not of type GrpcClient.
I have to cast the type to compile.

bestbeforetoday (changes requested):
This change really should not be necessary, and is certainly not a desirable change. Typings should be correct rather than explicitly force potentially incompatible type conversions using `unknown`.

The fabric-gateway connect() function was changed to take a `GrpcClient` interface changes to the private internals of the grpc-js `Client` class could break type compatibility if the application's versions of grpc-js differed from the one used when packaging fabric-gateway. The interface describes only the public members of the grpc-js Client class to minimise these issues. See hyperledger/fabric-gateway#555.

It does seem that some of the grpc-js classes referred to in the GrpcClient interface can also suffered from version-to-version type compatibility issues between different versions of grpc-js. To minimise this problem, grpc-js was removed as a direct dependency of the sample application, and the application instead relies on exactly the version of grpc-js resolved as a dependency of fabric-gateway.

Running locally from a clean environment and using the latest state of the fabric-samples **main** branch, I don't see any issues. `npm install` and `npm start` both run cleanly for me without this change. Running `npm ls @grpc/grpc-js` gives me this output:

```
asset-transfer-basic@1.0.0
└─┬ @hyperledger/fabric-gateway@1.2.2
  ├── @grpc/grpc-js@1.8.14
  └─┬ @hyperledger/fabric-protos@0.2.0
    └── @grpc/grpc-js@1.8.14 deduped
```

If you do see typing issues, it would be good to understand how they have happened and with what package versions, not to just apply unsafe casts to the code.

denyeart:
Based on bestbeforetoday comments, I'll go ahead and close.

PR #1013, opened 2023-03-22
Cancel existing workflows on subsequent PR push

denyeart:
Cancel existing workflows on subsequent PR push.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
 env:
   GO_VER: 1.19.6
   NODE_VER: 16.x
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -12,7 +12,7 @@ on:
     branches: [ "main", "release-2.5" ]
 
 concurrency:
-  group: ${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 env:
--- a/.github/workflows/test-fsat.yaml
+++ b/.github/workflows/test-fsat.yaml
@@ -9,7 +9,7 @@ on:
       - "release-2.5"
 
 concurrency:
-  group: ${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/test-fsat.yaml
+++ b/.github/workflows/test-fsat.yaml
@@ -8,6 +8,10 @@ on:
       - "main"
       - "release-2.5"
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
 jobs:
   ansible:
     runs-on: ubuntu-20.04
@@ -53,5 +57,3 @@ jobs:
         uses: ./.github/actions/fsat-setup
       - run: just test-console
         working-directory: full-stack-asset-transfer-guide
-
-
--- a/.github/workflows/test-network-events.yaml
+++ b/.github/workflows/test-network-events.yaml
@@ -12,7 +12,7 @@ on:
     branches: [ "main", "release-2.5" ]
 
 concurrency:
-  group: ${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/test-network-events.yaml
+++ b/.github/workflows/test-network-events.yaml
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
 jobs:
   events:
     runs-on: ubuntu-20.04
@@ -34,4 +38,3 @@ jobs:
         env:
           CHAINCODE_NAME: ${{ matrix.chaincode-name }}
           CHAINCODE_LANGUAGE: ${{ matrix.chaincode-language }}
-
--- a/.github/workflows/test-network-gateway.yaml
+++ b/.github/workflows/test-network-gateway.yaml
@@ -12,7 +12,7 @@ on:
     branches: [ "main", "release-2.5" ]
 
 concurrency:
-  group: ${{ github.head_ref }}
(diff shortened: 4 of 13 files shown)


.github/workflows/lint.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
```
denyeart, line 17: @bestbeforetoday 
I came to the same conclusion as you about canceling existing runs when a PR is pushed again, but implemented it with a little different `group` specification relative to yours at:
https://github.com/hyperledger/fabric-gateway/blob/b1659c8d92917b3237aaceb9f6d039c049c6897a/.github/workflows/pull_request.yml#L8-L10

github.head_ref will be the user's branch name, so I think the end result is the same... do you have a preference on the approach?

ryjones:
@bestbeforetoday @denyeart check out https://github.com/hyperledger/fabric-samples/pull/1015

.github/workflows/lint.yaml
bestbeforetoday: It looks like the fabric-samples build kicks of multiple workflows in response to a PR. To avoid one PR workflow cancelling other workflows for the same PR, you might need to include the `github.workflow` as part of the group name, so each workflow's group is unique. This example from the GitHub docs is essentially the same as the approach I took, just always using the ref rather than using the PR number where present:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow

.github/workflows/test-network-events.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-fsat.yaml
```diff
@@ -8,6 +8,10 @@ on:
       - "main"
       - "release-2.5"
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 12:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-network-gateway.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-network-hsm.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-network-k8s.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-network-ledger.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-network-off-chain.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-network-private.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-network-sbe.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/test-network-secured.yaml
```diff
@@ -11,6 +11,10 @@ on:
   pull_request:
     branches: [ "main", "release-2.5" ]
 
+concurrency:
+  group: ${{ github.head_ref }}
```
ryjones, line 15:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

ryjones (approved):

ryjones:
@denyeart @bestbeforetoday updated reference

bestbeforetoday (approved):
The changes look good to me. Any workflow triggered by a pull request she have the these concurrency settings and cancellation.

For a future PR, the workflows might be easier to manage if there was a single workflow triggered by a pull request which then orchestrated all the other required flows as callable workflows. Then there would be a single build for a pull request and only one place to specify the concurrency setting.

ryjones:
> For a future PR, the workflows might be easier to manage if there was a single workflow triggered by a pull request which then orchestrated all the other required flows as callable workflows. Then there would be a single build for a pull request and only one place to specify the concurrency setting.

True; however, you the run into the possibly spurious cancellation of runs when the triggering job runs for over six hours. Even though no component step workflow is hitting the limit, scheduling and the like could get the total run over 6 hours and cancel the whole thing.

PR #1040, opened 2023-05-10
Update FSAT to use Fabric 2.5 by default

satota2:
This patch updates full-stack-asset-transfer to Fabric 2.5 by default.

In particular, since k8s-builder-peer is set as the default value for peer image,
it needs to be updated to use 0.11.0, which is derived from Fabric v2.5 image, as the version default value.


satota2:
Sorry for submitting several empty pushes to force CI to re-run. 
The CI passed successfully. The last CI Job seems to be failing due to a timing issue.

PR #1047, opened 2023-05-31
Fix test-network to work with BFT consensus.

arkadiPiven:
Added a new option for creating channel:
Running ./network.sh createChannel -bft will initiate a channel running BFT orderers.

### DIFF
--- a/test-network/README.md
+++ b/test-network/README.md
@@ -2,6 +2,26 @@
 
 You can use the `./network.sh` script to stand up a simple Fabric test network. The test network has two peer organizations with one peer each and a single node raft ordering service. You can also use the `./network.sh` script to create channels and deploy chaincode. For more information, see [Using the Fabric test network](https://hyperledger-fabric.readthedocs.io/en/latest/test_network.html). The test network is being introduced in Fabric v2.0 as the long term replacement for the `first-network` sample.
 
+If you are planning to run the test network with consesnsus type BFT then please pass `-bft` flag as input to the `network.sh` script when creating the channel.
+That is to create a network use:
+```bash
+./network.sh up -bft
+```
+
+To create a channel use:
+
+```bash
+./network.sh createChannel -bft
+```
+
+To restart a running network use:
+
+```bash
+./network.sh restart -bft
+```
+
+Note that running the createChannel command will start the network, if it is not already running.
+
 Before you can deploy the test network, you need to follow the instructions to [Install the Samples, Binaries and Docker Images](https://hyperledger-fabric.readthedocs.io/en/latest/install.html) in the Hyperledger Fabric documentation.
 
 ## Using the Peer commands
--- a/test-network/bft-config/configtx.yaml
+++ b/test-network/bft-config/configtx.yaml
@@ -38,6 +38,9 @@ Organizations:
         Rule: "OR('OrdererMSP.admin')"
     OrdererEndpoints:
       - orderer.example.com:7050
+      - orderer.example.com:7052
+      - orderer.example.com:7056
+      - orderer.example.com:7058
   - &Org1
     # DefaultOrg defines the organization which is used in the sampleconfig
     # of the fabric.git development environment
@@ -108,13 +111,13 @@ Capabilities:
   # supported by both.
   # Set the value of the capability to true to require it.
   Channel: &ChannelCapabilities
-    # V2_0 capability ensures that orderers and peers behave according
-    # to v2.0 channel capabilities. Orderers and peers from
-    # prior releases would behave in an incompatible way, and are therefore
-    # not able to participate in channels at v2.0 capability.
-    # Prior to enabling V2.0 channel capabilities, ensure that all
-    # orderers and peers on a channel are at v2.0.0 or later.
-    V2_0: true
+    # V3.0 for Channel is a catchall flag for behavior which has been
+    # determined to be desired for all orderers and peers running at the v3.0.0
+    # level, but which would be incompatible with orderers and peers from
+    # prior releases.
+    # Prior to enabling V3.0 channel capabilities, ensure that all
+    # orderers and peers on a channel are at v3.0.0 or later.
+    V3_0: true
   # Orderer capabilities apply only to the orderers, and may be safely
   # used with prior release peers.
   # Set the value of the capability to true to require it.
@@ -177,13 +180,6 @@ Application: &ApplicationDefaults
 #
 ################################################################################
 Orderer: &OrdererDefaults
-  # Addresses used to be the list of orderer addresses that clients and peers
-  # could connect to.  However, this does not allow clients to associate orderer
-  # addresses and orderer organizations which can be useful for things such
-  # as TLS validation.  The preferred way to specify orderer addresses is now
-  # to include the OrdererEndpoints item in your org definition
-  Addresses:
-    - orderer.example.com:7050
   # Batch Timeout: The amount of time to wait before creating a batch
   BatchTimeout: 2s
   # Batch Size: Controls the number of messages batched into a block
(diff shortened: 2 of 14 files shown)


arkadiPiven:
@yacovm

test-network/bft-config/configtx.yaml
```diff
@@ -0,0 +1,317 @@
...
+        RequestAutoRemoveTimeout: 3m0s
+        ViewChangeResendInterval: 5s
+        ViewChangeTimeout: 20s
+        LeaderHeartbeatTimeout: 1m0s
+        CollectTimeout: 1s
+        RequestBatchMaxBytes: 10485760
+        IncomingMessageBufferSize: 200
+        RequestPoolSize: 100000
+        LeaderHeartbeatCount: 10
+      ConsenterMapping:
```
yacovm, line 283: @denyeart since we don't have any sample except the nano bash with BFT, I asked Arkadi to extend this with 4 orderers so we will have another sample.

test-network/scripts/configUpdate.sh
```diff
@@ -25,6 +25,7 @@ fetchChannelConfig() {
 
   infoln "Decoding config block to JSON and isolating config to ${OUTPUT}"
   set -x
+  cksum `which configtxlator`
```
yacovm, line 28: you don't need this

test-network/scripts/createChannel.sh
```diff
@@ -3,15 +3,19 @@
 # imports  
 . scripts/envVar.sh
 . scripts/utils.sh
+export PATH="${PWD}"/../../fabric/build/bin:"${ROOTDIR}"/../bin:"${PWD}"/../bin:"$PATH"
```
yacovm, line 6: are you sure this is needed?

yacovm:
Have we tested that adding a new node works? 

How do we set the policy when adding a new node?

test-network/scripts/createChannel.sh
arkadiPiven: You need this in case someone will want to use old binaries.
Me and Dave had the same approach when constructing nano-bash

yacovm:
@denyeart let's get this PR in, shall we?

test-network/bft-config/configtx.yaml
```diff
@@ -0,0 +1,317 @@
...
+  # Channel capabilities apply to both the orderers and the peers and must be
+  # supported by both.
+  # Set the value of the capability to true to require it.
+  Channel: &ChannelCapabilities
+    # V2_0 capability ensures that orderers and peers behave according
+    # to v2.0 channel capabilities. Orderers and peers from
+    # prior releases would behave in an incompatible way, and are therefore
+    # not able to participate in channels at v2.0 capability.
+    # Prior to enabling V2.0 channel capabilities, ensure that all
+    # orderers and peers on a channel are at v2.0.0 or later.
```
denyeart, line 117: Although this is just comments, would be good to update to v3.0.0.
You can copy/paste the comment from https://github.com/hyperledger/fabric/blob/main/sampleconfig/configtx.yaml#L113-L118

test-network/bft-config/configtx.yaml
```diff
@@ -0,0 +1,317 @@
...
+#
+################################################################################
+Orderer: &OrdererDefaults
+  # Addresses used to be the list of orderer addresses that clients and peers
+  # could connect to.  However, this does not allow clients to associate orderer
+  # addresses and orderer organizations which can be useful for things such
+  # as TLS validation.  The preferred way to specify orderer addresses is now
+  # to include the OrdererEndpoints item in your org definition
+  Addresses:
+    - orderer.example.com:7050
```
denyeart, line 187: Addresses section has been replaced by OrdererEndpoints section above (in the Org definition).
Let's remove this section for v3.0 samples.

test-network/bft-config/configtx.yaml
```diff
@@ -0,0 +1,317 @@
...
+        Rule: "OR('OrdererMSP.member')"
+      Writers:
+        Type: Signature
+        Rule: "OR('OrdererMSP.member')"
+      Admins:
+        Type: Signature
+        Rule: "OR('OrdererMSP.admin')"
+    OrdererEndpoints:
+      - orderer.example.com:7050
+      - orderer.example.com:7052
```
denyeart, line 41: Why only list 2 of the 4 orderer nodes?

test-network/configtx/configtx.yaml
```diff
@@ -13,91 +13,78 @@
...
+        Rule: "OR('OrdererMSP.member')"
+      Writers:
+        Type: Signature
+        Rule: "OR('OrdererMSP.member')"
+      Admins:
+        Type: Signature
+        Rule: "OR('OrdererMSP.admin')"
+    OrdererEndpoints:
+      - orderer.example.com:7050
+      - orderer.example.com:7052
```
denyeart, line 41: The regular test network (non-BFT) only has one orderer node.

test-network/scripts/createChannel.sh
denyeart: Most people that download the samples and try it out will not have a fabric directory. I'd suggest to remove this change.
We added it for the nano bash sample since a target user of that sample was a Fabric developer who compiles their own binaries.

test-network/network.sh
```diff
@@ -18,7 +18,7 @@
 # this script is actually in and infer location from there. (putting first)
 
 ROOTDIR=$(cd "$(dirname "$0")" && pwd)
-export PATH=${ROOTDIR}/../bin:${PWD}/../bin:$PATH
+export PATH="${PWD}"/../../fabric/build/bin:"${ROOTDIR}"/../bin:"${PWD}"/../bin:"$PATH"
```
denyeart, line 21: Most people that download the samples and try it out will not have a fabric directory. I'd suggest to remove this change.
We added it for the nano bash sample since a target user of that sample was a Fabric developer who compiles their own binaries.

test-network/scripts/orderer.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+
+channel_name=$1
+
+export PATH="${PWD}"/../../fabric/build/bin:"${ROOTDIR}"/../bin:"${PWD}"/../bin:"$PATH"
```
denyeart, line 6: Same comment about /fabric path as above, in each of the orderer sh scripts.

test-network/README.md
```diff
@@ -2,6 +2,13 @@
 
 You can use the `./network.sh` script to stand up a simple Fabric test network. The test network has two peer organizations with one peer each and a single node raft ordering service. You can also use the `./network.sh` script to create channels and deploy chaincode. For more information, see [Using the Fabric test network](https://hyperledger-fabric.readthedocs.io/en/latest/test_network.html). The test network is being introduced in Fabric v2.0 as the long term replacement for the `first-network` sample.
 
+If you are planning to run the test network with consesnsus type BFT then please pass `-bft` flag as input to the `network.sh` script when creating the channel.
+That is to create a channel use:
+
+```bash
+./network.sh createChannel -bft
+```
+
```
denyeart, line 11: Need to say the same thing about
`./network.sh up -bft`

So that the four orderer nodes start up.

test-network/network.sh
```diff
@@ -316,7 +319,8 @@ function deployCCAAS() {
 
 # Tear down running network
 function networkDown() {
-
+  local temp_compose=$COMPOSE_FILE_BASE
+  COMPOSE_FILE_BASE=compose-bft-test-net.yaml
```
denyeart, line 323: I'm not following this change... won't this always force the use of compose-bft-test-net.yaml instead of the default compose-test-net.yaml?
Why is the temp_compose required here?

arkadiPiven: You are correct.
The compose-bft-test-net.yaml is a super set of the compose-test-net.yaml. This means that if a user launched a regular network or a bft network, during shutdown all containers will be stopped no matter what initial configuration the user chose.
This actually spares me to pass any flags to the `./network.sh down` command, and to refer to different kinds of networks.

test-network/ADD_ORDERER_TUTORIAL.md
```diff
@@ -0,0 +1,310 @@
+# Adding Orderer To An Existing Network
```
yacovm, line 1: This should be part of the [official documentation](https://github.com/hyperledger/fabric/tree/main/docs/source), not in the samples.

test-network/scripts/add_new_orderer_to_config.py
```diff
@@ -0,0 +1,98 @@
...
+    new_identity['principal']['id_bytes'] = identity
+    identities.append(new_identity)
+    _log_update('block validation identities', identities_before_update, identities)
+
+    rule = config['channel_group']['groups']['Orderer']['policies']['BlockValidation']['policy']['value'][
+        'rule']
+    rule_before_update = copy.deepcopy(rule)
+    rule['n_out_of']['n'] = _calculate_bft_quorum(new_orderers_count)
+    rule['n_out_of']['rules'].append({'signed_by': new_orderers_count - 1})
+    _log_update('block validation rules', rule_before_update, rule)
```
denyeart, line 75: One general question and one specific question.

General question - I didn't see this file originally... is it used in this PR? Or will other instructions use it?

Specific question - I didn't see anything like these BlockValidation policies in the original configtx.yaml, like I did for the Orderer Endpoints (above) and consenter mapping (below). Can you educate me on this... why are they not in configtx.yaml? Is it in the channel config implicitly somehow originally? Is this new for BFT?

yacovm: > I didn't see anything like these BlockValidation policies in the original configtx.yaml, like I did for the Orderer Endpoints (above) and consenter mapping (below). Can you educate me on this... why are they not in configtx.yaml? Is it in the channel config implicitly somehow originally? Is this new for BFT?


I'll answer this one, and will let Arkadi answer the first one.

First of all, remember that the "name" of the policy in configtx.yaml is a mere pointer to a meta policy of "any orderer". 
The meta policies we have, or, the policies we have in general in Fabric, are not suitable to define BFT policies, as BFT policies depend on additional input found in the channel config. To add support for that, it requires invasive changes to the channel config infrastructure. 

However, a signature policy is a supported policy that can always be used and it is flexible enough to fit BFT block validation.

In the spirit of making Fabric more generic and pluggable, we don't want to define policy names for every type of consensus we can imagine. Some protocols require more signatures, some require less, some require a single threshold signature, etc. 
While in Raft it's easy to define a policy string ("any orderer"), for BFT it's complex to do that in the configtx.yaml, and it can also be error prone. 

For BFT, the Fabric tooling (configtxgen) overrides the block validation policy that is defined in the configtx.yaml, both in order to "protect the user from a misconfiguration", and also because it was my belief that a user should not need to understand the underlying BFT protocol to configure the policy. 
So, when a channel is created using configtxgen, the genesis block contains the right policy. When a channel is updated, the orderer itself checks that the policy was correctly computed via an [ad-hoc verification](https://github.com/hyperledger/fabric/blob/main/orderer/consensus/smartbft/configverifier.go#L93-L168).

arkadiPiven: For the general question:
This file is used in the following tutorial:
https://github.com/hyperledger/fabric/pull/4390

denyeart (approved):
The sample looks great now, let's go ahead and merge while we finish out the doc PR.

PR #1049, opened 2023-06-01
Update token_contract.go, modify the overflow judgment condition for function add().

yjwxfq:
The general overflow  judgment condition for function add is not in every case.
for example, b = 5, q = -3, sum = 2, 
if (sum < q) == (b >= 0 && q >= 0) wil be true.
I modify this line for making it satisfied with every case.


denyeart (approved):
Good catch, thanks!

PR #1053, opened 2023-06-08
Update token_contract.go, modify the overflow judgment condition for function sub().

yjwxfq:
The general overflow judgment condition for function sub is not in every case.
The judgment result is wrong in the condition of b < 0 && q >=0
for example, b = -3, q = 2, diff = -5, it's not overflow,
but if (diff > b) == (b >= 0 && q >= 0) wil be true.

for another example, b = -3, q = maxint, b-q is overflow,
but if (diff > b) == (b >= 0 && q >= 0) wil be false.

I modify this line for making it use the same judgment logic as the function add().


token-erc-20/chaincode-go/chaincode/token_contract.go
```diff
@@ -723,7 +723,7 @@ func sub(b int, q int) (int, error) {
 	var diff int
 	diff = b - q
 
-	if (diff > b) == (b >= 0 && q >= 0) {
+	if (diff < b || diff < -q) == (b >= 0 && q <= 0) {
```
denyeart, line 726: The equation is getting confusing... could we simplify it by validating that both b and q are positive integers at the beginning of the function?  Same thing for add().

yjwxfq: That would weaken the generality of the function. And what is the significance of the overflow judgment in the previous source code.

denyeart: I am indeed proposing to weaken the generality of the function. Token balances and transfers should always utilize positive integers, right? So why is there a need to support negative integers in this chaincode? Having the function explicitly only support positive integers would make it more expressive and scoped to the intent wouldn't it?

denyeart (approved):
Thanks!

PR #1060, opened 2023-06-21
Fix transferFrom function so that we only delete token balances after…

robevansuk:
… we've verified the funds are available first - not before

### DIFF
--- a/token-erc-1155/chaincode-go/chaincode/contract.go
+++ b/token-erc-1155/chaincode-go/chaincode/contract.go
@@ -9,6 +9,7 @@ package chaincode
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/hyperledger/fabric-protos-go/ledger/queryresult"
 	"sort"
 	"strconv"
 	"strings"
@@ -949,6 +950,8 @@ func removeBalance(ctx contractapi.TransactionContextInterface, sender string, i
 		}
 		defer balanceIterator.Close()
 
+		var deferredDeletions = []func() error{}
+
 		// Iterate over keys that store balances and add them to partialBalance until
 		// either the necessary amount is reached or the keys ended
 		for balanceIterator.HasNext() && partialBalance < neededAmount {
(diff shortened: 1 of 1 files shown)


robevansuk:
Just tagging previous reviewers in the hopes someone can take a look @denyeart @yacovm

yacovm:
> Just tagging previous reviewers in the hopes someone can take a look @denyeart @yacovm

I don't have merge rights for this repository though.look at: https://github.com/hyperledger/fabric-samples/blob/main/MAINTAINERS.md

robevansuk:
Tagging maintainers @jkneubuh @mbwhite @nikhil550 @satota2 for a review. This Pr defers deleting token balances until after the necessary funds have been allocated. Without it, token balances are deleted before the necessary funds are identified. When testing in isolation this lead to unexpected results where the token balances were not as expected for failed transactions.

satota2 (commented):
@robevansuk Thank you for your contribution.
Maybe it's my lack of understanding, but since the deletions occurred within a single transaction, token loss did not seem to occur upon failure.

Would it be possible for you to provide a specific example of a case where this would cause a problem?

robevansuk:
I think you may be right - if the transaction fails, the token loss wouldn't occur. However testing the function in isolation does highlight that the deletion occurs before the required balance has been identified, ready for a transfer to occur. This means testing in isolation doesn't work as expected since the data is deleted, then an error occurs (when enough tokens are not identified), and the state becomes inconsistent with what is expected. So ultimately this change improves the code enough to make it worth while. Deferring the deletion until after the check for partialbalance > required funds works better from a logical point of view. Deleting the token balances, then erroring out is not testable, as it was previously.

token-erc-1155/chaincode-go/chaincode/contract.go
```diff
@@ -972,47 +975,64 @@ func removeBalance(ctx contractapi.TransactionContextInterface, sender string, i
 				selfRecipientKeyNeedsToBeRemoved = true
 				selfRecipientKey = queryResponse.Key
 			} else {
-				err = ctx.GetStub().DelState(queryResponse.Key)
-				if err != nil {
-					return fmt.Errorf("failed to delete the state of %v: %v", queryResponse.Key, err)
-				}
+				deferredDeletions = append(deferredDeletions, deferredDelete(ctx, queryResponse))
```
denyeart, line 978: DelState() won't delete at this time. It will build a write set that will perform the deletes upon successful completion of the chaincode and validation of the transaction. It is therefore implicitly deferred already. Introducing an explicit defer in my opinion makes the chaincode more confusing to the user, and gives the wrong idea that there is some type of explicit deferral possible in the chaincode programming model.

I see your point however about making the implicit deferral more obvious to the user. I'd suggest that the better way to educate the user would be to add a comment at the existing DelState() call indicating that the DelState() calls will get collected in the write set and will only be executed if the remaining checks are passed and the transaction gets validated. This will educate the user on the chaincode programming model and keep the code more readable.

robevansuk: This makes sense but when testing with shimtest.MockStub, the change is made at the time its executed and is not part of any writeSet that will ignore the deletion.

A better solution might be to push the deletion further down in the chain of events but I'd like confirmation whether this would be worthwhile/acceptable before I make the change.

PR #1081, opened 2023-08-15
chore: add missing deps of application-gateway-typescript

Nasfame:
- there were a couple of **deps** missing like `@grpc/grpc-js`
- added 'em

### DIFF
--- a/asset-transfer-basic/application-gateway-typescript/package.json
+++ b/asset-transfer-basic/application-gateway-typescript/package.json
@@ -19,6 +19,7 @@
     "author": "Hyperledger",
     "license": "Apache-2.0",
     "dependencies": {
+        "@grpc/grpc-js": "^1.9.0",
         "@hyperledger/fabric-gateway": "^1.1.1"
     },
     "devDependencies": {
--- a/asset-transfer-basic/application-gateway-typescript/pnpm-lock.yaml
+++ b/asset-transfer-basic/application-gateway-typescript/pnpm-lock.yaml
@@ -1,0 +1,1167 @@
+lockfileVersion: '6.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+dependencies:
+  '@grpc/grpc-js':
+    specifier: ^1.9.0
+    version: 1.9.0
+  '@hyperledger/fabric-gateway':
+    specifier: ^1.1.1
+    version: 1.1.1
+
+devDependencies:
+  '@tsconfig/node14':
+    specifier: ^1.0.3
+    version: 1.0.3
+  '@types/node':
+    specifier: ^14.18.16
+    version: 14.18.16
+  '@typescript-eslint/eslint-plugin':
+    specifier: ^5.22.0
+    version: 5.22.0(@typescript-eslint/parser@5.22.0)(eslint@8.14.0)(typescript@4.6.4)
+  '@typescript-eslint/parser':
+    specifier: ^5.22.0
+    version: 5.22.0(eslint@8.14.0)(typescript@4.6.4)
+  eslint:
+    specifier: ^8.14.0
+    version: 8.14.0
+  typescript:
+    specifier: ~4.6.4
+    version: 4.6.4
+
+packages:
+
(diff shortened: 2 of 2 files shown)


asset-transfer-basic/application-gateway-typescript/pnpm-lock.yaml
denyeart, line 1: We typically do not check in the lock files.

asset-transfer-basic/application-gateway-typescript/package.json
```diff
@@ -19,6 +19,7 @@
     "author": "Hyperledger",
     "license": "Apache-2.0",
     "dependencies": {
+        "@grpc/grpc-js": "^1.9.0",
```
denyeart, line 22: @bestbeforetoday Is this necessary?

asset-transfer-basic/application-gateway-typescript/pnpm-lock.yaml
Nasfame: think its best you do. Transitive dependency versions are preserved. NodeJS community highly recommends it

asset-transfer-basic/application-gateway-typescript/package.json
Nasfame: FYI i had to manually install it.

bestbeforetoday: Using `npm`, as the instructions for the samples do, this is not necessary. `@grpc/grpc-js` is a dependency of `@hyperledger/fabric-gateway` so gets installed and is visible to the sample application. `pnpm` looks to have been used instead of `npm`, and that behaves differently. `pnpm` only allows access to modules explicitly specified as dependencies in the application’s `package.json` file.

The reason the `@grpc/grpc-js`dependency was removed from the sample application’s `package.json` is that updates to `@hyperledger/fabric-gateway` meant that different dependency versions of `@grpc/grpc-js` could exist. This caused TypeScript errors when private properties of public gRPC classes changed. To minimise the chance of similar errors occurring in the future with the samples, the explicit dependency was removed.

Nasfame: ok. Why is there no package-lock.json?

bestbeforetoday: What do you need the `package-lock.json` for?

Nasfame: Ofcourse for dep version control

bestbeforetoday: I don’t really want to tie the sample applications to a specific patch release of `@hyperledger/fabric-gateway` and other Fabric packages. My preference would be for them to use the latest minor/patch release. Having a `package-lock.json` would require that to be updates for all samples on every release, which is something I would prefer not to have to do.

If the consensus is that having a `package-lock.json` tie dependencies down to specific versions that were used the last time the code was modified, I could live with that. It might mean issues like the gRPC library version incompatibility described above don’t surface so easily. You could argue this is good for users of the samples. You could argue this is worse for users of the Fabric APIs in general as issues are not identified and resolved so quickly.

In the past I have found occasionally instances where having a `package-lock.json` seems top have caused issues - perhaps with dependencies containing native code - when the runtime environment uses a different Node version or platform. I don’t know if this is something likely to cause issues here but it makes me wary of shipping `package-lock.json` files as part of samples.

Nasfame: I thought these samples were for beginners and i wanted it to work as expected without any issues

bestbeforetoday: As I understand it, they do work as expected. The problem was that you used `pnpm` instead of `npm` as described in the instructions.

We could enhance the samples to ensure that they work with `pnpm` in addition to `npm`. I think this should include testing **all** of the samples using `pnpm` in the build, not just changing the dependencies for one sample.

We could also include `package-lock.json` files for **all** of the samples, not just one sample.

Those two changes are not related so _might_ be better tackled in separate pull requests.

Nasfame: ok.

PR #1083, opened 2023-08-25
Improve BFT sample on test-network

satota2:
This patch includes:
- Fixed a minor bug in the parsing of the BFT flag
- Update of printHelp for use of the BFT flag
- Fixed explanations

### DIFF
--- a/test-network/scripts/createChannel.sh
+++ b/test-network/scripts/createChannel.sh
@@ -94,6 +94,10 @@ setAnchorPeer() {
 }
 
 
+## User attempts to use BFT orderer in Fabric network with CA
+if [ $BFT -eq 1 ] && [ -d "organizations/fabric-ca/ordererOrg/msp" ]; then
+  fatalln "Fabric network seems to be using CA. This sample does not yet support the use of consensus type BFT and CA together."
+fi
 
 ## Create channel genesis block
 FABRIC_CFG_PATH=$PWD/../config/
--- a/test-network/scripts/utils.sh
+++ b/test-network/scripts/utils.sh
@@ -14,7 +14,8 @@ function printHelp() {
     println "  network.sh \033[0;32mup\033[0m [Flags]"
     println
     println "    Flags:"
-    println "    -ca <use CAs> -  Use Certificate Authorities to generate network crypto material"
+    println "    -ca - Use Certificate Authorities to generate network crypto material"
+    println "    -bft - Use Orderers with consensus type BFT (Not available in Fabric v2.x)"
     println "    -c <channel name> - Name of channel to create (defaults to \"mychannel\")"
     println "    -s <dbtype> - Peer state database to deploy: goleveldb (default) or couchdb"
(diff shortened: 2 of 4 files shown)


test-network/scripts/utils.sh
```diff
@@ -24,16 +25,18 @@ function printHelp() {
     println "    -h - Print this message"
     println
     println " Possible Mode and flag combinations"
-    println "   \033[0;32mup\033[0m -ca -r -d -s -verbose"
-    println "   \033[0;32mup createChannel\033[0m -ca -c -r -d -s -verbose"
+    println "   \033[0;32mup\033[0m -ca -bft -r -d -s -verbose"
+    println "   \033[0;32mup createChannel\033[0m -ca -bft -c -r -d -s -verbose"
```
lehors, line 29: This is in contradiction with the note added "that currently this sample does not yet support the use of consensus type BFT and CA together", isn't it?

test-network/scripts/utils.sh
```diff
@@ -125,9 +130,9 @@ function printHelp() {
     println "    -h - Print this message"
     println
     println " Possible Mode and flag combinations"
-    println "   \033[0;32mup\033[0m -ca -r -d -s -verbose"
-    println "   \033[0;32mup createChannel\033[0m -ca -c -r -d -s -verbose"
-    println "   \033[0;32mcreateChannel\033[0m -c -r -d -verbose"
+    println "   \033[0;32mup\033[0m -ca -bft -r -d -s -verbose"
+    println "   \033[0;32mup createChannel\033[0m -ca -bft -c -r -d -s -verbose"
+    println "   \033[0;32mcreateChannel\033[0m -bft -c -r -d -verbose"
```
lehors, line 135: same here.

lehors (commented):
There is a contradiction in the information being added. Ideally we would add code to prevent users from specifying both -ca and -bft if that combination isn't currently supported although this could be done in a different PR.

satota2:
@lehors Thank you for your reviews.

I fixed the PRs based on your comments and submited it.
Please take the time to review it.

> add code to prevent users from specifying both -ca and -bft if that combination isn't currently supported

Since there are cases where up command is executed separately, I have added check logic to prevent createChannel using BFT orderer on CA-enabled Fabric network.

test-network/scripts/createChannel.sh
```diff
@@ -94,6 +94,10 @@ setAnchorPeer() {
 }
 
 
+## User attempts to use BFT orderer in Fabric network with CA
+if [ $BFT -eq 1 ] && [ -d "organizations/fabric-ca/ordererOrg/msp" ]; then
+  fatalln "Fabric network seems to be using CA. This sample does not yet support the use of consensus type BFT and CA together."
+fi
```
lehors, line 100: This test should be done after parsing the command line arguments so that it is also triggered if one simply does: "./network.sh up -ca -bft"

lehors (approved):
That works. Thank you!

PR #1086, opened 2023-09-07
Update TypeScript HSM sample application

bestbeforetoday:
- Update to Node 16.
- Simplify SKI code using Node standard packages and avoiding use of jsrsasign.


bestbeforetoday:
@hyperledger/fabric-samples-maintainers I think this PR is ready for review. My intention was to deliver a small tidy up of the HSM sample but, when doing so, I discovered that large numbers of samples were failing in the build. I have fixed the majority of them in the second commit associated with this PR. The remaining failures are in Kubernetes deployment of the Full-stack Asset Transfer Guide. I do not know what changes are required to fix this failure.

denyeart (approved):
Thanks @bestbeforetoday !

jkneubuhl:
Hi @bestbeforetoday and @denyeart 

Sorry about the test failures falling off the vine.  It looks like there may have been a little bit of drift for the version of `kind` that is used by the test suite / sample app.  The binaries are pinned to a slightly out of date revision of kind / kubernetes (1.24.x) and it may be having troubles coming up on the latest github test runners.  The kind / kube clients are up to 1.28.x.

PR #1091, opened 2023-09-28
Move publishing of REST sample Image from AZP to GHA

satota2:
This patch moves publishing of REST sample image from AZP to GHA.
This was the last job running on AZP.

Also, this patch will fix an error encounterd during the image build.
Publishing of REST sample image is now only executed when pushing to main.


satota2:
Hi @denyeart
https://lists.hyperledger.org/g/fabric/message/11994

I have attempted to create a patch for migrating the remaining CI jobs of fabric-samples to GHA, as mentioned in the following message:
https://lists.hyperledger.org/g/fabric/message/11994

I would appreciate it if you could review it.
Additionally, I have tried to change the method for "Publishing of REST sample image" to use GITHUB_TOKEN. However, I haven't been able to test it locally or in my own environment, so please verify it.

denyeart (approved):
Thank you... I'll merge and we'll see if it works from GitHub Actions...

PR #1095, opened 2023-10-06
Add chaincode utilities

celder628:
This change is meant to help cover end user functionality that was previously covered by the blockchain VSCode Plugin.

Functions added:
- cc mode with package, list chaincode, invoke and query functions
- auto sequencing for chaincode deployment
- move variables into config file

### DIFF
--- a/test-network/network.config
+++ b/test-network/network.config
@@ -1,0 +1,58 @@
+# default image tag, example: "2.5.4".  "default" will download the latest. (-i)
+IMAGETAG="default"
+
+# default ca image tag, example: "1.5.7". "default" will download the latest. (-cai)
+CA_IMAGETAG="default"
+
+# Using crpto vs CA. default is cryptogen
+CRYPTO="cryptogen"
+
+# max number of retries before giving up (-r)
+MAX_RETRY=5
+
+# default for delay between commands (-d)
+CLI_DELAY=3
+
+# channel name defaults to "mychannel" (-c)
+CHANNEL_NAME="mychannel"
+
+# default database (-s)
+DATABASE="leveldb"
+
+# default org (-org)
+ORG=1
+
+# chaincode language defaults to "NA" (-ccl)
+CC_SRC_LANGUAGE="go"
+
+# Chaincode version (-ccv)
+CC_VERSION="1.0.1"
+
+# chaincode name defaults to "NA" (-ccn)
+CC_NAME="basic"
+
+# default to running the docker commands for the CCAAS (-ccaasdocker)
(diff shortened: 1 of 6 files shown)


test-network/scripts/utils.sh
```diff
@@ -146,9 +198,26 @@ function printHelp() {
...
 }
 
+function installPrereqs() {
+
+  infoln "installing prereqs"
+
+  FILE=../install-fabric.sh     
+  if [ ! -f $FILE ]; then
+    curl -sSLO https://raw.githubusercontent.com/hyperledger/fabric/main/scripts/install-fabric.sh && chmod +x install-fabric.sh
+    cp install-fabric.sh
```
denyeart, line 213: I think you intended to copy it to the parent directory fabric-samples?
```suggestion
    cp install-fabric.sh ..
```

test-network/network.config
```diff
@@ -0,0 +1,60 @@
...
+#
+# -ccl
+# chaincode language defaults to "NA"
+CC_SRC_LANGUAGE="go"
+# -ccv
+# Chaincode version
+CC_VERSION="1.0.1"
+# -ccn
+# chaincode name defaults to "NA"
+CC_NAME="asset-transfer-basic-go"
```
denyeart, line 34: The doc tutorials use a default name of "basic" so that any language can be used without having to change the instructions for post-deployment steps. You may want to use default name of "basic" for consistency, so that people can toggle between the existing tutorials and this new way.

test-network/network.config
```diff
@@ -0,0 +1,60 @@
+# -i  
+# default image tag
+IMAGETAG="2.5.4"
+# -cai
+# default ca image tag
+CA_IMAGETAG="1.5.7"
+
+# Using crpto vs CA. default is cryptogen
+CRYPTO="cryptogen"
```
denyeart, line 9: It would be easier to read if there was a blank line between each property,
especially since some have the flag name and some don't (e.g. `-r` below, does it go with the above property or the next property?).

denyeart (commented):
Nice work! I had to review very carefully to find a few suggestions :-)

The auto sequencing is very nice, this really helps with iterative development. I tried it and it works seamlessly!

test-network/network.config
celder628: I moved the property to the same line as the description and added a blank line between properties.

test-network/network.config
celder628: Corrected to "basic"

denyeart (approved):

PR #1132, opened 2023-11-05
Fix test-network chaincode-as-a-service deployment

bestbeforetoday:
Resolves a regression introduced in e73bb717db47185a8ccd1b701503b369cd2a73e1.

Previous change defaulted to "auto" as the sequence number but omitted the call to generate the correct sequence number for the "auto" value. This resulted in "auto" being used as the sequence number, which is an error since an integer is required.

### DIFF
--- a/test-network/scripts/deployCCAAS.sh
+++ b/test-network/scripts/deployCCAAS.sh
@@ -177,6 +177,8 @@ installChaincode 1
 infoln "Install chaincode on peer0.org2..."
 installChaincode 2
 
+resolveSequence
+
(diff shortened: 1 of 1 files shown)


bestbeforetoday:
@denyeart @celder628 This should allow builds that make use of fabric-samples and the test-network do test chaincode-as-a-service to work again, which includes [fabric-test](https://github.com/hyperledger/fabric-test) and [fabric-contract-api-go](https://github.com/hyperledger/fabric-contract-api-go).

denyeart (approved):
Thanks Mark!

PR #1145, opened 2023-12-07
Use latest v2.5 fabric-chaincode-java

bestbeforetoday:
- Update Gradle wrappers
- Update shadow plugin


denyeart (approved):
Wow, big PR, nice job, thanks!

PR #1183, opened 2024-02-29
Read first client cert from cert directory

bestbeforetoday:
Allow client applications to work regardless of whether network was creating using a CA or cryptogen since they create different client certificate file names.


denyeart (approved):
Nice, thank you!

PR #1184, opened 2024-03-16
Update go.mod

Ekhwatenge:


denyeart:
In go.mod, the major.minor version is used.

PR #1185, opened 2024-03-19
Fix to configtx.yaml to enable it to work as intended with the test network and channel tutorials

schrodingerscode:
It seems the profile name was incorrectly changed to from "TwoOrgsApplicationGenesis" as the fabric tutorial requires to "ChannelUsingRafts".  This latter profile name creates an error when setting up channels using the test network and tutorials.  This update changes the profile name back to "TwoOrgsApplicationGenesis" as is required for the test network tutorials for which this file was created.

### DIFF
--- a/test-network/configtx/configtx.yaml
+++ b/test-network/configtx/configtx.yaml
@@ -257,7 +257,7 @@ Channel: &ChannelDefaults
 #
 ################################################################################
 Profiles:
-  ChannelUsingRaft:
+  TwoOrgsApplicationGenesis:
     <<: *ChannelDefaults
     Orderer:
       <<: *OrdererDefaults


denyeart:
We were going to evolve to "ChannelUsingRaft", but the doc pull request https://github.com/hyperledger/fabric/pull/4528 never got merged because the commit wasn't signed. I'm fine going back to "TwoOrgsApplicationGenesis", but please grep fabric-samples for "ChannelUsingRaft" and update all occurrences so that we are consistent.

schrodingerscode:
Ok. I have made the same update to all 5 (4 additional) instances that grep returned.  Specifically, each instance of "ChannelUsingRafts" was changed back to "TwoOrgsApplicationGenesis".

denyeart:
@schrodingerscode @satota2 The doc changes to shift to "ChannelUsingRaft" have finally been merged in  https://github.com/hyperledger/fabric/pull/4781 and https://github.com/hyperledger/fabric/pull/4788, so let's keep fabric-samples using "ChannelUsingRaft". I'll close this PR.

schrodingerscode:
Sounds good.  Thanks.

PR #1191, opened 2024-04-04
Update Go gRPC dependency

bestbeforetoday:
grpc.Dial() is (soon to be) deprecated in current gRPC versions. Use grpc.NewClient() instead.


satota2 (commented):
Hello @bestbeforetoday ,
Thank you for submitting this PR. I have a question about the `NewClient()` function. It seems like it would work just specifying `localhost:port` for the endpoint. Is there a specific need to include `dns:///`? Since 'dns' is used by default, I thought it might not be necessary to specify it explicitly. I'm not very familiar with the newer implementations of `NewClient()` in gRPC, so I would appreciate if you could share any practices or reasoning behind this choice.

Other that, LGTM!

bestbeforetoday:
If no URI scheme is specified, dns is used by default. The [documentation](https://pkg.go.dev/google.golang.org/grpc#NewClient) does specify that the argument should be a URI and gives an example using the `dns:///` prefix. We could omit the schema but I don’t think there is any harm in specifying the complete URI as intended.

satota2 (approved):
@bestbeforetoday 
Thank you for your response. I understand now that specifying the schema explicitly is recommended in the documentation, so your code is appropriate.

PR #1195, opened 2024-04-12
test-network envVar.sh script improvement

denyeart:
Change shell script to use single equals.

In my shell environment single bracket with double equals did not work. It caused failure when running the chaincode-external tutorial README.

It looks like for maximum portability, it is best to use single bracket with single equals.


denyeart:
@mergifyio rebase

denyeart:
@satota2 I found this while testing your chaincode-external updates...

satota2 (approved):
@denyeart Thanks!

PR #1199, opened 2024-04-29
Update nano test network

### LINKED ISSUE #1198: Nano test network problem
I think commit 81390fbbbfb3e1ba2757cb7cce6b723ff65acfb8 has broken the nano test network- the `./` was removed from the `peer1admin.sh` etc. scripts in `network.sh`, so the start fails because they cannot be found.

jt-nti:
Fix regression in network.sh and update peer scripts to allow chaincodeListenAddress and chaincodeAddress to be overridden if required

Also updates the nano test network readme to describe the new environment variables for the k8s builder

Fixes #1198


denyeart (approved):
Thanks James!

PR #1200, opened 2024-04-29
test-network error message if jq not installed

denyeart:
With removal of fabric-tools image, test-network
now depends on jq being installed locally.

This commit logs an error message if jq commands
fail due to jq not being installed locally.


satota2 (approved):
@denyeart LGTM!

PR #1203, opened 2024-04-30
Switch microfab to hyperledger-labs version

denyeart:
full-stack-asset-transfer points to an old image.
Point to the hyperledger-labs microfab image instead.

Also, pass the orderer endpoint in lifecycle commands
instead of using the channel config fallback.


mbwhite (commented):
LGTM

mbwhite (approved):

Issue #1204, opened 2024-05-01
Test-network-k8s network script gets last version Fabric bin files instead of requested version


amiramte:
In Fabric-samples when using test-network-k8s in order to bring up cluster and network , it always downloads the latest version of Fabric binary files regardless of if any versions had been specified by any means. It may lead to unkonwn version compability issues.
To be more detailed about the reason it should be explained that  when the "network" script runs, at the very first stages the function named "check_prereqs"  is called  in which the presence of pre-requirements will be checked. In that function availability of /bin and its inner binary files is checked and in case of absence, the Fabric "bootstrap.sh" from "https://raw.githubusercontent.com/hyperledger/fabric/main/scripts/bootstrap.sh" is used to obtain the binary files and place them to /bin folder. But this "bootstrap.sh" script is called without telling which version of bin files should be grabbed so the latest version will be.
In order to solve this issue I have added two variables to the script call, one for Fabric binaries and one for Fabric CA.

denyeart:
Resolved by https://github.com/hyperledger/fabric-samples/pull/1241

PR #1205, opened 2024-05-01
Fix of issue #1204

amiramte:
Fix solution of  issue #1204 titled Test-network-k8s network script gets last version Fabric bin files instead of requested version.
When test-network-k8s is called with a specific but not the last version of Fabric hyper-ledger and the binary files are not available into ./bin folder, the script always downloaded the last version of binary files.

### DIFF
--- a/test-network-k8s/scripts/prereqs.sh
+++ b/test-network-k8s/scripts/prereqs.sh
@@ -47,7 +47,7 @@ function check_prereqs() {
   if [[ $? -ne 0 ]]; then
     echo "Downloading LATEST Fabric binaries and config"
     curl -sSL https://raw.githubusercontent.com/hyperledger/fabric/main/scripts/bootstrap.sh \
-      | bash -s -- -s -d
+      | bash -s -- ${FABRIC_VERSION} ${FABRIC_CA_VERSION} -s -d
 
     # remove sample config files extracted by the installation script
     rm config/configtx.yaml


test-network-k8s/scripts/prereqs.sh
```diff
@@ -47,7 +47,7 @@ function check_prereqs() {
   if [[ $? -ne 0 ]]; then
     echo "Downloading LATEST Fabric binaries and config"
     curl -sSL https://raw.githubusercontent.com/hyperledger/fabric/main/scripts/bootstrap.sh \
-      | bash -s -- -s -d
+      | bash -s -- ${FABRIC_VERSION} ${FABRIC_CA_VERSION} -s -d
```
denyeart, line 50: Good catch.
However, isn't FABRIC_VERSION and FABRIC_CA_VERSION a two digit number in these scripts (e.g. 2.5), whereas the bootstrap.sh script requires a three digit version for pulling binaries (e.g. 2.5.7).

amiramte: Actually in my local copy of network script, I am using 3 digit versions. However I can define a pair of auxiliary variables in order to save both 2 digit and 3 digit version's formats in network script.

jt-nti: Would it be enough to make the defaults 3 digit versions? Alternatively, it might be nice if the fabric install script accepted 2 digit versions and "did the right thing" ...and it might be good to use the [newer install script](https://github.com/hyperledger/fabric/blob/main/scripts/install-fabric.sh) instead of bootstrap.sh

denyeart:
Superseded by https://github.com/hyperledger/fabric-samples/pull/1241

PR #1206, opened 2024-05-06
Remove double quotes for env variables

denyeart:
Double quotes have caused issues in some environments, e.g. see https://github.com/hyperledger/fabric/issues/4358


satota2 (approved):
@denyeart Thank you for your update. LGTM!

PR #1207, opened 2024-05-21
Bump dependencies

denyeart:
Go to 1.22
CouchDB to 3.3.3
Fabric to 2.5.7
Fabric-CA to 1.5.10

### DIFF
--- a/.github/actions/test-network-setup/action.yaml
+++ b/.github/actions/test-network-setup/action.yaml
@@ -3,7 +3,7 @@ description: Set up the Test Network Runtime
 inputs:
   go-version:
     description: Version of go
-    default: '1.21'
+    default: '1.22'
   node-version:
     description: Version of node
     default: 18.x
@@ -12,10 +12,10 @@ inputs:
     default: 11.x
   fabric-version:
     description: Version of Hyperledger Fabric
-    default: 2.5.7
+    default: 2.5.8
   ca-version:
     description: Version of Hyperledger Fabric CA
-    default: 1.5.10
+    default: 1.5.11
 
 runs:
   using: "composite"
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -16,7 +16,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  GO_VER: '1.21'
+  GO_VER: '1.22'
   NODE_VER: 18.x
   JAVA_VER: 11.x
 
--- a/full-stack-asset-transfer-guide/infrastructure/sample-network/network
+++ b/full-stack-asset-transfer-guide/infrastructure/sample-network/network
@@ -33,8 +33,8 @@ function context() {
   export ${name}="${!override_name:-${default_value}}"
 }
 
-context FABRIC_VERSION            2.5.1
-context FABRIC_CA_VERSION         1.5.6
+context FABRIC_VERSION            2.5.8
+context FABRIC_CA_VERSION         1.5.11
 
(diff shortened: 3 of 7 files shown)


denyeart:
@bestbeforetoday The java chaincode lint failure here is new. This is the first test execution since the java chaincode 2.5.2 release last week, could that be the cause?

bestbeforetoday:
The `org.hyperledger.fabric-chaincode-java:fabric-chaincode-shim:2.5.2` release does seem to trigger this failure, although I don't obviously spot anything that changed in the chaincode-java code to cause this. It seems to be an issue with the ancient version of Mockito (5.5 years old) being used in this specific chaincode sample that fails. Updating to a newer version of Mockito makes everything happy again.

I have opened PR #1208 to update the test dependencies, which should fix this error. It solves it for locally at least.

denyeart:
@Mergifyio rebase

satota2 (changes requested):
@denyeart Thank you for the PR.
This is really a minor point, but the following line should be updated as well.
https://github.com/hyperledger/fabric-samples/blob/d3e2a90ad585f5a9903217b0286f55afc83414f4/test-network/scripts/utils.sh#L20-L21
https://github.com/hyperledger/fabric-samples/blob/d3e2a90ad585f5a9903217b0286f55afc83414f4/test-network/scripts/utils.sh#L162-L163

I think updating with another PR is an option. Please consider it.

denyeart:
@satota2 Thanks for the feedback, done, I agree it is good to keep the versions updated in the same pull request so that we can easily reference them later for future updates.

.github/actions/test-network-setup/action.yaml
```diff
@@ -3,7 +3,7 @@ description: Set up the Test Network Runtime
 inputs:
   go-version:
     description: Version of go
-    default: '1.21'
+    default: '1.22'
```
pfi79, line 6: maybe 1.22.3?

.github/workflows/lint.yaml
```diff
@@ -16,7 +16,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  GO_VER: '1.21'
+  GO_VER: '1.22'
```
pfi79, line 19: maybe 1.22.3

bestbeforetoday: Why would you want to restrict the Go version to a specific patch release and not allow the most recent patch release to be selected?

pfi79: If I read the documentation for go correctly, that's the way it's supposed to be now.

https://tip.golang.org/doc/go1.21

`
Go 1.21 introduces a small change to the numbering of releases. In the past, we used Go 1.N to refer to both the overall Go language version and release family as well as the first release in that family. Starting in Go 1.21, the first release is now Go 1.N.0. Today we are releasing both the Go 1.21 language and its initial implementation, the Go 1.21.0 release. These notes refer to “Go 1.21”; tools like go version will report “go1.21.0” (until you upgrade to Go 1.21.1). See “[Go versions](https://tip.golang.org/doc/toolchain#version)” in the “Go Toolchains” documentation for details about the new version numbering.
`

bestbeforetoday: This value is being to configure the setup-go action: https://github.com/actions/setup-go

pfi79: ok. Then don't.

satota2 (approved):
@denyeart Thank you for updating. LGTM!

PR #1210, opened 2024-05-24
Update test-network-k8s

jt-nti:
Add required permissions for upcoming k8s builder release

Also adds ttl to install jobs and configures the k8s builder prefix


satota2 (approved):
@jt-nti Thank you for the PR.
I can confirm that both v0.13.0 and the image built 3 days ago are working properly.

PR #1215, opened 2024-06-05
Add external builders for golang and node chaincode to test-network-n…

celder628:
…ano-bash

The external builders will build and launch binary chaincode instead of docker containers.


denyeart (approved):

denyeart:
Great change!

Issue #1087, opened 2023-09-15
Enhancement: Remove all samples that use legacy client SDKs


bestbeforetoday:
**As a** blockchain application developer
**I want** the client application samples to use only the Fabric Gateway client API
**So that** I am guided to use supported APIs when building my client application

The [legacy SDKs were all deprecated](https://lists.hyperledger.org/g/fabric/message/11754) following the release of Fabric v2.5 LTS. The only current supported API for building client applications is the [Fabric Gateway client API](https://hyperledger.github.io/fabric-gateway/). To avoid confusion for users, and reduce the maintenance and build resource burden for this repository, all of the samples that make use of the legacy client SDKs could be removed, leaving only the client application samples that use the Fabric Gateway client API.

bestbeforetoday:
Closing as completed since all legacy samples that have Gateway equivalents have been removed. Conversion (or removal) of remaining legacy samples can be tracked elsewhere.

PR #1216, opened 2024-06-07
Remove legacy sample applications

bestbeforetoday:
The removed samples make use of deprecated legacy client SDKs. They all have equivalent samples implemented using the currently supported Fabric Gateway client API, and are therefore redundant.

Contributes to #1087


asset-transfer-basic/chaincode-external/README.md
```diff
@@ -159,16 +159,14 @@ Now that we have started the chaincode service and deployed it to the channel, w
 
 ## Using the Asset-Transfer-Basic external chaincode
 
-Open yet another terminal and navigate to the `fabric-samples/asset-transfer-basic/application-javascript` directory:
+Open yet another terminal and navigate to the `fabric-samples/asset-transfer-basic/application-gateway-go` directory:
 ```
-cd fabric-samples/asset-transfer-basic/application-javascript
+cd fabric-samples/asset-transfer-basic/application-gateway-go
```
denyeart, line 164: Any reason to use application-gateway-go instead of application-gateway-typescript? It seems more new users are familiar with javascript/typescript which is probably why the old instructions used javascript.

denyeart (approved):
Thanks, this helps focus new users on the right SDKs. 
Just one minor question, I'll go ahead and merge.

asset-transfer-basic/chaincode-external/README.md
bestbeforetoday: Any of the client application languages implementations will work fine. I figured that, since this chaincode sample uses Go, the Go client application probably made more sense. One language rather than two.

PR #1219, opened 2024-06-13
Use more specific chaincode package versions

bestbeforetoday:
This demonstrates good practice in restricting the chaincode package versions to those applicable for the specific Fabric (major/minor) version targeted for deployment.

Also some corrections to the repository README. Particularly referring to other branches for samples targeted at earlier Fabric versions, since samples in the main branch may exploit features not available in older Fabric releases, which can cause confusion for end users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,12 @@
 
 # Hyperledger Fabric Samples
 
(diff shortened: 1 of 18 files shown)


denyeart (approved):
Thanks!

PR #1221, opened 2024-06-14
Add chaincode logging for binary chaincode in test-network-nano-bash

celder628:
- Correct sh compability in ca_utils.sh
- Correct linux compatibility in external builder


denyeart (approved):
Excellent!

PR #1230, opened 2024-06-23
Fix directory path for chaincode-external

urgetolearn:


denyeart (approved):
Good catch, thank you!

PR #1241, opened 2024-07-31
test-network-k8s: Improve prereqs logic

### LINKED ISSUE #1204: Test-network-k8s network script gets last version Fabric bin files instead of requested version

satota2:
It seems that PR #1205 has stalled, so I have created a new PR that builds upon the considerations from that one while addressing the issues raised. 

This PR resolves issue #1204 and also addresses the concerns mentioned in PR #1205. Specifically, it:
- Uses the newer install script instead of bootstrap.sh.
- Downloads binaries corresponding to the image versions, supporting both two-digit and three-digit Fabric version specifications.

Related issue:
Resolves #1204

### DIFF
--- a/test-network-k8s/scripts/prereqs.sh
+++ b/test-network-k8s/scripts/prereqs.sh
@@ -42,17 +42,41 @@ function check_prereqs() {
     exit 1
   fi
 
-  # Use the local fabric binaries if available.  If not, go get them.
+  # Define the sed expression to extract the version number
+  VERSION_SED_EXPR='s/^ Version: v\?\(.*\)$/\1/p'
+
+  # Use the fabric peer and ca containers to check fabric image versions
+  # NOTE: About extracting the version number:
+  # In older versions, the prefix 'v' was not included in the version string,
+  # but in recent versions, 'v' has been added.
+  # The following commands remove the optional 'v' to standardize the format.
+  FABRIC_IMAGE_VERSION=$(${CONTAINER_CLI} run --rm ${FABRIC_PEER_IMAGE} peer version | sed -ne "$VERSION_SED_EXPR")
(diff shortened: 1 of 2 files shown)


bestbeforetoday (approved):
The same `sed` expression to extract a version number is repeated several times. Perhaps for ease of maintenance this could be extracted to a function so there is only one place to change it in the future? Otherwise this looks like a good change to me

satota2:
> The same `sed` expression to extract a version number is repeated several times. Perhaps for ease of maintenance this could be extracted to a function so there is only one place to change it in the future? Otherwise this looks like a good change to me

@bestbeforetoday 
Apologies for the delayed response; I was on summer vacation in Japan.

Thank you for your review. You are absolutely right. I've updated the code to extract the regular expression into a variable for easier maintenance. Please take a look when you have a moment.

bestbeforetoday (approved):
👍

denyeart (approved):

PR #1248, opened 2024-09-18
Bump ubuntu to 22.04 in CI

denyeart:
Since Fabric v3.0 builds on ubuntu 22.04,
it will be necessary for samples CI to run on ubuntu 22.04.

Both Fabric v2.5 components (ubuntu 20.04) and
Fabric v3.0 components (ubuntu 22.04) work on
ubuntu 22.04 runtime.

The update also requires shell script updates to pass linting.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -22,7 +22,7 @@ env:
 
 jobs:
   go:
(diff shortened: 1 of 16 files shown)


bestbeforetoday (approved):
Looks good to me

satota2 (approved):
@denyeart LGTM!

PR #1250, opened 2024-09-19
remove golang toolchain as nit fix

SamYuan1990:
by remove golang tool chain from go.mod, try to fix https://github.com/hyperledger/fabric/issues/4951

### DIFF
--- a/.github/actions/test-network-setup/action.yaml
+++ b/.github/actions/test-network-setup/action.yaml
@@ -3,7 +3,7 @@ description: Set up the Test Network Runtime
 inputs:
   go-version:
     description: Version of go
-    default: '1.22'
+    default: '1.21'
   node-version:
     description: Version of node
     default: 18.x
--- a/asset-transfer-basic/chaincode-go/go.mod
+++ b/asset-transfer-basic/chaincode-go/go.mod
@@ -2,8 +2,6 @@ module github.com/hyperledger/fabric-samples/asset-transfer-basic/chaincode-go
 
 go 1.21
 
-toolchain go1.22.4
-
 require (
 	github.com/hyperledger/fabric-chaincode-go/v2 v2.0.0-20240618210511-f7903324a8af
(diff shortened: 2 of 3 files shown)


.github/actions/test-network-setup/action.yaml
```diff
@@ -3,7 +3,7 @@ description: Set up the Test Network Runtime
 inputs:
   go-version:
     description: Version of go
-    default: '1.22'
```
bestbeforetoday, line 6: I think the Go version was intentionally moved to 1.22 since 1.21 is no longer supported since Go 1.23 was released. See the [Go release policy](https://go.dev/doc/devel/release#policy).

asset-transfer-basic/chaincode-go/go.mod
```diff
@@ -2,8 +2,6 @@ module github.com/hyperledger/fabric-samples/asset-transfer-basic/chaincode-go
 
 go 1.21
 
-toolchain go1.22.4
```
bestbeforetoday, line 5: I think I caused this when I bulk updated dependencies to the v2 chaincode APIs. By default the `go mod` command sets the toolchain to the specific Go version is used to run the command, which might impose more restrictive version constraints than the `go` directive.

In [other repositories](https://github.com/hyperledger/fabric-gateway/commit/798c6ef971820ea461c2e8226dcbe55613c12cae) we have been taking the approach of avoiding a `toolchain` directive but specifying a full (3-digit) version for the `go` directive. For current Go versions, the `go` version is used as the `toolchain` if no toolchain is specified, and toolchain must be a full (3-digit) Go version. That might be the right thing to do here too.

.github/actions/test-network-setup/action.yaml
SamYuan1990: good point, how about we upgrade go mod files into golang 1.22?

bestbeforetoday:
@SamYuan1990 I think I removed all the **toolchain** directives from go.mod files in PR #1260.

PR #1261, opened 2024-10-18
Remove cli from podman compose file

### LINKED ISSUE #1258: CLI to be removed from podman compose files
Now that CLI is deprecated, I think we should remove the reference to CLI from the compose files. This is preventing the network from starting when we set CONTAINER_CLI = podman

The file to be changed is given below. 

https://github.com/hyperledger/fabric-samples/blob/main/test-network/compose/podman/podman-compose-test-net.yaml

twoGiants:
I removed the `cli` entry from the `podman-compose-test-net.yaml` file.

This PR closes #1258.

### DIFF
--- a/test-network/compose/podman/podman-compose-test-net.yaml
+++ b/test-network/compose/podman/podman-compose-test-net.yaml
@@ -14,6 +14,3 @@ services:
     volumes:
       - ./podman/peercfg:/etc/hyperledger/peercfg
 
-  cli:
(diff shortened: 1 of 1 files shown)


bestbeforetoday:
It looks like this change was missed from PR #1186, which removed the CLI container from the Docker version of these files, so seems like a good change to me.

denyeart (approved):

PR #1269, opened 2024-11-11
Add off-chain-data go client application

twoGiants:
### Off Chain Data Go Application

This PR contains an implementation of the off-chain data store sample in go using the block event listening capability of the [Fabric Gateway client API](https://hyperledger.github.io/fabric-gateway/). The issue reference is [here](https://github.com/hyperledger/fabric-gateway/issues/190) in the Fabric Gateway SDK repository.

It follows the same procedure and uses mostly the same naming as the off-chain data typescript sample [here](https://github.com/hyperledger/fabric-samples/tree/main/off_chain_data/application-typescript). There are some differences in the go package structure compared to the typescript modules and I added a couple of tests to the parser which helped me better understand the unpacking of the block data structure. The Go implementation is more similar to the Java version. More about that below in the [Differences](#differences) section.

@bestbeforetoday gave me this useful [link](https://github.com/hyperledger/fabric/tree/main/internal/pkg/gateway/event) where I found the [iterator_test.go](https://github.com/hyperledger/fabric/blob/main/internal/pkg/gateway/event/iterator_test.go) to be helpful and this [link](https://github.com/hyperledger/fabric-sdk-node/blob/main/fabric-common/lib/BlockDecoder.js#L135) where the "Endorser Transaction" comments were helpful.

### Summary

- [x] created project directory, `app.go,` `getAllAssets.go`, `transact.go` and `listen.go` files
- [x] reused and changed the logic for `connect.go` from the private data application
- [x] implemented the asset transfer basic contract in the `contract` package
- [x] implemented  `getAllAssets.go` and `transact.go`
- [x] implemented the block parser and all the supporting data structures
- [x] implemented the block processor and all the supporting data structures
- [x] implemented `listen.go`
- [x] updated README.md with the command to run the go application
- [x] updated script which runs the application in the Github Actions workflow
- [x] manual tests

### Differences
- I didn't use any interfaces. I removed them after the suggestion of @bestbeforetoday.
- I did not implement any methods which are not used anywhere, like `toProto` or `getSignatureHeader`.
- The block parser with all it's underlying structs (which are mostly private) is in the `parser` package.
- There are some simple tests for the cache utility and the block unwrapping.
- The block processor with it's underlying struct `transaction` (which is private) is in the `processor` package.
- The store is in the `store` package and because it resembles a bit a flat file db I called it `flatFile.go`
- The utilities are in the `utils` package.
- In all the additional packages I followed the encapsulation rule: if it's not referenced outside the package it's private.

### Additional Information
I process the block events in a go routine and use context to stop the processing. Instead of panicking everywhere down the line where an error can occur I handle the errors by wrapping them, returning them, log them in the go routine, then return and shutdown gracefully. Panicking results in the context, gateway and checkpointer not getting closed.

### Questions
1. [Here](https://github.com/hyperledger/fabric-samples/pull/1269/files#diff-f3638c01da62938f42f45b28e2dac48c23cd8e3a72a8f50eeef6ff220657f157R51) is a TODO proposing to simplify the `processor.transaction` data structure.
1. And [here](https://github.com/hyperledger/fabric-samples/pull/1269/files#diff-65188beb179d45b93019f1b532a5767482b774d41ac7395aecbe66d4e1e26a00R51) is a TODO proposing to encapsulate the final unwrapping of the `kvWrite` in the parser instead of having it in `transaction.writes`.
1. I assume I should squash everything to one commit? It might be a big commit message.


off_chain_data/application-go/blockParser.go
```diff
@@ -0,0 +1,397 @@
...
+	block           *common.Block
+	validationCodes []byte
+	transactions    []Transaction
+}
+
+func NewParsedBlock(block *common.Block) Block {
+	validationCodes := getTransactionValidationCodes(block)
+
+	return &ParsedBlock{block, validationCodes, nil}
+}
```
bestbeforetoday, line 31: A more typical pattern in Go is to return concrete structs, and use interfaces for parameter types received by functions. Interfaces are typically used for return types only where different implementation structs might be returned.

For this case you might be better not to have an interface at all.

```suggestion
type Block struct {
	block           *common.Block
	validationCodes []byte
	transactions    []Transaction
}

func ParseBlock(block *common.Block) Block {
	validationCodes := getTransactionValidationCodes(block)

	return &Block{block, validationCodes, nil}
}
```

To make it clearer you might consider moving the parsing code (and associated structs) into a different package. Then each Block variable in the application code will be referred to as either `common.Block` or `parser.Block` (for example).

off_chain_data/application-go/blockParser.go
```diff
@@ -0,0 +1,397 @@
...
+
+func (pb *ParsedBlock) GetNumber() uint64 {
+	header := assertDefined(pb.block.GetHeader(), "missing block header")
+	return header.GetNumber()
+}
+
+// TODO: needs cache, getPayloads, parsePayload
+func (pb *ParsedBlock) GetTransactions() []Transaction {
```
bestbeforetoday, line 39: Go typically doesn't use `Get` for getters (although it can make sense in some cases). A more typical naming would just be `Transactions`. See [Effective Go](https://go.dev/doc/effective_go#Getters).

```suggestion
func (pb *ParsedBlock) Transactions() []Transaction {
```

off_chain_data/application-go/blockParser.go
```diff
@@ -0,0 +1,397 @@
...
+	return nil
+}
+
+func (pb *ParsedBlock) ToProto() *common.Block {
+	return nil
+}
+
+// Implements identity.Identity Interface
+type IdentityImpl struct {
```
bestbeforetoday, line 48: If it's really an implementation class only created by implementation of this package then I would make it private (first character lower case). Often an interface is not used and the concrete class is returned, but reusing the fabric-gateway Identity interface might make sense here.

off_chain_data/application-go/go.mod
```diff
@@ -0,0 +1,22 @@
+module offChainData
+
+go 1.22.7
+
+toolchain go1.23.0
```
bestbeforetoday, line 5: I think I would recommend setting the minimum required Go version to 1.22.0 and removing the toolchain so it builds with the locally installed version of Go, provided it is at least the minimum required version. You can do that by running:

```shell
go get go@1.22.0 toolchain@none
```

twoGiants: Thx. Will do.

off_chain_data/application-go/blockParser.go
twoGiants: > A more typical pattern in Go is to return concrete structs, and use interfaces for parameter types received by functions.

Thx, didn't knew. Will do that.

> Interfaces are typically used for return types only where different implementation structs might be returned.

Of course, that is right and not only in go. Will keep it in mind.

> To make it clearer you might consider moving the parsing code (and associated structs) into a different package. Then each Block variable in the application code will be referred to as either `common.Block` or `parser.Block` (for example).

Yes, that is exactly my plan once I make it work.

off_chain_data/application-go/blockParser.go
twoGiants: Yes yes, this was also bothering me. I remember that one well. Getters and setters aren't a good pattern anyway imho. I agree mostly with Allen Holub on this one => good read [Why getter and setter methods are evil](https://www.infoworld.com/article/2161183/why-getter-and-setter-methods-are-evil.html). But I would not go as far as not using them at all.

off_chain_data/application-go/blockParser.go
twoGiants: I will review and update the encapsulation once I refactored the code into the parser package.

off_chain_data/application-go/blockParser.go
twoGiants: > For this case you might be better not to have an interface at all.

The TypeScript implementation is using interfaces. I thought I was supposed to follow it closely?

I will implement a version without interfaces and see how it is.

twoGiants:
Don't close the PR. I am working on it. New commits are coming CW 52.

off_chain_data/application-go/listen.go
```diff
@@ -0,0 +1,93 @@
...
+
+				if err := blockProcessor.Process(); err != nil {
+					fmt.Println("\033[31m[ERROR]\033[0m", err)
+					return
+				}
+			}
+		}
+	}()
+
+	wg.Wait()
```
bestbeforetoday, line 91: I'm not sure of the reason for using a wait group for a single goroutine that you then wait on before continuing. Maybe this can just be inline code?

It should not be necessary to check the context while reading. If the context used for the `BlockEvents` call is closed, the returned channel should also get closed, which would mean the range loop will exit.

```suggestion
	for blockProto := range blocks {
		blockProcessor := processor.NewBlock(
			parser.ParseBlock(blockProto),
			checkpointer,
			store.ApplyWritesToOffChainStore,
			channelName,
		)

		if err := blockProcessor.Process(); err != nil {
			fmt.Println("\033[31m[ERROR]\033[0m", err)
			return
		}
	}

```

off_chain_data/application-go/go.mod
```diff
@@ -0,0 +1,20 @@
+module offChainData
```
bestbeforetoday, line 1: By convention Go package names are all lowercase. See https://go.dev/blog/package-names

bestbeforetoday (changes requested):
First thing, it runs great for me so this looks nicely done. Thank you! Sorry for large number of comments. This is just polishing the material to make it a really clean sample.

I would be careful to try and keep all the code that listens for blocks, navigates to their transactions and passes the read/write sets to some kind of store close together. Perhaps all in `listen.go`. This is the key learning point of the sample so it needs to be easy to access and navigate. The block parser should be separate (as you've done) so it is a reusable package on its own. The store implementation can also be a little saperate to avoid cluttering the key listener code. I might put it in a different file within the same package though, just as private (lowercase) content.

The **Application** section of `off_chain_data/README.md` should be updated to include references to key parts of the Go implementation, similar to the existing TypeScript and Java entries.

To catch some issues, it might be worth running this command in the `application-go` directory:

```shell
go run honnef.co/go/tools/cmd/staticcheck@latest -f stylish -checks all ./...
```

To avoid specific checks (like having package-level godoc), you can exclude specific rules. For example:

```shell
go run honnef.co/go/tools/cmd/staticcheck@latest -f stylish -checks 'all,-ST1000' ./...
```

off_chain_data/application-go/app.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"google.golang.org/grpc"
+)
+
+var allCommands = map[string]func(clientConnection *grpc.ClientConn){
```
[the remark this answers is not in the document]
twoGiants: Omitted.

off_chain_data/application-go/processor/block.go
```diff
@@ -0,0 +1,147 @@
...
+		if err := aTransaction.process(); err != nil {
+			return fmt.Errorf("in %s: %w", funcName, err)
+		}
+
+		channelHeader, err := validTransaction.ChannelHeader()
+		if err != nil {
+			return fmt.Errorf("in %s: %w", funcName, err)
+		}
+		transactionId := channelHeader.GetTxId()
+		b.checkpointer.CheckpointTransaction(blockNumber, transactionId)
```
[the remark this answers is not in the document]
twoGiants: Thx. Done.

off_chain_data/application-go/processor/block.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	b.checkpointer.CheckpointBlock(blockNumber)
```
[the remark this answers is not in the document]
twoGiants: Thx. Done.

off_chain_data/application-go/contract/model.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return Asset{
+		ID:             id.String(),
+		Color:          utils.RandomElement(colors),
+		Size:           uint64(utils.RandomInt(maxInitialSize) + 1),
+		Owner:          utils.RandomElement(Owners),
+		AppraisedValue: uint64(utils.RandomInt(maxInitialValue) + 1),
+	}
+}
```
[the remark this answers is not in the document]
twoGiants: I am also using it in the parser/block_test.go. I will still move it to transact then and use a simpler construction in the test. Ok?

off_chain_data/application-go/store/model.go
```diff
@@ -0,0 +1,25 @@
+package store
+
+// Apply writes for a given transaction to off-chain data store, ideally in a single operation for fault tolerance.
+type Writer = func(data LedgerUpdate) error
+
+// Ledger update made by a specific transaction.
+type LedgerUpdate struct {
+	BlockNumber   uint64
+	TransactionId string
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/processor/block.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		if err := aTransaction.process(); err != nil {
+			return fmt.Errorf("in %s: %w", funcName, err)
+		}
+
+		channelHeader, err := validTransaction.ChannelHeader()
+		if err != nil {
+			return fmt.Errorf("in %s: %w", funcName, err)
+		}
+		transactionId := channelHeader.GetTxId()
```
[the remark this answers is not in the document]
twoGiants: > Go style prefers all capitals for [well-known initialisms](https://go.dev/wiki/CodeReviewComments#initialisms).

This seems not to apply to the fabric-protos-go-apiv2 package: `channelHeader.GetTxId()`.

Done.

off_chain_data/application-go/processor/block.go
```diff
@@ -0,0 +1,147 @@
...
+	blockTransactionIds := []string{}
+	for _, transaction := range transactions {
+		channelHeader, err := transaction.ChannelHeader()
+		if err != nil {
+			return 0, fmt.Errorf("in %s: %w", funcName, err)
+		}
+		blockTransactionIds = append(blockTransactionIds, channelHeader.GetTxId())
+	}
+
+	lastTransactionId := b.checkpointer.TransactionID()
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/utils/utils.go
```diff
@@ -0,0 +1,74 @@
...
+// Pick a random element from an array, excluding the current value.
+func DifferentElement(values []string, currentValue string) string {
+	candidateValues := []string{}
+	for _, v := range values {
+		if v != currentValue {
+			candidateValues = append(candidateValues, v)
+		}
+	}
+	return RandomElement(candidateValues)
+}
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/utils/utils.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Return the value if it is defined; otherwise panics with an error message.
+func AssertDefined[T any](value T, message string) (T, error) {
+	if any(value) == any(nil) {
+		var zeroValue T
+		return zeroValue, errors.New(message)
+	}
+
+	return value, nil
+}
```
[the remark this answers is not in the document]
twoGiants: Understood. Removed.

off_chain_data/application-go/utils/utils.go
```diff
@@ -0,0 +1,74 @@
...
+			value, err = f()
+			if err != nil {
+				var zeroValue T
+				return zeroValue, fmt.Errorf("in Cache: %w", err)
+			}
+			cached = true
+		}
+		return value, nil
+	}
+}
```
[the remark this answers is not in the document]
twoGiants: > everywhere I see it used, you are immediately invoking the returned cache function and then discarding the cache function. No caching is actually happening.

I am not following. Can you also check the tests `utils_test.go:9, Test_cachePrimitiveFunctionResult`. The cache should work as tested there. If I check the usage of the cache function in the `parser.Block.Transactions` method what should happen is that when `Transactions` is called the first time `utils.Cache` will run the first time and cache the the result of `func() ([]*Transaction, error)`. Next time when `Transactions` is called the cached value is provided. Am I missing something?

I will test the execution times with and without the cache function just to be sure. Because I thought to have seen a difference once I added it. Before it was slowly unpacking each transaction and after adding the cache it became nearly instant.

off_chain_data/application-go/contract/model.go
```diff
@@ -0,0 +1,45 @@
...
+
+var (
+	colors = []string{"red", "green", "blue"}
+	Owners = []string{"alice", "bob", "charlie"}
+)
+
+const (
+	maxInitialValue = 1000
+	maxInitialSize  = 10
+)
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/contract/contract.go
```diff
@@ -0,0 +1,71 @@
...
+	); err != nil {
+		return fmt.Errorf("in DeleteAsset: %w", err)
+	}
+	return nil
+}
+
+func (atb *AssetTransferBasic) GetAllAssets() ([]byte, error) {
+	result, err := atb.contract.Evaluate("GetAllAssets")
+	if err != nil {
+		return []byte{}, fmt.Errorf("in GetAllAssets: %w", err)
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/parser/block.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"offChainData/utils"
+
+	"github.com/hyperledger/fabric-protos-go-apiv2/common"
+	"google.golang.org/protobuf/proto"
+)
+
+type Block struct {
+	block        *common.Block
+	transactions []*Transaction
```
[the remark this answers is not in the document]
twoGiants: Removed.

off_chain_data/application-go/go.mod
twoGiants: Done.

off_chain_data/application-go/listen.go
twoGiants: You're right. It's not needed. There must've been a reason why I put it there while developing but I can't recall.

I removed it.

off_chain_data/application-go/processor/block.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (b *block) getNewTransactions() ([]*parser.Transaction, error) {
+	funcName := "getNewTransactions"
+
+	transactions, err := b.parsedBlock.Transactions()
+	if err != nil {
+		return nil, fmt.Errorf("in %s: %w", funcName, err)
+	}
+
+	lastTransactionId := b.checkpointer.TransactionID()
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/utils/utils.go
```diff
@@ -0,0 +1,74 @@
...
+
+func EnvOrDefault(key, defaultValue string) string {
+	result := os.Getenv(key)
+	if result == "" {
+		return defaultValue
+	}
+	return result
+}
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/utils/utils.go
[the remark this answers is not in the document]
twoGiants: > No caching is actually happening.

I figured it out. I added a `cached*` property on the structs everywhere the cache is used an initialize it now if it's `nil` with the cached function. Is that ok?

off_chain_data/application-go/parser/block.go
```diff
@@ -0,0 +1,101 @@
...
+
+	payloads, err := b.parse(commonPayloads)
+	if err != nil {
+		return nil, fmt.Errorf("in %s: %w", funcName, err)
+	}
+
+	b.cachedTransactions = b.createTransactionsFrom(payloads)
+
+	return b.cachedTransactions, nil
+}
```
bestbeforetoday, line 47: I think there might be a simpler way to do the lazy unmarshal and caching of results by using the [sync](https://pkg.go.dev/sync) package from the standard library. This might be worth trying:

```suggestion
type Block struct {
	block        *common.Block
	transactions func() ([]*Transaction, error)
}

func ParseBlock(block *common.Block) *Block {
	result := &Block{
		block: block,
	}
	result.transactions = sync.OnceValues(result.unmarshalTransactions)

	return result
}

func (b *Block) Number() uint64 {
	return b.block.GetHeader().GetNumber()
}

func (b *Block) Transactions() ([]*Transaction, error) {
	return b.transactions()
}

func (b *Block) unmarshalTransactions() ([]*Transaction, error) {
	envelopes, err := b.unmarshalEnvelopesFromBlockData()
	if err != nil {
		return nil, err
	}

	commonPayloads, err := b.unmarshalPayloadsFrom(envelopes)
	if err != nil {
		return nil, err
	}

	payloads, err := b.parsePayloads(commonPayloads)
	if err != nil {
		return nil, err
	}

	return b.createTransactionsFrom(payloads), nil
}
```

off_chain_data/application-go/parser/block.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (b *Block) unmarshalEnvelopesFromBlockData() ([]*common.Envelope, error) {
```
bestbeforetoday, line 49: When it's a method on Block, it seems like repetition to say "FromBlockData". Perhaps keep it simple?

```suggestion
func (b *Block) unmarshalEnvelopes() ([]*common.Envelope, error) {
```

off_chain_data/application-go/parser/block.go
```diff
@@ -0,0 +1,101 @@
...
+		envelope := &common.Envelope{}
+		if err := proto.Unmarshal(blockData, envelope); err != nil {
+			return nil, fmt.Errorf("in unmarshalEnvelopesFromBlockData: %w", err)
+		}
+		result = append(result, envelope)
+	}
+	return result, nil
+}
+
+func (*Block) unmarshalPayloadsFrom(envelopes []*common.Envelope) ([]*common.Payload, error) {
```
bestbeforetoday, line 61: Since the Block isn't used this doesn't need to be method on Block. Maybe it can just be called `unmarshalPayloads` too. If you prefer more explanation then `unmarshalPayloadsFromEnvelopes`. Go doesn't have function/method overloading based on parameter types. The name must be unique.

off_chain_data/application-go/parser/block.go
```diff
@@ -0,0 +1,101 @@
...
+
+	result := []*payload{}
+	for i, commonPayload := range commonPayloads {
+		statusCode := validationCodes[i]
+
+		payload := parsePayload(commonPayload, int32(statusCode))
+		is, err := payload.isEndorserTransaction()
+		if err != nil {
+			return nil, fmt.Errorf("in %s: %w", funcName, err)
+		}
```
bestbeforetoday, line 86: I appreciate that you're following the TypeScript implementation structure pretty closely here, but the error handling in Go makes this pretty ugly. Since we are always unmarshaling the payload's channel header anyway to check if it's an endorser transactions, how about the `parsePayload` does that work up-front? Then `isEndorserTransaction` can just return a `bool` and no error, which is much simpler.

Perhaps the TypeScript implementation should be doing this too, since the cache doesn't save any work, just adds complexity. That can be tackled another time.

bestbeforetoday (changes requested):
Thank you for the excellent work. This is a great improvement on the last time I looked at the code. Sorry for the delay and having lots of comments... again. Most are very minor suggestions but a few I think are more important, like the error handling to ensure execution surfaces errors correctly to the automated build.

Let me know if any of the suggestions are not clear enough. I am happy to help with any tricky changes.

Once again, great work!

off_chain_data/application-go/contract/contract.go
```diff
@@ -0,0 +1,71 @@
...
+		client.WithArguments(
+			id,
+		),
+	); err != nil {
+		return fmt.Errorf("in DeleteAsset: %w", err)
+	}
+	return nil
+}
+
+func (atb *AssetTransferBasic) GetAllAssets() ([]byte, error) {
```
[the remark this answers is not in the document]
twoGiants: Like that?

```go
func (atb *AssetTransferBasic) GetAllAssets() ([]Asset, error) {
	assetsRaw, err := atb.contract.Evaluate("GetAllAssets")
	if err != nil {
		return nil, err
	}

	if len(assetsRaw) == 0 {
		return []Asset{}, nil
	}

	var assets []Asset
	if err := json.Unmarshal(assetsRaw, &assets); err != nil {
		return nil, err
	}

	return assets, nil
}
```

off_chain_data/application-go/getAllAssets.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	if len(assets) == 0 {
+		fmt.Println("no assets")
+		return
+	}
+
+	fmt.Println(formatJSON(assets))
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/transact.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func newTransactApp(smartContract *atb.AssetTransferBasic) *transactApp {
+	return &transactApp{smartContract, 10}
+}
+
+func (t *transactApp) run() {
+	var wg sync.WaitGroup
+
+	for i := 0; i < t.batchSize; i++ {
```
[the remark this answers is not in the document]
twoGiants: `i` will be unused. The linter and compiler won't accept it.

off_chain_data/application-go/transact.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Generate a random integer in the range 0 to max - 1.
+func randomInt(max int) int {
+	result, err := rand.Int(rand.Reader, big.NewInt(int64(max)))
+	if err != nil {
+		panic(err)
+	}
+
+	return int(result.Int64())
+}
```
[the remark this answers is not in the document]
twoGiants: Yes, it works fine. Thanks! 

Updated.

off_chain_data/application-go/transact.go
```diff
@@ -0,0 +1,135 @@
...
+
+// Pick a random element from an array, excluding the current value.
+func differentElement(values []string, currentValue string) string {
+	candidateValues := []string{}
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/transact.go
```diff
@@ -0,0 +1,135 @@
...
+		return fmt.Errorf("in %s: %w", funcName, err)
+	}
+	fmt.Println("Created asset", anAsset.ID)
+
+	// Transfer randomly 1 in 2 assets to a new owner.
+	if randomInt(2) == 0 {
+		newOwner := differentElement(owners, anAsset.Owner)
+		oldOwner, err := t.smartContract.TransferAsset(anAsset.ID, newOwner)
+		if err != nil {
+			return fmt.Errorf("in %s: %w", funcName, err)
```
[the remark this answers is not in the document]
twoGiants: > Same comment for other instances of this error handling style.

I removed the wrapping in `transact.go` but have left it in `listen.go` and everywhere down the line. Are you asking for the `transact.go` or for every instance in this PR?

off_chain_data/application-go/contract/contract.go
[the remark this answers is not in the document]
bestbeforetoday: I haven't tried it but that looks good at first glance. Just remember that `nil` is equivalent to an empty slice — you don't need to actually allocate a slice for an empty one — so instead of:

```go
return []Asset{}, nil
```

I think you should be able to do:

```go
return nil, nil
```

off_chain_data/application-go/transact.go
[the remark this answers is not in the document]
bestbeforetoday: If it makes sense to you to wrap, that's fine. My concern is that these particular messages just add the name of the function in the code where the error is raised, which looks to me like debug information rather than explanatory end user error messages. The function names will already appear in the stack trace printed when panic is called with an error.

If it makes sense to add some explanatory contextual text to a lower level error so that it makes more sense for the user, go ahead and wrap. If not, and especially for places where we really don't ever expect to see an error when running the sample, I'd be tempted to just pass the underlying error up so there is less cognitive load on somebody reading the code.

bestbeforetoday:
I've answered a few questions you raised (and an error you spotted in my suggestion!). I've also tried to close all the comment already covered to it's easier to see what might still need to be considered. Please let me know if you want to discuss or pair on any any changes.

It would be great to solve the more important comments, like the wait group error handling, so we can get this in. It's a great and **really** useful sample.

off_chain_data/application-go/contract/contract.go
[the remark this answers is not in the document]
twoGiants: > return nil, nil

Yes, I will remember it. Corrected it. Thanks!

off_chain_data/application-go/transact.go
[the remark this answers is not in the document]
twoGiants: Done.

twoGiants:
> I've answered a few questions you raised (and an error you spotted in my suggestion!). I've also tried to close all the comment already covered to it's easier to see what might still need to be considered. Please let me know if you want to discuss or pair on any any changes.
> 
> It would be great to solve the more important comments, like the wait group error handling, so we can get this in. It's a great and **really** useful sample.

Thanks for answering the questions I raised Mark. I had to suddenly interrupt my work and was not able to address the other comments you wrote and commit what I have fixed so far.

Every comment you wrote will be addressed and implemented. I always start with the easy ones and work my way up to the more complex and important ones. Maybe a couple more questions will arise, I will let you know.

off_chain_data/application-go/app.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"google.golang.org/grpc"
+)
+
+var allCommands = map[string]func(*grpc.ClientConn){
```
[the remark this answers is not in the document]
twoGiants: 1. Agree. Done.
2. I think you're right. I pass all the errors anyway up the call chain. Lets handle them here centrally.
3. I like the type alias more. Done.

off_chain_data/application-go/transact.go
```diff
@@ -0,0 +1,135 @@
...
+func (t *transactApp) run() {
+	var wg sync.WaitGroup
+
+	for i := 0; i < t.batchSize; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			if err := t.transact(); err != nil {
+				fmt.Println("\033[31m[ERROR]\033[0m", err)
```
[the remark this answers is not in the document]
twoGiants: > Another option might be to create a Context [WithCancelCause](https://pkg.go.dev/context@go1.22.0#WithCancelCause) along with the WaitGroup, and pass any errors to the context's CancelCauseFunc. After waiting for the WaitGroup, you can check if the context has an associated error and pass that up.

I implemented this option and it works fine. The only issue I see now is that we don't know where the error originates. Any error we encounter will be logged by the panic in app.go without any stack trace or additional info where it came from.

How would you go about it?

off_chain_data/application-go/store.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	result := math.Floor(valueAsFloat)
+	if valueAsFloat < 0 {
+		panic(fmt.Errorf("invalid SIMULATED_FAILURE_COUNT value: %s", valueAsString))
+	}
+
+	return uint(result)
```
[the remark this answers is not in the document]
twoGiants: Yes, much better. Done.

off_chain_data/application-go/parser/block.go
twoGiants: Yes, this is great. I added it everywhere.

off_chain_data/application-go/transact.go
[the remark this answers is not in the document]
twoGiants: > If it makes sense to you to wrap, that's fine. My concern is that these particular messages just add the name of the function in the code where the error is raised, which looks to me like debug information rather than explanatory end user error messages. The function names will already appear in the stack trace printed when panic is called with an error.

I am not sure here. I welcome your advice. I added it because I did not panic in the parser but returned the error and logged it in the `listen.go` function. I thought it was helpful to see where the errors originates. I added the function name as I have seen it in [this book](https://learning.oreilly.com/library/view/learning-go-2nd/9781098139285/ch09.html#id109).

> ...especially for places where we really don't ever expect to see an error when running the sample, I'd be tempted to just pass the underlying error up so there is less cognitive load...

Yes that make sense. I removed it.

Do you have a rule of thumb where to wrap errors and add contextual info?

off_chain_data/application-go/parser/block.go
twoGiants: Agree. Done.

off_chain_data/application-go/parser/block.go
twoGiants: > Maybe it can just be called `unmarshalPayloads` too.

If I read the code where the method is called it reads like that: `b.unmarshalPayloads(envelopes)`. I think the current name reads better: `b.unmarshalPayloadsFrom(envelopes)`

> If you prefer more explanation then `unmarshalPayloadsFromEnvelopes`.

Now it would sound: `... b.unmarshalPayloadsFromEnvelopes(envelopes)`. I still think the current name reads better.

I left the method on the struct because it is a step in the pipeline of unwrapping the block and does not make sense on its own. It is same in `endorserTransaction.go` and the methods there. Making those steps separate functions feels like breaking cohesion even though its in the same file. Wdyt?

But I want to follow good practices and idiomatic go. Let me know if I should extract it from the struct.

off_chain_data/application-go/parser/block.go
twoGiants: Changed it. Let me know if that's ok.

twoGiants:
I just pushed the last changes. The rework is done. All suggestions were clear but lets see if I understood and implemented correctly. Don't worry about new comment, I appreciate the reviews. It's a good way for me to grow my skills.

off_chain_data/application-go/app.go
```diff
@@ -0,0 +1,61 @@
...
+
+	client := newGrpcConnection()
+	defer client.Close()
+
+	for _, name := range commands {
+		command := allCommands[name]
+
+		if err := command(client); err != nil {
+			panic(err)
+		}
```
bestbeforetoday, line 43: You will want to avoid the panic when you get the write simulation error. Perhaps something like this:

```suggestion
		if err := command(client); err != nil {
			if errors.Is(err, errExpected) {
				fmt.Println(err)
				return
			}

			panic(err)
		}
```

bestbeforetoday (commented):
A suggestion to handle the expected error without panic when you simulate a write failure. This will allow the build to pass.

off_chain_data/application-go/store.go
```diff
@@ -0,0 +1,93 @@
...
+func simulateFailureIfRequired() error {
+	if simulatedFailureCount > 0 && transactionCount >= simulatedFailureCount {
+		transactionCount = 0
+		return errors.New("expected error: simulated write failure")
+	}
+
+	transactionCount += 1
+
+	return nil
+}
```
bestbeforetoday, line 83:
```suggestion
fvar errExpected = errors.New("expected error: simulated write failure")

func simulateFailureIfRequired() error {
	if simulatedFailureCount > 0 && transactionCount >= simulatedFailureCount {
		transactionCount = 0
		return errExpected
	}

	transactionCount += 1
	return nil
}
```

off_chain_data/application-go/listen.go
```diff
@@ -0,0 +1,270 @@
...
+	validTransactions, err := b.validTransactions()
+	if err != nil {
+		return err
+	}
+
+	for _, validTransaction := range validTransactions {
+		txProcessor := transactionProcessor{
+			b.parsedBlock.Number(),
+			validTransaction,
+			// TODO use reference to parent and get blockNumber, store and channelName from parent
```
bestbeforetoday, line 99: The transactions has the channel name in its channel header. I personally would not pass the blockProcessor to the transactionProcessor. It tightly couples the two implementations. I prefer to just pass the information that the transactionProcessor needs, which should be the block number, transaction and a store function (or store interface if you prefer to use a struct that maintains some state around the storing of writes, like the write and simulated failure count).

off_chain_data/application-go/listen.go
```diff
@@ -0,0 +1,270 @@
...
+		Writes:        writes,
+	}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (t *transactionProcessor) writes() ([]write, error) {
+	// TODO this entire code should live in the parser and just return the kvWrite which
```
bestbeforetoday, line 221: I think what you have here is fine. The Fabric protobuf structure groups the writes under a namespace, and then we can easily filter out all the read/write sets for system chaincodes by namespace., and then just get the writes since the reads don't modify ledger state. The parser needs to make all of that information available though. Other applications might need it and we want the parser to be reusable.

The Go version is much more verbose than the TypeScript version, which makes it look more complex. You might prefer to break out some of the blocks of logic within this function into other functions so the logic is easier to follow. Another tool to consider now that fabric-samples is using Go 1.23 is the new functions that have appeared in the standard **slices** package, like [slices.DeleteFunc](https://pkg.go.dev/slices@go1.23.0#DeleteFunc).

off_chain_data/application-go/app.go
twoGiants: Looks good. Done.

off_chain_data/application-go/store.go
twoGiants: Done.

off_chain_data/application-go/listen.go
twoGiants: Yes, you're right it tightly couples it. I'll leave it as it is.

I removed the channel name member and the to-do.

I created a `store` interface and implemented it with an `offChainStore` struct.

off_chain_data/application-go/listen.go
twoGiants: Sure, lets keep it then. I broke out some blocks of logic and used `slices.DeleteFunc` which is really nice, thanks for the advice!

bestbeforetoday:
The build failures look to be because of Docker Hub rate limiting. I have made a change that should avoid this in the future. If you rebase your change on the latest main branch content, hopefully we will see a clean build.

twoGiants:
> The build failures look to be because of Docker Hub rate limiting. I have made a change that should avoid this in the future. If you rebase your change on the latest main branch content, hopefully we will see a clean build.

[One](https://github.com/hyperledger/fabric-samples/actions/runs/13497782933/job/37710636330?pr=1269) is still failing.

off_chain_data/application-go/parser/endorserTransaction.go
```diff
@@ -0,0 +1,115 @@
...
+		if err := proto.Unmarshal(transactionAction.GetPayload(), chaincodeActionPayload); err != nil {
+			return nil, err
+		}
+
+		result = append(result, chaincodeActionPayload)
+	}
+	return result, nil
+}
+
+func (*endorserTransaction) extractChaincodeEndorsedActionsFrom(chaincodeActionPayloads []*peer.ChaincodeActionPayload) ([]*peer.ChaincodeEndorsedAction, error) {
```
bestbeforetoday, line 64: An error is never returned so you don't need the error return value. The results are just passed to `unmarshalProposalResponsePayloadsFrom` so you might find it easier to include this logic in that method. `chaincodeActionPayload.GetAction().GetProposalResponsePayload()`

off_chain_data/application-go/parser/endorserTransaction.go
```diff
@@ -0,0 +1,115 @@
...
+	result := []*peer.ChaincodeEndorsedAction{}
```
bestbeforetoday, line 65: This is really a nitpick... I think the result slice you create here (or rather its underlying array) is going to get discarded as soon as you append to it in the loop since there isn't capacity to append more elements. You can save the work of creating storage that will be discarded by using a nil value, which still works as an empty slice:

```go
var result []*peer.ChaincodeEndorsedAction
```

Or you can make a slice with capacity to hold `len(chaincodeActionPayloads)` elements up-front.

Same comment for the other instances of creating an empty (zero capacity) slice that you will then append to in this file.

Like I say, it's a really minor detail. Change it only if you want to.

off_chain_data/application-go/store.go
```diff
@@ -0,0 +1,93 @@
...
+
+	ocs.clearLastWrites()
+
+	if err := ocs.marshal(data.Writes); err != nil {
+		return err
+	}
+
+	if err := ocs.persist(); err != nil {
+		return err
+	}
```
bestbeforetoday, line 41: Using object state that you then need to manage instead of just passing arguments between private methods looks like a red flag to me. There is much more potential for things to go wrong since you are mutating more state in your object, and state is left lying around longer than you need it. For example, `ocs.writes` still contains data after you have finished persisting it and this method returns.

Also look out for error handling where you return only the error if it is non-nil; otherwise nil. Just return the error value directly. It is really easy to do this by accident.

Consider a simpler pattern, something like:

```go
	writes, err := ocs.marshal(data.Writes)
	if err != nil {
		return err
	}

	return ocs.persist(writes)
```

off_chain_data/application-go/store.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		return writeErr
+	}
+
+	if err := f.Close(); err != nil {
+		return err
+	}
+
+	return nil
```
bestbeforetoday, line 92:
```suggestion
	return f.Close()
```

off_chain_data/application-go/transact.go
```diff
@@ -0,0 +1,142 @@
...
+		}()
+	}
+
+	wg.Wait()
+
+	if err := context.Cause(ctx); err != nil {
+		return err
+	}
+
+	return nil
```
bestbeforetoday, line 78:
```suggestion
	return context.Cause(ctx)
```

bestbeforetoday (commented):
This is looking really good now. Just a few very small niggles that would be nice to tidy up and I think it is ready to go.

off_chain_data/application-go/parser/block.go
twoGiants: > ...I can see an argument for keeping it as a Block method to aid readability.
I keep it then. Thanks!

off_chain_data/application-go/parser/endorserTransaction.go
twoGiants: This is great, done.

off_chain_data/application-go/store.go
twoGiants: > For example, `ocs.writes` still contains data after you have finished persisting it and this method returns.

Actually, this was intended. But I was unsure about it. I thought it can make sense to keep the last write.

I believe I misunderstood this message then:

> a store function (or store interface if you prefer to use a struct that **maintains** some state around the storing of writes, like **the write** and simulated failure count).

I removed it and used the simpler pattern. Thanks for pointing out! I am remembering more and more of the good practices thank to your reviews.

off_chain_data/application-go/store.go
twoGiants: Done.

off_chain_data/application-go/transact.go
```diff
@@ -0,0 +1,142 @@
...
+
+	contract := gateway.GetNetwork(channelName).GetContract(chaincodeName)
+
+	smartContract := atb.NewAssetTransferBasic(contract)
+	app := newTransactApp(smartContract)
+	if err := app.run(); err != nil {
+		return err
+	}
+
+	return nil
```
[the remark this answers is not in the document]
twoGiants: Done.

off_chain_data/application-go/transact.go
twoGiants: Done.

twoGiants:
Two things are left on my mind:

1. `connect.go` panics (like in the other samples) instead of returning errors like everywhere else in this sample. Keep or change?
2. `block_test.go` tests are rudimentary. They helped me building the parser but they are incomplete. Keep, remove or extend?

Once you approve I will squash everything to one commit.

PR #1272, opened 2024-11-28
Fix build with new microfab release

bestbeforetoday:
The new microfab release seems to start slower than the previous release. This might be due to a change from solo to raft consensus. This causes failures in the full-stack-asset-transfer appdev tests. Chaincode deployment fails since the channel is not yet available.

Instead of increasing the sleep time after launching microfab, this change attempts to wait however long is required by looking for the "Microfab started" message in the container log before proceeding.

The test script is also updated to correctly stop the external chaincode process when the script exits. Previously the process ID of the npm command used to lauch the chaincode was captured rather than the actual chaincode process.

### DIFF
--- a/full-stack-asset-transfer-guide/justfile
+++ b/full-stack-asset-transfer-guide/justfile
@@ -168,15 +168,16 @@ microfab: microfab-down
             }
 
         ],
-        "capability_level":"V2_0"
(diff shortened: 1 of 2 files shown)


denyeart (approved):
Good find!

PR #1286, opened 2024-12-30
Generate crypto using CA for BFT consensus fix

cclchd-soundaryaayyappan:
**Bug Description:**
Encountered an issue in fabric-samples version 3.0 where deployment of ordering org with BFT consensus is providing error at generating artifacts step using CA.

**Steps to Reproduce:**
Run generate_artifacts.sh script with arguments included like ./generate_artifacts -o BFT -ca

**Expected Behavior:**
It should generate crypto-config and channel-artifacts required for the deployment.

**Environment:**
fabric-samples version: 3.0
Hyperledger Fabric version: 3.0
OS: Ubuntu 22.04

**Changes done to fix it:**
Make below changes,
1) ca/createEnrollments.sh -> Updated line 79 from

registerAndEnroll "5052" "orderer4" "orderer4pw" "orderer" "" "${orderer4_dir}" "${orderer_org}" "${orderer_org_tls}"
to
registerAndEnroll "5052" "orderer4" "orderer4pw" "orderer" "" "${orderer4_dir}" "${orderer_org_dir}" "${orderer_org_tls}"

2) Updated below Identity field in all 4 orderers in bft-config/configtx.yaml 

from

Identity: ../crypto-config/ordererOrganizations/example.com/orderers/orderer.example.com/msp/signcerts/orderer.example.com-cert.pem
to
Identity: ../crypto-config/ordererOrganizations/example.com/orderers/orderer.example.com/msp/signcerts/cert.pem

### DIFF
--- a/test-network-nano-bash/bft-config/configtx.yaml
+++ b/test-network-nano-bash/bft-config/configtx.yaml
@@ -287,28 +287,28 @@ Profiles:
           Host: 127.0.0.1
           Port: 6050
           MSPID: OrdererMSP
-          Identity: ../crypto-config/ordererOrganizations/example.com/orderers/orderer.example.com/msp/signcerts/orderer.example.com-cert.pem
+          Identity: ../crypto-config/ordererOrganizations/example.com/orderers/orderer.example.com/msp/signcerts/cert.pem
           ClientTLSCert: ../crypto-config/ordererOrganizations/example.com/orderers/orderer.example.com/tls/server.crt
           ServerTLSCert: ../crypto-config/ordererOrganizations/example.com/orderers/orderer.example.com/tls/server.crt
         - ID: 2
           Host: 127.0.0.1
           Port: 6051
           MSPID: OrdererMSP
-          Identity: ../crypto-config/ordererOrganizations/example.com/orderers/orderer2.example.com/msp/signcerts/orderer2.example.com-cert.pem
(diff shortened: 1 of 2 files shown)


test-network-nano-bash/bft-config/configtx.yaml
```diff
@@ -287,28 +287,28 @@ Profiles:
           Host: 127.0.0.1
           Port: 6050
           MSPID: OrdererMSP
-          Identity: ../crypto-config/ordererOrganizations/example.com/orderers/orderer.example.com/msp/signcerts/orderer.example.com-cert.pem
+          Identity: ../crypto-config/ordererOrganizations/example.com/orderers/orderer.example.com/msp/signcerts/cert.pem
```
denyeart, line 290: I think this will break the default scenario where cryptogen is used.
Maybe when the certs are created using CA in createEnrollments.sh we need to copy the pem file to the existing name produced by cryptogen.

cclchd-soundaryaayyappan: @denyeart  Thanks for the review and comment. You're right that these changes will affect the default scenario where the cryptogen is used.

But the changes I mentioned in the files is completely related to the orderer service using Fabric CA. When I try to deploy ordering service with BFT consensus, generation of crypto is happening only if we made these changes, otherwise unable to deploy such a configuration of ordering nodes.

bestbeforetoday:
Support for running the test-network using CA and BFT consensus was added by pull request #1301 so I will close this one. Thank you for taking the time to develop and submit this change.

PR #1290, opened 2025-01-10
Fix issue 1289 with missing parameter in peer chaincode lifecycle com…

munapower:
Added the orderer endpoint as a variable and added the missing flag in the Exercise instructions.


denyeart (approved):
Thank you!

PR #1291, opened 2025-01-14
Update test-network-k8s to support both v2.5 and v3.0

satota2:
The existing orderer configuration file is incompatible with v3.0, and the v3.0 configuration does not work with v2.5.
To support both versions, configuration settings have been updated to use environment variables instead of referencing static configuration files.


bestbeforetoday (approved):
Great fix, thank you!

PR #1294, opened 2025-01-21
test-network-k8s: Introduce initial support for BFT orderers

satota2:
This PR adds initial support for BFT orderers in the test-network-k8s.
When `TEST_NETWORK_ORDERER_TYPE` is set to `bft`, the network launches four orderers configured with SmartBFT.

I plan to update the documentation after this PR is merged.

### How to try

```bash
cd test-network-k8s

export TEST_NETWORK_FABRIC_VERSION=3.0
export TEST_NETWORK_ORDERER_TYPE=bft

./network kind
./network cluster init
./network up
./network channel create

./network chaincode deploy asset-transfer-basic ../asset-transfer-basic/chaincode-java
./network chaincode invoke asset-transfer-basic '{"Args":["InitLedger"]}'
./network chaincode query  asset-transfer-basic '{"Args":["ReadAsset","asset1"]}'
```

PR #1296, opened 2025-01-29
CI: Add BFT Orderer support

satota2:
### Description
This patch adds BFT Orderer testing to the CI workflows. The following test environments are updated:
- test-network
- test-network-k8s

### Supplementary Information
Currently, the fabric-samples CI is primarily focused on the latest LTS version, Fabric v2.5.x. However, BFT Orderer support is available only in Fabric v3.0 or later.
To address this, the BFT Orderer tests explicitly specify Fabric v3.0, while other tests continue to use the default version (v2.5.x).
I believe this approach ensures that both the latest LTS version (v2.5) and v3.0 tests are supported with minimal changes.

### Related PRs
- https://github.com/hyperledger/fabric-samples/pull/1294

### DIFF
--- a/.github/workflows/test-network-bft-orderer.yaml
+++ b/.github/workflows/test-network-bft-orderer.yaml
@@ -1,0 +1,45 @@
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+name: Test Network BFT Orderer 🍟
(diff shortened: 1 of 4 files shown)


bestbeforetoday (approved):
Perfect, thank you!

PR #1303, opened 2025-02-18
Add connection profile to test-network-nano-bash

denyeart:
Add connection profile to test-network-nano-bash
to enable legacy SDK apps.

Approach copied from test-network connection profiles.


satota2 (approved):
@denyeart LGTM!

PR #1304, opened 2025-02-20
Fix intermittent errors with test-network ca

denyeart:
If CA server hasn't finished initialization then the initial register and enroll requests will fail.
Waiting 3 seconds resolves the issue.


bestbeforetoday (approved):
It would be nice if there was a better way to poll for the CA being functional instead of just having an arbitrary wait. It's better than having intermittent failures though.

PR #1305, opened 2025-02-20
Fix intermittent errors with test-network ca

denyeart:
If CA server hasn't finished initialization then the initial register and enroll requests will fail.
Instead of waiting 3 seconds, actually check if CA service is ready.


denyeart:
@bestbeforetoday I agree with your feedback and addressed...

bestbeforetoday (approved):
Nice improvement!

PR #1307, opened 2025-02-23
Pull Fabric images from GitHub Container Registry

bestbeforetoday:
By default, Docker images are referenced from Docker Hub. This has now applied rate limiting that is causing builds to fail. Current Fabric v2.5 and chaincode container Docker images are also published to GitHub Container Registry, which does not impose rate limits and will be located closer to the GitHub Actions runners.

This change:

- Updates to latest Fabric component versions, which are available in GHCR. The Fabric install script defaults to GHCR as the Docker registry for versions that are available here.
- Pulls and re-tags chaincode container images from GHCR in advance of running tests in the build. This avoids them being pulled from Docker Hub during test execution.

### DIFF
--- a/.github/actions/fsat-setup/action.yaml
+++ b/.github/actions/fsat-setup/action.yaml
@@ -6,16 +6,16 @@ inputs:
     default: 18.x
   just-version:
     description: Just Version
-    default: '1.24.0'
+    default: "1.24.0"
   k9s-version:
     description: k9s Version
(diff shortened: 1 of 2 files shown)


denyeart (approved):
Thanks Mark, I believe this completes the CI shift to ghcr across all the important repositories.  I'll close out the overall tracking issue https://github.com/hyperledger/fabric/issues/4998

PR #1309, opened 2025-02-27
Implement TypeScript version of asset-transfer-ledger-queries

### LINKED ISSUE #1232: Add TypeScript version of asset-transfer-ledger-queries chaincode
## Descriptions

Ideally, various languages would have the same support, but currently, the TypeScript version of the asset-transfer-ledger-queries chaincode is not supported.

So, implementing a TypeScript version of the asset-transfer-ledger-queries chaincode should be considered.
This sample would include ledger query functionalities like GetHistoryForKey, making it more valuable for TypeScript users.

## Related Dissucussions

https://github.com/hyperledger/fabric-samples/pull/1226

satoshi-ito-ht:
This commit implements the TypeScript version of the asset-transfer-ledger-queries Chaincode.

Resolves #1232

### DIFF
--- a/asset-transfer-ledger-queries/chaincode-typescript/src/assetTransferLedger.ts
+++ b/asset-transfer-ledger-queries/chaincode-typescript/src/assetTransferLedger.ts
@@ -1,0 +1,482 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// ====CHAINCODE EXECUTION SAMPLES (CLI) ==================
+
+// ==== Invoke assets ====
+// peer chaincode invoke -C CHANNEL_NAME -n asset_transfer -c '{"Args":["CreateAsset","asset1","blue","35","Tom","100"]}'
+// peer chaincode invoke -C CHANNEL_NAME -n asset_transfer -c '{"Args":["CreateAsset","asset2","red","50","Tom","150"]}'
+// peer chaincode invoke -C CHANNEL_NAME -n asset_transfer -c '{"Args":["CreateAsset","asset3","blue","70","Tom","200"]}'
+// peer chaincode invoke -C CHANNEL_NAME -n asset_transfer -c '{"Args":["TransferAsset","asset2","jerry"]}'
+// peer chaincode invoke -C CHANNEL_NAME -n asset_transfer -c '{"Args":["TransferAssetByColor","blue","jerry"]}'
+// peer chaincode invoke -C CHANNEL_NAME -n asset_transfer -c '{"Args":["DeleteAsset","asset1"]}'
+
+// ==== Query assets ====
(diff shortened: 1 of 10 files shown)


bestbeforetoday (changes requested):
This looks great. Some small changes I think are needed before merging:

1. Remove `tslint.json`. The sample uses eslint (with typescript-eslint), not the deprecated tslint. I don't think think this tslint configuration file is used.
2. Include an `npm-shrinkwrap.json` file to minimise the chances of future breakages when dependencies release new versions by pinning the resolved dependency versions. Just run `npm install && npm shrinkwrap` locally, then check in the created shrinkwrap file.
3. Add a `.gitignore` in the `asset-transfer-ledger-queries/chaincode-typescript` directory to ignore the `dist` subdirectory generated by the TypeScript compilation.
4. Add `typescript` to the chaincode-language matrix in `.github/workflows/test-network-ledger.yaml` so this chaincode is tested by the automated build.

asset-transfer-ledger-queries/chaincode-typescript/src/assetTransferLedger.ts
```diff
@@ -0,0 +1,474 @@
...
+        let result = await iterator.next();
+
+        while (!result.done) {
+            const historyRecord = result.value;
+            let asset: Asset | null = null;
+            if (historyRecord.value.length > 0) {
+                asset = JSON.parse(historyRecord.value.toString()) as Asset;
+            }
+            // Convert the timestamp to a Date object.
+            const seconds = historyRecord.timestamp.seconds;
```
satota2, line 326: Timestamp includes nanos as a field, so I think it is necessary to convert this to Date as well.

```typescript
interface Timestamp {
    seconds: Long;
    nanos: number;
}
```

satota2 (changes requested):
@satoshi-ito-ht
Thank you for your contribution.
I've also added some review comments, so please check them out.

bestbeforetoday (approved):

satota2 (approved):
@bestbeforetoday 
> What do you think, @satota2?

Thank you for sharing your thoughts.
I agree with your opinion that the behavior should be consistent across all languages.
It is a tough decision whether to remove the JavaScript implementation as you mentioned, or modify it to match the other languages. Removing it could reduce maintenance costs, but I’m not sure if the community and users feel that TypeScript alone is sufficient, making JavaScript unnecessary.

One idea is to create an issue or discussion about removing JavaScript, allowing the community to share their thoughts. If there are no strong objections within a certain period, we could proceed with the removal.
Does this approach sound good to you?

> Any removal can happen in a different pull request so I am happy with this change.

Yes, I also agree with this PR. I think we can merge it now and continue discussing removal or unification in a separate issue or PR.

Thank you, @satoshi-ito-ht, for your updates!

PR #1310, opened 2025-03-05
NixOS support (Multi-Linux Distro Support)

AadithyanRaju:
using bash scriptname helps in deployment in nixos 
cause output of (which bash) is /run/current-system/sw/bin/bash instead of /bin/bash

### DIFF
--- a/ci/scripts/run-k8s-test-network-basic.sh
+++ b/ci/scripts/run-k8s-test-network-basic.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash -e
+#!/usr/bin/env bash
 #
 # Copyright IBM Corp All Rights Reserved
 #


bestbeforetoday:
Thank you for taking the time to address the issue you see running in environments where bash is not located at /bin/bash. Does it work on your system to change the shebang at the top of all the test-network script from:

```shell
#!/bin/bash
```

to:

```shell
#!/usr/bin/env bash
```

That seems to me a cleaner (or at least more standard) solution.

AadithyanRaju:
Yes it does, i will make the changes.

test-network/scripts/utils.sh
```diff
@@ -210,6 +210,7 @@ function installPrereqs() {
   FILE=../install-fabric.sh     
   if [ ! -f $FILE ]; then
     curl -sSLO https://raw.githubusercontent.com/hyperledger/fabric/main/scripts/install-fabric.sh && chmod +x install-fabric.sh
+    sed -i '1s|.*|#!/usr/bin/env bash|' install-fabric.sh
```
AadithyanRaju, line 213: I added this line to change the shebang of install-fabric after it get downloaded.

bestbeforetoday: In-place sed commands often fail on other platforms, like MacOS, so I would be very careful with them. It is proibably better to submit a pull request to the core Fabric repository to update the [install script](https://github.com/hyperledger/fabric/blob/main/scripts/install-fabric.sh) directly.

AadithyanRaju: ok

bestbeforetoday (approved):
Thank you for the contribution!

bestbeforetoday (changes requested):
There is a problem with the scripts that pass options to bash in the shebang. From one of the failed build steps:

```
/usr/bin/env: ‘bash -e’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines
```

This will need to be fixed.

bestbeforetoday (approved):
Thank you again for making this change.

PR #1313, opened 2025-03-07
ERC20 realization update for javascript

XZSt4nce:
- JSDoc returns Promises in async functions
- Using DRY for ClientAccountID, BalanceOf etc
- Added function ClientAccountMSPID
- Added function CheckAuthorization
- Using new function _update for minting, burning and transfering tokens
- Additional functions for approving, spending allowance and checking empty values
- Updated and repaired tests

### DIFF
--- a/token-erc-20/chaincode-javascript/lib/tokenERC20.js
+++ b/token-erc-20/chaincode-javascript/lib/tokenERC20.js
@@ -27,7 +27,7 @@ class TokenERC20Contract extends Contract {
      * As a work around, we use `TokenName` as an alternative function name.
      *
      * @param {Context} ctx the transaction context
-     * @returns {String} Returns the name of the token
+     * @returns {Promise<String>} Returns the name of the token
     */
     async TokenName(ctx) {
 
@@ -43,7 +43,7 @@ class TokenERC20Contract extends Contract {
      * Return the symbol of the token. E.g. “HIX”.
      *
      * @param {Context} ctx the transaction context
-     * @returns {String} Returns the symbol of the token
+     * @returns {Promise<String>} Returns the symbol of the token
     */
     async Symbol(ctx) {
 
@@ -51,6 +51,7 @@ class TokenERC20Contract extends Contract {
         await this.CheckInitialized(ctx);
 
         const symbolBytes = await ctx.stub.getState(symbolKey);
+
         return symbolBytes.toString();
     }
 
@@ -59,7 +60,7 @@ class TokenERC20Contract extends Contract {
      * e.g. 8, means to divide the token amount by 100000000 to get its user representation.
      *
      * @param {Context} ctx the transaction context
-     * @returns {Number} Returns the number of decimals
+     * @returns {Promise<Number>} Returns the number of decimals
     */
     async Decimals(ctx) {
 
@@ -68,6 +69,7 @@ class TokenERC20Contract extends Contract {
 
         const decimalsBytes = await ctx.stub.getState(decimalsKey);
         const decimals = parseInt(decimalsBytes.toString());
+
         return decimals;
     }
 
@@ -75,7 +77,7 @@ class TokenERC20Contract extends Contract {
      * Return the total token supply.
      *
      * @param {Context} ctx the transaction context
-     * @returns {Number} Returns the total token supply
+     * @returns {Promise<Number>} Returns the total token supply
     */
     async TotalSupply(ctx) {
 
@@ -84,6 +86,7 @@ class TokenERC20Contract extends Contract {
 
         const totalSupplyBytes = await ctx.stub.getState(totalSupplyKey);
         const totalSupply = parseInt(totalSupplyBytes.toString());
+
         return totalSupply;
     }
 
@@ -92,7 +95,7 @@ class TokenERC20Contract extends Contract {
      *
      * @param {Context} ctx the transaction context
      * @param {String} owner The owner from which the balance will be retrieved
-     * @returns {Number} Returns the account balance
+     * @returns {Promise<Number>} Returns the account balance
      */
     async BalanceOf(ctx, owner) {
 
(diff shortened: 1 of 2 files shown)


token-erc-20/chaincode-javascript/lib/tokenERC20.js
```diff
@@ -331,154 +248,232 @@ class TokenERC20Contract extends Contract {
...
+    }
+
+    /**
+     * ClientAccountMSPID returns the MSP id of the requesting client's account.
+     * In this implementation, the client account MSP ID is the clientMspId itself.
+     *
+     * @param {Context} ctx the transaction context
+     * @returns {String} Returns the account MSP id
+     */
+    ClientAccountMSPID(ctx) {
```
bestbeforetoday: There is only one call to this function within the smart contract. Perhaps that one call should just be replaced with `ctx.clientIdentity.getMSPID()` and this function removed. The capitalised function name suggests it is intended to be evaluated by clients but this seems redundant since the client must already know its MSP ID in order to interact with the smart contract.

token-erc-20/chaincode-javascript/lib/tokenERC20.js
```diff
@@ -331,154 +248,232 @@ class TokenERC20Contract extends Contract {
...
-        const totalSupplyBytes = await ctx.stub.getState(totalSupplyKey);
-        if (!totalSupplyBytes || totalSupplyBytes.length === 0) {
-            throw new Error('totalSupply does not exist.');
+        if (valueInt < 0) { // transfer of 0 is allowed in ERC20, so just validate against negative amounts
+            throw new Error('transfer amount cannot be negative');
         }
-        const totalSupply = this.sub(parseInt(totalSupplyBytes.toString()), amountInt);
+
+        let totalSupply = await this.TotalSupply(ctx);
+        if (this.isEmpty(from)) {
```
bestbeforetoday:
```suggestion
        if (!from) {
```

token-erc-20/chaincode-javascript/lib/tokenERC20.js
```diff
@@ -331,154 +248,232 @@ class TokenERC20Contract extends Contract {
...
-        await this.CheckInitialized(ctx);
+    /**
+     * @param {Context} ctx the transaction context
+     * @param {String} owner The owner of tokens
+     * @param {String} spender The spender
+     * @param {Number} value The amount of token to be transferred
+     * @param {Boolean} emitEvent Whether to emit event
+     */
+    async _approveEvent(ctx, owner, spender, value, emitEvent) {
+        if (this.isEmpty(owner)) {
```
bestbeforetoday:
```suggestion
        if (!owner) {
```

token-erc-20/chaincode-javascript/lib/tokenERC20.js
```diff
@@ -331,154 +248,232 @@ class TokenERC20Contract extends Contract {
...
+            throw new Error('invalid approver');
+        }
+        if (this.isEmpty(spender)) {
```
bestbeforetoday:
```suggestion
        if (!spender) {
```

token-erc-20/chaincode-javascript/lib/tokenERC20.js
```diff
@@ -331,154 +248,232 @@ class TokenERC20Contract extends Contract {
...
+     * Transfers a `value` amount of tokens from `from` to `to`, or alternatively mints (or burns) if `from`
+     * (or `to`) is the zero address. All customizations to transfers, mints, and burns should be done by overriding
+     * this function.
+     *
+     * @param {Context} ctx the transaction context
+     * @param {String} from The sender
+     * @param {String} to The recipient
+     * @param {Number} value The amount of token to be transferred
+     */
+    async _update(ctx, from, to, value) {
```
bestbeforetoday: I appreciate that you are trying to reduce code duplication across transaction functions but I am concerned that the logic of this `_update` implementation introduces a lot of complexity. Some flow explicitly require `from` and `to` values while others rely on those values being falsy. This means that this function has four potential modes of invocation:

1. `from` and `to` both set.
2. `from` set and `to` unset.
3. `from` unset and `to` set.
4. `from` and `to` both unset.

This creates high cyclomatic complexity. I think it would be preferable if each (valid) combination was implemented separately with each combination having only the parameters actually needed to make the intent very clear, despite some similarities between them. Alternatively you could still avoid duplication by splitting specific parts of this update flow into separate functions and having transfer, mint and burn compose those parts in the appropriate combination.

token-erc-20/chaincode-javascript/test/tokenERC20.test.js
```diff
@@ -119,10 +108,9 @@ describe('Chaincode', () => {
             mockStub.createCompositeKey.withArgs('balance', ['Bob']).returns('balance_Bob');
             mockStub.getState.withArgs('balance_Bob').resolves(null);
 
-            const response = await token._transfer(ctx, 'Alice', 'Bob', '1000');
-            sinon.assert.calledWith(mockStub.putState.getCall(0), 'balance_Alice', Buffer.from('0'));
-            sinon.assert.calledWith(mockStub.putState.getCall(1), 'balance_Bob', Buffer.from('1000'));
-            expect(response).to.equals(true);
+            await token._transfer(ctx, 'Alice', 'Bob', '1000');
+            sinon.assert.calledWith(mockStub.putState.getCall(4), 'balance_Alice', Buffer.from('0'));
+            sinon.assert.calledWith(mockStub.putState.getCall(5), 'balance_Bob', Buffer.from('1000'));
```
bestbeforetoday: The need to change the call number highlights the fragility of this testing approach. The call number should not matter. I suggest this instead, which checks that a call was made with the required arguments regardless of the call ordering:

```suggestion
            sinon.assert.calledWith(mockStub.putState, 'balance_Alice', Buffer.from('0'));
            sinon.assert.calledWith(mockStub.putState, 'balance_Bob', Buffer.from('1000'));
```

Same suggestion for all the other changes in this file that required a different call number.

PR #1320, opened 2025-03-21
update fabric operations console docs

Thomas-Leung:
Update docs for Mac users when deploying fabric operations console on local machines.


denyeart (approved):
Good guidance!

PR #1324, opened 2025-04-14
Improve logging for test-network-nano-bash CA

denyeart:
Make the CA script logging more readable for users trying to understand the fabric-ca-client commands.

Print the actual command to the log and remove the debug logging which adds noise.


satota2 (approved):
@denyeart Thanks for the improvement. LGTM!

Issue #1273, opened 2024-11-29
Failed to install tokengen for using in fabric-token-sdk


MohammadrezaNasrabadi:
I want to set up the hyperlerdger with token network, as provided in token-sdk fabric-samples. One step is installing the tokengen command, However, it seems like there is some dependencies problem on pkcs11 module.

After issuing the below command:

```bash
go install github.com/hyperledger-labs/fabric-token-sdk/cmd/tokengen@v0.3.0
```

The below error output is shown:
```
# github.com/hyperledger/fabric/bccsp/pkcs11
/root/go/pkg/mod/github.com/hyperledger/fabric@v1.4.0-rc1.0.20230401164317-bd8e24856939/bccsp/pkcs11/pkcs11.go:41:21: undefined: pkcs11.Ctx
/root/go/pkg/mod/github.com/hyperledger/fabric@v1.4.0-rc1.0.20230401164317-bd8e24856939/bccsp/pkcs11/pkcs11.go:51:23: undefined: pkcs11.SessionHandle
/root/go/pkg/mod/github.com/hyperledger/fabric@v1.4.0-rc1.0.20230401164317-bd8e24856939/bccsp/pkcs11/pkcs11.go:52:22: undefined: pkcs11.SessionHandle
/root/go/pkg/mod/github.com/hyperledger/fabric@v1.4.0-rc1.0.20230401164317-bd8e24856939/bccsp/pkcs11/pkcs11.go:55:32: undefined: pkcs11.ObjectHandle
/root/go/pkg/mod/github.com/hyperledger/fabric@v1.4.0-rc1.0.20230401164317-bd8e24856939/bccsp/pkcs11/pkcs11.go:326:51: undefined: pkcs11.SessionHandle
/root/go/pkg/mod/github.com/hyperledger/fabric@v1.4.0-rc1.0.20230401164317-bd8e24856939/bccsp/pkcs11/pkcs11.go:338:46: undefined: pkcs11.SessionHandle
/root/go/pkg/mod/github.com/hyperledger/fabric@v1.4.0-rc1.0.20230401164317-bd8e24856939/bccsp/pkcs11/pkcs11.go:370:50: undefined: pkcs11.SessionHandle
/root/go/pkg/mod/github.com/hyperledger/fabric@v1.4.0-rc1.0.20230401164317-bd8e24856939/bccsp/pkcs11/pkcs11.go:385:51: undefined: pkcs11.SessionHandle
[...]

PR #1325, opened 2025-04-14
enable CGO on installation of tokensdk

MohammadrezaNasrabadi:
According to the issue https://github.com/hyperledger/fabric-samples/issues/1273 , installation of tokensdk might be not successful. After research, I've found the CGO_ENABLED variable should be set to 1.

In addition, after merging the changes of #1315 there were some dependency errors when building the tokensdk images.

Signed-off-by: Mohammad Reza Nasrabadi <nasr.mareza1378@gmail.com

### DIFF
--- a/token-sdk/README.md
+++ b/token-sdk/README.md
@@ -82,6 +82,7 @@ Prerequisites:
 
 - bash
 - golang 1.20+
+- build-essential
 - git
 - docker
 - docker-compose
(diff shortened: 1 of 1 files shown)


MohammadrezaNasrabadi:
> > According to the issue #1273 , installation of tokensdk might be not successful. After research, I've found the CGO_ENABLED variable should be set to 1.
> > In addition, after merging the changes of #1315 there were some dependency errors when building the tokensdk images.
> 
> could you split those two updates into separate PRs? first of all, they are not directly related.
> 
> then, I wrote the first part, as suggested in #1273; this might be some dependency issue that needs to be localized and understood; switching to static compilation only shadows the issue and does not really solve it.

Allright. The changes related to tokensdk applications dependencies were transported to PR #1326 . I can spend time to work on solving the `pkcs11` dependencies and will update on this PR.

bestbeforetoday:
@adecaro @KElkhiyaoui @arner @alexandrosfilios As fabric-token-sdk maintainers, please could one of you confirm that these changes are good?

bestbeforetoday (changes requested):
I was able to run successfully without explicitly enabling CGO. See this [issue comment](https://github.com/hyperledger/fabric-samples/issues/1273#issuecomment-3127232690) for details. I don't think the right answer is to document that CGO is enabled in all cases. Different platforms / environments seem to have different default values for CGO_ENABLED; perhaps for good reason.

PR #1336, opened 2025-07-24
Update fabric-gateway to v1.8.0

bestbeforetoday:


satota2 (commented):
@bestbeforetoday 

Thank you for the update!
Just a quick clarification: I see `go 1.23.0` and `toolchain go1.24.5` specified in `off_chain_data/application-go/go.mod`.
Is there a specific reason for requiring Go 1.24.5 as the toolchain, instead of aligning it with 1.23.0?
I understand toolchain is useful for build reproducibility, but wondering if it might be optional or if there's a known compatibility issue with 1.23.x.

PR #1349, opened 2025-09-08
Use Gradle wrapper in asset-transfer-ledger-queries

bestbeforetoday:
The CI script invoked Gradle directly, which could picker newer Gradle versions not compatible with the build definition. This change uses the Gradle wrapper included in the client application sample.


satota2 (approved):
@bestbeforetoday LGTM. Thank you for the update.

PR #1357, opened 2025-10-27
Fix private-data TypeScript chaincode

bestbeforetoday:
The build is only testing the Go chaincode for the asset-transfer-private-data sample. The behavior of the TypeScript chaincode implementation is not consistent with the Go and Java versions, which prevents it from working correctly.

This change fixes the TypeScript chaincode implementation for the asset-transfer-private-data sample so that it works correctly. Additionally, some JSON property names are explicitly set in the Go client application sample, which prevented it from working with the Java and TypeScript chaincode implementations.

### DIFF
--- a/asset-transfer-private-data/chaincode-typescript/src/assetTransferTransientInput.ts
+++ b/asset-transfer-private-data/chaincode-typescript/src/assetTransferTransientInput.ts
@@ -2,7 +2,7 @@
   SPDX-License-Identifier: Apache-2.0
 */
 
-import { nonEmptyString, positiveNumber } from "./utils";
+import { nonEmptyString, positiveNumber } from './utils';
 
 export class TransientAssetProperties {
     objectType: string;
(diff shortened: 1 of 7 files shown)


asset-transfer-private-data/chaincode-typescript/src/assetTransferTransientInput.ts
```diff
@@ -12,104 +12,142 @@ export class TransientAssetProperties {
     appraisedValue: number;
 
     constructor(transientMap: Map<string, Uint8Array>) {
-        const transient = transientMap.get("asset_properties");
+        const transient = transientMap.get('asset_properties');
```
Tsafaras: The project could use a strict formatter, no? Like [biome](https://biomejs.dev/).

Tsafaras (commented):
I tested it and working as expected! Thanks for the quick fix.

Left some questions that I have.

denyeart (approved):

PR #1414, opened 2026-05-04
changed weak ordering implementation in sort.slice

### LINKED ISSUE #1413: ERC-1155 multi recipient transfer can produce non deterministic endorsements
It happens because sortedKeysToID() has an inconsistent comparator in contract.go, and its output is used to drive state writes. For keys sharing To but different ID, ordering is undefined.

It can be reproduced by calling BatchTransferFromMultiRecipient with repeated recipient and multiple token IDs; 
peers can iterate/write in different orders and return mismatched proposal payloads.

rootp1:
1. The original code compared the To (recipient) strings when the numeric IDs were different, and fell back to comparing IDs when they were the same. This was completely backwards and broke Go's strict weak ordering. We fixed it to accurately compare IDs first, and properly fallback to the To string on identical IDs.
2. successfully implemented a unit test checking identical IDs for different To strings (and vice-versa) to explicitly assert that the Go map randomness is predictably flattened to an identically ordered sequence on every peer simulation.
3. Since token-erc-1155 currently only contains chaincode-go, there are no Javascript/Java/Typescript counterparts requiring this identical fix in this repository.

Closes #1413

### DIFF
--- a/token-erc-1155/chaincode-go/chaincode/contract.go
+++ b/token-erc-1155/chaincode-go/chaincode/contract.go
@@ -1113,9 +1113,9 @@ func sortedKeysToID(m map[ToID]uint64) []ToID {
 	// Sort the slice first according to ID if equal then sort by recipient ("To" field)
 	sort.Slice(keys, func(i, j int) bool {
 		if keys[i].ID != keys[j].ID {
-			return keys[i].To < keys[j].To
+			return keys[i].ID < keys[j].ID
 		}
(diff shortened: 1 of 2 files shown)


bestbeforetoday (approved):
Thank you for this fix and the clear unit test to confirm the correct behaviour.

rootp1:
> Thank you for this fix and the clear unit test to confirm the correct behaviour.

Thankyou for the appreciation @bestbeforetoday

Issue #1410, opened 2026-05-01
Hitting docker hub rate limits using the k8s test network


jt-nti:
I worked around the problem by using an alternative container registry but that involved editing various scripts and config files. The fabric images were easier because `TEST_NETWORK_FABRIC_CONTAINER_REGISTRY` already provides a way to override the default container registry. I'm thinking of creating a PR to switch the default fabric container registry to `ghcr.io`, and to add a new `TEST_NETWORK_3RD_PARTY_CONTAINER_REGISTRY` env var to make it easier to use other registries.

Does that sound ok?

Docker hub is not the only container registry with rate limits though, so it would also be useful to update the sample to work with API keys, or maybe that's just a readme update?

bestbeforetoday:
Sounds like a good plan to me. Some related changes from docker.io to ghcr.io happened in hyperledger/fabric-samples#1307 and hyperledger/fabric#5146. The Fabric install script uses `FABRIC_DOCKER_REGISTRY` as an environment variable override to the registry portion of the Docker image name.

jt-nti:
Something like #1415 ?

PR #1416, opened 2026-05-19
test-network-k8s: ghcr.io for Hyperledger images

jt-nti:
Default to using ghcr.io for Hyperledger images in the k8s test network

See #1410


jt-nti:
I was having more difficulty getting the tests to pass with the other changes in #1415  so I've split out the ghcr.io change for Hyperledger images to start with.

bestbeforetoday (approved):

PR #1428, opened 2026-07-02
Group Dependabot updates and security fixes by ecosystem

ryjones:
Add grouped update entries for the gomod, npm, docker, maven, and gradle ecosystems (in addition to the existing github-actions entry). Each ecosystem uses a recursive directories glob so all sample manifests are covered, and defines separate groups for version updates and security updates so each ecosystem produces a single consolidated PR.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,87 @@ updates:
     directory: "/"
     schedule:
       interval: weekly
+    groups:
+      github-actions:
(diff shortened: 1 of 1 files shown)


bestbeforetoday (approved):
I suspect that there will be times when this causes some friction as specific updates cause issues for PRs batching large numbers of updates, but we can cross that bridge when we come to it. Happy to give this a go.

PR #641, opened 2022-02-11
Issue 602

kieranomahony98:
Flow:

Check if docker network is running. 
If network is not running exit
If running, and containers are present + certs are present skip network up
If running, containers are not present then we should networkUp
If containers exist but certs don’t, bring network down and then networkUp

When containers are running but for some reason peerOrganisation dir was not present, the following error: 
`Failed to generate channel configuration transaction...` may be awkward to debug. So this is why there is a check to see if peerOrg does not exist but the containers do. 


Test Cases: 

1) Docker is not running 
```
./network.sh up createChannel -ca
Using docker and docker-compose
Creating channel 'mychannel'.
If network is not up, starting nodes with CLI timeout of '5' tries and CLI delay of '3' seconds and using database 'leveldb 
docker network is required to be running to create a channel
```

2) Docker is running but containers are not, peerOrganizations directory exists
```
./network.sh up createChannel -ca
Using docker and docker-compose
Creating channel 'mychannel'.
If network is not up, starting nodes with CLI timeout of '5' tries and CLI delay of '3' seconds and using database 'leveldb 
Bringing up network
LOCAL_VERSION=2.4.1
DOCKER_IMAGE_VERSION=2.4.1
CA_LOCAL_VERSION=1.5.2
CA_DOCKER_IMAGE_VERSION=1.5.2
WARN[0000] Found orphan containers ([ca_org2 ca_org1 ca_orderer]) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up. 
[+] Running 4/4
 ⠿ Container peer0.org2.example.com  Started                                                                                                                                                                                                                               0.6s
 ⠿ Container peer0.org1.example.com  Started                                                                                                                                                                                                                               0.6s
 ⠿ Container orderer.example.com     Started                                                                                                                                                                                                                               0.6s
 ⠿ Container cli                     Started  
```
3) Containers running but  peerOrganizations directory does not exist
```
./network.sh up createChannel -ca
Using docker and docker-compose
Creating channel 'mychannel'.
If network is not up, starting nodes with CLI timeout of '5' tries and CLI delay of '3' seconds and using database 'leveldb with crypto from 'Certificate Authorities'
Decomposing compose-test-net.yaml
[+] Running 10/7
 ⠿ Container orderer.example.com          Removed                                                                                                                                                                                                                          0.2s
 ⠿ Container cli                          Removed                                                                                                                                                                                                                         10.2s
[...]

PR #1446, opened 2026-09-03
Fix createChannel wrongly recreating an already-running network

### LINKED ISSUE #1243: test network incorrectly recreates a fabric network when creating a channel
Doing the following
```
./network.sh up -s couchdb
./network.sh createChannel -c mychannel
```

results in the fabric network being recreated by the createChannel command using levelDB. The reason for this is due to this check in the `createChannel` Function of `network.sh`

```
 [[ $len -lt 4 ]] || [[ ! -d "organizations/peerOrganizations" ]] && bringUpNetwork="true" || echo "Network Running Already"
```

as no fabric-tools image now pulled and launched the check that there are less than 4 images running is not correct.

I do have to question as to why does the createChannel attempt to start a network if no network exists ? The paradim of doing a single action may be better here and just have the up command do it

But as of now it is wrong so a quick fix would be
```
 [[ $len -lt 3 ]] || [[ ! -d "organizations/peerOrganizations" ]] && bringUpNetwork="true" || echo "Network Running Already"
```

wakqasahmed:
Fixes #1243.

`createChannel()` decides whether the network is already up by counting running `hyperledger/`-prefixed containers and treating fewer than 4 as "not running". That threshold was correct back in 2022 (#641, when a persistent `cli` container was still part of the default network) but the default network today only launches 3 (orderer + 2 peers) - the `cli` container's gone. So `createChannel` always sees 3 < 4, decides the network isn't up, and calls `networkUp()` again - which doesn't inherit `-s couchdb` from the original `up` command, so it silently overwrites the running CouchDB-backed network with a fresh LevelDB one, exactly as reported.

I verified this against real containers rather than just reasoning from the compose files: reproduced with `./network.sh up -s couchdb` followed by a separate `./network.sh createChannel -c mychannel` - confirmed the peers get `Recreate`d, `couchdb0`/`couchdb1` end up orphaned, and the log switches to `using database leveldb`. With the threshold corrected to 3, the same steps now print `Network Running Already` and leave the running containers untouched.

Fixed both occurrences of the stale `4` - the count check that decides whether to bring the network down to resync certs (line 337) has the same assumption baked in and would stay broken if only the reported line were changed.

### DIFF
--- a/test-network/network.sh
+++ b/test-network/network.sh
@@ -334,12 +334,12 @@ function createChannel() {
   CONTAINERS=($($CONTAINER_CLI ps | grep hyperledger/ | awk '{print $2}'))
   len=$(echo ${#CONTAINERS[@]})
 
-  if [[ $len -ge 4 ]] && [[ ! -d "organizations/peerOrganizations" ]]; then
+  if [[ $len -ge 3 ]] && [[ ! -d "organizations/peerOrganizations" ]]; then
     echo "Bringing network down to sync certs with containers"
     networkDown
   fi
 
-  [[ $len -lt 4 ]] || [[ ! -d "organizations/peerOrganizations" ]] && bringUpNetwork="true" || echo "Network Running Already"
+  [[ $len -lt 3 ]] || [[ ! -d "organizations/peerOrganizations" ]] && bringUpNetwork="true" || echo "Network Running Already"
 
   if [ $bringUpNetwork == "true"  ]; then
     infoln "Bringing up network"


wakqasahmed:
Hi @bestbeforetoday — not sure who owns review for this one — it's been a little while, green and mergeable. Could you take a look or redirect me?

wakqasahmed:
Hi @bestbeforetoday — checking back in — no reviewer yet, still quiet. Happy to make any changes needed.

